PR #9, opened 2020-12-14
User Login + Member fetching

### LINKED ISSUE #1: Setup User creation with Login
The following must provide:
- [x] Ability to store users in DB
- [x] Ability to register new users
- [x] Ability to login with a given username/password
- [x] Users should have hashed/salted passwords
- [x] Users should use JWT for authentication on future APIs

majora2007:
Closes #1 and Closes #9


API/Controllers/AccountController.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                Username = user.UserName,
+                Token = _tokenService.CreateToken(user),
+                IsAdmin = user.IsAdmin
+            };
+        }
+        
+        private async Task<bool> UserExists(string username)
+        {
+            return await _context.Users.AnyAsync(user => user.UserName == username.ToLower());
```
asong91, line 91: Should this be moved to `UserRepository` instead ?

majora2007: Yeah, I have it in my current branch, I can raise another PR with UserExists and the finds refactored.

PR #20, opened 2021-01-07
Cover Images

### LINKED ISSUE #11: Generate Series Cover Image
Implement the ability to find the most appropriate cover image for a Series cover. 

Generally the first image of the first volume/chapter is used, unless there is an image file named folder in the archive. 

This will be overridden if there is a cover image file in the directory the archives are.

majora2007:
Closes #11 

Implements ability to set cover images from archive files and to force updating DB entries. Cover images will be compressed and a thumbnail will be saved rather than raw image due to size of raw files (1MB vs 60 KB)


majora2007:
I can't get the build to work, I'm merging for now with a broken build, even though it builds locally.

PR #42, opened 2021-01-25
server settings + Lots of Regex

### LINKED ISSUE #30: Ability to store Server Settings in DB.

### LINKED ISSUE #31: Logger should write to file, logger should be included in more calls.

### LINKED ISSUE #32: Admin: Reset a user's password

### LINKED ISSUE #37: Log to file

### LINKED ISSUE #39: BUG: On scanLibrary if there are errors or thrown exceptions, nothing writes to DB and an infinite loop of hangfire retrying is created.

### LINKED ISSUE #40: BUG: Scan Library does not properly remove series that don't exist or have been corrected due to better parsing.

### LINKED ISSUE #41: BUG: In some cases (Dorohedoro), not all parsed Infos will get added to the ConcurrentDictionary.
so what's happening is some volumes which are on disk and are being paresd correctly don't show up in the DB or DB entity creation code. 

I traced it down to a piece of code that updates the concurent dictionary. What's happening is that I scan the directories using multiple threads for speed. Each thread finds files, parses them and tries to add them to the concurent dictionary. 

Well for some reason, I just observed one thread add to the dictionary then another thread try to update the key (same key), but update failed, so it went to add to an empty dictionary. Meaning we lost the first object added.

majora2007:
A large PR that was tested against my own Manga library. Initial support for Server Settings is in, but nothing is usable in FE yet. This PR includes major regex tweaks to get the library working AND a large change to how we keep track of series/parsed infos between threads due to a bug with ConcurrentBag/ConcurrentDictionary where data was lost between threads. 

This also removes HTTPS from the codebase. This is due to publishing the app, we shouldn't have cert shipping. 

Fixes #41 
Fixes #40 
Fixes #30 
Fixes #39 
Fixes #37 
Fixes #32 
Fixes #31


majora2007:
Test works fine locally, going to merge and ensure it works on next PR.

PR #55, opened 2021-02-10
Partial Chapter support + ScanLibrary rewrite

### LINKED ISSUE #44: Investigate extras for manga (chapter 11.1)
Support for partial chapters (extras, double format)

### LINKED ISSUE #50: BUG: ScanLibrary issues causing fallout
There is a bug in ScanLibrary for Chapters where multiple chapters get created and thus exceptions are caused by SingleOrDefault() causing whole series to be skipped.

### LINKED ISSUE #51: BUG: ScanLibrary blocks main thread
When ScanLibrary is running, the main thread gets blocked and APIs and Hangfire cannot load properly.

### LINKED ISSUE #52: Investigate splitting Metadata (cover image) generation from Scanning code.
Parsing disk scans is cheap. For a decently sized library we are talking < 3 seconds. 

Since it is so cheap to parse Disk IO and this doesn't have to be done in real time, we can move archive work (metadata) off into a separate task that executes after the initial DB update from Scan Library.

This will ensure the entities get entered into DB as fast as possible and metadata can follow. 

This story is to investigate the performance benefit vs code complexity.

### LINKED ISSUE #53: ScanLibrary: Bugs and test coverage
During partial support implementation, many bugs have been found in ScanLibrary. Many of those bugs are fixed, but upon fixing them and refactoring the code to be more streamlined, it left me wanting a way to mock testing (have to use PrivateObjects or protected internal hacks) for the code. 

I need to build out a test utility that generates fake archive files for testing and then cleans up after itself.

majora2007:
This PR orignally was intended to support partial chapters, but during testing I found and fixed enough bugs to warrant rewriting how ScanLibrary works to be slightly slower (more verbose), but easier to understand. For a critical part of the application, I opted to keep it as simple as possible.

Fixes #44 
Fixes #50 
Fixes #51 
Fixes #52 
Fixes #53

### DIFF
--- a/API.Tests/API.Tests.csproj
+++ b/API.Tests/API.Tests.csproj
@@ -25,7 +25,7 @@
     </ItemGroup>
 
     <ItemGroup>
-      <Folder Include="Services\Test Data\ArchiveService" />
+      <Folder Include="Services\Test Data\ArchiveService\ComicInfos" />
     </ItemGroup>
 
 </Project>
--- a/API.Tests/Helpers/PrivateObjectPrivateType.cs
+++ b/API.Tests/Helpers/PrivateObjectPrivateType.cs
@@ -1,0 +1,1864 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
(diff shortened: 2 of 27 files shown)


majora2007:
Tests do work, issue with Github Actions. Will fix towards end of release.

Issue #244, opened 2021-06-02
Account registration says a password must be between 4 and 32 characters, but Login requirements is min of 6.


majora2007:
Steps to Reproduce:
1. Delete DB
2. Register an account and type a password that is 4 characters

Expected:
Minimum should report 6 characters.

PR #247, opened 2021-06-03
fix: wrong password length validation message

leo2d:
fix: wrong password length validation when registering a new user or resetting password

#244


majora2007 (approved):
LGTM

Issue #246, opened 2021-06-02
Changing a users password throws and error and logs the user out


majora2007:
Steps to reproduce:
1. Login as an admin
2. Open Server Settings -> Users -> Change password for a non-admin user (you can create one before this step)
3. Click Change password button (middle button)
4. Type any new password

Expected:
Modal should close and success toaster should appear

Actual:
Modal remains open, error toaster appears and user is logged out and put on login page

(This is likely due to error interceptor handling the error wrong)

majora2007:
I found this in Sentry for this

Server returned code 401 with body "You are not permitted to this operation."

PR #252, opened 2021-06-03
fix: error when resetting password of a non admin user

leo2d:
#246


majora2007:
I hate you Github Actions, code looks good, I'll merge.

PR #253, opened 2021-06-03
Backend for Recently Added Page

therobbiedavis:
- Tweaked Recently Added on the homepage to accept a custom limit
- Added new route, 'series-by-recently-added' to get pagination for recently added page

### DIFF
--- a/API/Controllers/SeriesController.cs
+++ b/API/Controllers/SeriesController.cs
@@ -148,7 +148,7 @@ namespace API.Controllers
         }
 
         [HttpGet("recently-added")]
-        public async Task<ActionResult<IEnumerable<SeriesDto>>> GetRecentlyAdded(int libraryId = 0, int limit = 20)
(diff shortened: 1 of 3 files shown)


API/Controllers/SeriesController.cs
```diff
@@ -148,7 +148,7 @@ public async Task<ActionResult> UpdateSeries(UpdateSeriesDto updateSeries)
         }
 
         [HttpGet("recently-added")]
-        public async Task<ActionResult<IEnumerable<SeriesDto>>> GetRecentlyAdded(int libraryId = 0, int limit = 20)
+        public async Task<ActionResult<IEnumerable<SeriesDto>>> GetRecentlyAdded(int limit, int libraryId = 0)
```
majora2007, line 151: We don't need to add a new API, but instead can just pass [FromQuery] UserParams userParams. This is where I store Pagination information, like in other methods here (see Collections or Series).

PR #348, opened 2021-06-30
Set dark mode to true and migration

### LINKED ISSUE #347: Dark mode is on by default
Making the default style dark mode. 

Users can still change to light mode if they prefer by changing dark mode to false.

therobbiedavis:
- Set SiteDarkMode = true in appuserpreference
- Ran migration for darkmode to be default true in db

Closes #347


API/Data/Migrations/20210630155028_DefaultDarkMode.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
majora2007, line 1: Migration wasn't created, so we can rollback the migration and thus remove these files. 

By defaulting it on the model, it will ensure when the first user is created, the settings has dark mode enabled.

majora2007 (changes requested):

API/Data/Migrations/20210630155028_DefaultDarkMode.cs
therobbiedavis: Removed migrations.

Marking resolved.

majora2007 (approved):

PR #375, opened 2021-07-06
Raw Image Support

majora2007:
* New: Ability to add Raw Image folders to Kavita via new library Types Images (Comic) and Images (Manga).

Images must belong to a folder, they cannot exist in the root directory. It is important to at least put a second folder (minimum) with a Volume of Chapter, else you will end up with each image as a special which is not easily readable. 

image
image


majora2007:
Don't merge this.

majora2007:
This needs a slight change for cover generation on image only directories. The code should search for a cover or folder file then fallback to the first image, like the archive code does.

PR #382, opened 2021-07-08
Added stats folder persistence

Kizaing:
Added a symbolic link to persist the stats folder between docker container updates.


majora2007:
Retargetted to monorepo branch since build system changes are happening there and this makes merge easier.

PR #412, opened 2021-07-21
Fixed: Series detail spacing issue while keeping a restriction on height for webbtoons

therobbiedavis:
## Fixed:
- Fixed: When the series detail thumbnail was too long it would push down the Volumes/Specials selection.

## Changed:
- Modified: increased the max-height to incorporate a percentage width for responsive scaling.


majora2007 (approved):
LGTM. Works as expected.

PR #442, opened 2021-07-27
Filtering First Pass

### LINKED ISSUE #439: First Pass Filtering
With the addition of mixed media libraries, it can be difficult to clean up series. Since filtering is planned for next release, this is a stop gap to allow for some quick filtering.

majora2007:
This is a first pass towards implementing Filtering. This gives the users the ability to filter against the format of the series from a detail page. The UI is non-final and more filtering options will be added (or reworked) in a later release. 

# Added
- Added: Added "In Progress" view to see everything you are currently reading
- Added: Added the ability to filter series based on format from "In Progress", "Recently Added", "Library Detail" pages.
- Added: Added total items to the above pages to showcase total series within Kavita

# Dev
- Changed out the lazy loader library from ng-lazyload-images to lazysizes
- Card-Layout-Component now does not refresh between pagination operations

filter first pass

Closes #439

### DIFF
--- a/API/Controllers/SeriesController.cs
+++ b/API/Controllers/SeriesController.cs
@@ -28,12 +28,12 @@ namespace API.Controllers
             _unitOfWork = unitOfWork;
         }
 
-        [HttpGet]
-        public async Task<ActionResult<IEnumerable<Series>>> GetSeriesForLibrary(int libraryId, [FromQuery] UserParams userParams)
+        [HttpPost]
+        public async Task<ActionResult<IEnumerable<Series>>> GetSeriesForLibrary(int libraryId, [FromQuery] UserParams userParams, [FromBody] FilterDto filterDto)
         {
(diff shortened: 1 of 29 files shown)


majora2007:
Some notes for myself:

- [ ] Bump Version
- [ ] Apply CSS transitions on open/close of the filter area

majora2007:
This has a major issue in that the lazy loading image library we use does not work well with DOM Mutation events occur, due to using IntersectionObserver. This means that we are forced to refresh the whole page for the lazy loading to resolve, however that means we cannot have persistence in our filtering area.

I have tried with the following to replace:
- [ x ] Native, loading="lazy". Issues: Lack of placeholder, Out of order loading
- [ x ] Yall.js. Issues: Lack of placeholder, Out of order loading
- [ x ] lazysizes. Issues: First load doesn't trigger images to load non-placeholder images (this is almost perfect)

PR #473, opened 2021-08-11
Notifier Tweaks

majora2007:
# Fixed
- Fixed: Fixed some https issue on some users machines when checking Github.

# Changed
- Changed: Tweaked the wording to imply you can leave the page for check for update.


majora2007:
This is on hold until I find a better solution.

PR #532, opened 2021-08-28
[BUGFIX] Fix some null crashes

ChristofferGreen:
On first startup of Kavita some data may be null.
Added protection against accesses of such null data.


majora2007:
These null checks are not needed. The reason is var user = await _unitOfWork.UserRepository.GetUserByUsernameAsync(User.GetUsername()) has a check and if the user doesn't exist, we throw an Exception which halts the code from moving forwards, as the user needs to log in.

PR #538, opened 2021-08-31
Add support for mangapy syntax to parser

ChristofferGreen:
# Added
- Added: (Parser) Added support for MangaPy's default naming convention to parser: vol_001-1.cbz

### DIFF
--- a/API.Tests/Parser/MangaParserTests.cs
+++ b/API.Tests/Parser/MangaParserTests.cs
@@ -26,6 +26,7 @@ namespace API.Tests.Parser
         [InlineData("Akame ga KILL! ZERO v01 (2016) (Digital) (LuCaZ).cbz", "1")]
         [InlineData("v001", "1")]
         [InlineData("Vol 1", "1")]
+        [InlineData("vol_356-1", "356")] // Mangapy syntax
         [InlineData("No Volume", "0")]
         [InlineData("U12 (Under 12) Vol. 0001 Ch. 0001 - Reiwa Scans (gb)", "1")]
         [InlineData("[Suihei Kiki]_Kasumi_Otoko_no_Ko_[Taruby]_v1.1.zip", "1")]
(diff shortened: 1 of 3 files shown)


API.Tests/Parser/MangaParserTests.cs
```diff
@@ -26,6 +26,7 @@ public MangaParserTests(ITestOutputHelper testOutputHelper)
         [InlineData("Akame ga KILL! ZERO v01 (2016) (Digital) (LuCaZ).cbz", "1")]
         [InlineData("v001", "1")]
         [InlineData("Vol 1", "1")]
+        [InlineData("vol_356-1", "356")] // Mangapy syntax
```
majora2007, line 29: What is the -1? Is it a chapter, if so, we should add a case for it.

API/Parser/Parser.cs
```diff
@@ -83,6 +83,11 @@ public static class Parser
                 @"(?<Series>.*)(\b|_|)(S(?<Volume>\d+))",
                 RegexOptions.IgnoreCase | RegexOptions.Compiled,
             RegexTimeout),
+            // vol_001-1.cbz
```
majora2007, line 86: Put in your comment this is a format for MangaPy itself. So: // vol_001-1.cbz for MangaPy default naming convention

majora2007:
In order for the build to properly parse, your PR needs to use markdown like the other PRs.

For example, this PR's body should be:

# Added
- Added: Added support for MangaPy's default naming convention to parser <Series>-vol_001-1.cbz. 

If there is an issue, make sure you link it on the PR or use Closes ISSUENUMBER or Fixes ISSUENUMBER if it is a defect. 

Once you add chapter support, I'll test and then can merge if all is good.

majora2007 (changes requested):
Please update PR body to standard and add chapter support

API.Tests/Parser/MangaParserTests.cs
ChristofferGreen: -1, -2, -3, etc is a number for corrections, if they find spelling errors and so on they upload a new version and increase that number. In theory, at least, all the manga I get downloaded have -1 though (weird).

majora2007:
Oh interesting, okay then let me pull it down. I'm planning a release
today, so we can include this. Please adjust the PR title and the comment
as requested.

PR #553, opened 2021-09-04
Versioning Fix & Improvement

### LINKED ISSUE #544: Changelog view is showing the latest version as installed even if you're behind a version
Be on version v0.4.4 and have v0.4.5 out, the v0.4.5 will show as installed.

Inpacchi:
# Added
* Added: "Available" now displays on all newer, uninstalled updates

# Fixed
* Fixed: The changelog now shows the correct version currently installed, as well as any versions available for update

Fixes #544

### DIFF
--- a/API/Services/Tasks/VersionUpdaterService.cs
+++ b/API/Services/Tasks/VersionUpdaterService.cs
@@ -90,14 +90,12 @@ namespace API.Services.Tasks
         private UpdateNotificationDto CreateDto(GithubReleaseMetadata update)
         {
             if (update == null || string.IsNullOrEmpty(update.Tag_Name)) return null;
-            var buildVersion = BuildInfo.Version;
             var updateVersion = new Version(update.Tag_Name.Replace("v", string.Empty));
-            var currentVersion = buildVersion.ToString();
+            var currentVersion = BuildInfo.Version.ToString();
 
             if (updateVersion.Revision == -1)
             {
-                currentVersion = buildVersion.ToString()
-                    .Substring(0, buildVersion.ToString().LastIndexOf("."));
+                currentVersion = currentVersion.Substring(0, currentVersion.LastIndexOf("."));
             }
 
             return new UpdateNotificationDto()
--- a/API/Services/Tasks/VersionUpdaterService.cs
+++ b/API/Services/Tasks/VersionUpdaterService.cs
@@ -90,13 +90,14 @@ namespace API.Services.Tasks
         private UpdateNotificationDto CreateDto(GithubReleaseMetadata update)
         {
             if (update == null || string.IsNullOrEmpty(update.Tag_Name)) return null;
+            var buildVersion = BuildInfo.Version;
             var updateVersion = new Version(update.Tag_Name.Replace("v", string.Empty));
-            var currentVersion = BuildInfo.Version.ToString();
+            var currentVersion = buildVersion.ToString();
 
             if (updateVersion.Revision == -1)
             {
-                currentVersion = BuildInfo.Version.ToString()
-                    .Substring(0, BuildInfo.Version.ToString().LastIndexOf("."));
+                currentVersion = buildVersion.ToString()
+                    .Substring(0, buildVersion.ToString().LastIndexOf("."));
             }
 
             return new UpdateNotificationDto()
--- a/API/Services/Tasks/VersionUpdaterService.cs
+++ b/API/Services/Tasks/VersionUpdaterService.cs
@@ -91,9 +91,13 @@ namespace API.Services.Tasks
         {
             if (update == null || string.IsNullOrEmpty(update.Tag_Name)) return null;
             var updateVersion = new Version(update.Tag_Name.Replace("v", string.Empty));
-            var currentVersion = updateVersion.Revision == -1
-                ? BuildInfo.Version.ToString().Substring(0, 5)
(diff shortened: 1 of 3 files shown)


API/Services/Tasks/VersionUpdaterService.cs
```diff
@@ -90,12 +90,14 @@ public async Task<IEnumerable<UpdateNotificationDto>> GetAllReleases()
         private UpdateNotificationDto CreateDto(GithubReleaseMetadata update)
         {
             if (update == null || string.IsNullOrEmpty(update.Tag_Name)) return null;
-            var version = update.Tag_Name.Replace("v", string.Empty);
-            var updateVersion = new Version(version);
+            var updateVersion = new Version(update.Tag_Name.Replace("v", string.Empty));
+            var currentVersion = updateVersion.Revision == -1
+                ? BuildInfo.Version.ToString().Substring(0, 5)
```
majora2007, line 95: Don't substring as if we have 0.10.1.1, it will break. If anything, you can substring(0, BuildInfo.Version.ToString().LastIndexOf("."))

but move the BuildInfo.Version.ToString() to a local variable.

UI/Web/src/app/admin/changelog/changelog.component.html
```diff
@@ -1,7 +1,10 @@
 <ng-container *ngFor="let update of updates; let indx = index;">
     <div class="card w-100 mb-2" style="width: 18rem;">
         <div class="card-body">
-          <h5 class="card-title">{{update.updateTitle}}&nbsp;<span class="badge badge-secondary" *ngIf="update.updateVersion === update.currentVersion">Installed</span></h5>
+          <h5 class="card-title">{{update.updateTitle}}&nbsp;
+              <span class="badge badge-secondary" *ngIf="update.updateVersion <= update.currentVersion">Installed</span>
```
majora2007, line 5: We can use an else statement on the ngIf to avoid potential bugs and duplicate logic.
```
*ngIf="update.updateVersion <= update.currentVersion; else available"
<ng-template #available>
 <span class="badge badge-secondary">Available</span>
          </h5>
</ng-template>
```

majora2007 (changes requested):
Please update the PR Body and the following comments.

UI/Web/src/app/admin/changelog/changelog.component.html
Inpacchi: Didn't realize if...else logic was available in Angular templating. That's pretty awesome and good to know. Made the change.

API/Services/Tasks/VersionUpdaterService.cs
Inpacchi: Duh, don't know how I didn't think of that. Made the change.

Inpacchi:
PR body updated.

API/Services/Tasks/VersionUpdaterService.cs
```diff
@@ -90,12 +90,18 @@ public async Task<IEnumerable<UpdateNotificationDto>> GetAllReleases()
...
         {
             if (update == null || string.IsNullOrEmpty(update.Tag_Name)) return null;
-            var version = update.Tag_Name.Replace("v", string.Empty);
-            var updateVersion = new Version(version);
+            var updateVersion = new Version(update.Tag_Name.Replace("v", string.Empty));
+            var currentVersion = BuildInfo.Version.ToString();
+
+            if (updateVersion.Revision == -1)
+            {
+                currentVersion = BuildInfo.Version.ToString()
```
majora2007, line 98: cache this into a variable to avoid duplicate string creations

Inpacchi: So you mean instead of modifying the currentVersion variable, create a new variable?

majora2007: No, I mean just cache the BuildInfo.Version.ToString() then use that in the following lines of code. 

Essentially, just want to avoid 2 string creations.

Inpacchi: Ah, I understand now. Got it!

API/Services/Tasks/VersionUpdaterService.cs
```diff
@@ -90,12 +90,19 @@ public async Task<IEnumerable<UpdateNotificationDto>> GetAllReleases()
...
             if (update == null || string.IsNullOrEmpty(update.Tag_Name)) return null;
-            var version = update.Tag_Name.Replace("v", string.Empty);
-            var updateVersion = new Version(version);
+            var buildVersion = BuildInfo.Version;
+            var updateVersion = new Version(update.Tag_Name.Replace("v", string.Empty));
+            var currentVersion = buildVersion.ToString();
+
+            if (updateVersion.Revision == -1)
+            {
+                currentVersion = buildVersion.ToString()
```
majora2007, line 99: You're close, so it's like this:
```
var currentVersion = buildVersion.ToString();
currentVersion = currentVersion .SubString(0, buildVersion.LastIndexOf("."))
```
You don't want to keep calling toString because it keeps creating a temp string.

Inpacchi: Ah, okay. Sorry, still brushing up on my C#. Python has spoiled me, lol.

majora2007 (approved):

PR #555, opened 2021-09-04
Added client & server-side checks for whitespace in search query (#547)

### LINKED ISSUE #547: Search doesn't return results when you start your query with a space
Any query that starts or ends with a space does not return results.

Inpacchi:
# Fixed
* Fixed: Search queries will be stripped of leading whitespace

Fixes #547

### DIFF
--- a/API/Controllers/LibraryController.cs
+++ b/API/Controllers/LibraryController.cs
@@ -227,6 +227,11 @@ namespace API.Controllers
         {
             queryString = queryString.Replace(@"%", "");
 
+            if (char.IsWhiteSpace(queryString, 0))
+            {
+                queryString = queryString.TrimStart();
+            }
+
             var user = await _unitOfWork.UserRepository.GetUserByUsernameAsync(User.GetUsername());
             // Get libraries user has access to
             var libraries = (await _unitOfWork.LibraryRepository.GetLibrariesForUserIdAsync(user.Id)).ToList();
--- a/API/Controllers/LibraryController.cs
+++ b/API/Controllers/LibraryController.cs
@@ -227,9 +227,9 @@ namespace API.Controllers
         {
             queryString = queryString.Replace(@"%", "");
 
-            if (char.IsWhiteSpace(queryString, 0))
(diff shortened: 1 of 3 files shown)


API/Controllers/LibraryController.cs
```diff
@@ -227,6 +227,11 @@ public async Task<ActionResult<IEnumerable<SearchResultDto>>> Search(string quer
         {
             queryString = queryString.Replace(@"%", "");
 
+            if (char.IsWhiteSpace(queryString, 0))
+            {
+                queryString = queryString.TrimStart();
```
majora2007, line 232: Why can't we just use Trim itself? Does this work if we have ending space?

UI/Web/src/app/nav-header/nav-header.component.ts
```diff
@@ -72,7 +72,7 @@ export class NavHeaderComponent implements OnInit, OnDestroy {
 
   onChangeSearch(val: string) {
       this.isLoading = true;
-      this.searchTerm = val;
+      this.searchTerm = val.trimStart();
```
majora2007, line 75: Always trim. Please test to make sure it works in more than just the initial bug. The way I write bugs is very loose. It's expected to do full testing for other potential missed cases.

majora2007:
This removes spaces as you type, you cannot type 2 separate works.

API/Controllers/LibraryController.cs
```diff
@@ -227,6 +227,11 @@ public async Task<ActionResult<IEnumerable<SearchResultDto>>> Search(string quer
         {
             queryString = queryString.Replace(@"%", "");
 
+            if (char.IsWhiteSpace(queryString, 0) || char.IsWhiteSpace(queryString, queryString.Length - 1))
```
majora2007, line 230: This isn't very clean. Keep the code simple. Why doesn't just trimming work? Without the checks.

majora2007 (changes requested):

majora2007:
Closing this PR and moving story to dedicated release.

majora2007:
This has been implemented without issues in https://github.com/Kareadita/Kavita/pull/682

Deleting this branch as it had issues and developer has not been active in discord for weeks.

PR #594, opened 2021-09-23
Fixed iOS scaling issue

therobbiedavis:
# Fixed
- Fixed issue with scaling with mange reader on iOS. *Note* Browser is likely to crash with high resolution images due to filesize and caching multiple pages.


majora2007:
I'm going to hold on this for the current release and validate in v0.4.7, due to the nature of changing how we render on manga reader.

majora2007:
I'm declining this. For right now, the complexity it adds is not worth it.

Issue #664, opened 2021-10-13
When adding an odd number of series to library, one is always missing


majora2007:
Add 3 series not yet in a library, scan library.

Expected: 3 series added
Actual: 2 series added

majora2007:
Users were able to reproduce again: 
861 series, 8420 files. everything shows up, but Kavita stops getting metadata around page 7
Another user with 2267 series has same issue.

PR #681, opened 2021-10-16
Attempt to solve metadata refresh issue

majora2007:
# Changed
- Changed: Change chunk size to be a fixed 50 to validate if it's causing issue with refresh (#664)


majora2007:
This fixes #664

PR #698, opened 2021-10-20
Breaking Changes: Docker Parity

majora2007:
# Change
- Changed: Moved all config (db, logs, cache, appsettings.json, etc) to config/ directory. This is a **breaking change** for Docker users. You must rebuild your containers following the readme.


majora2007:
Issues I'm seeing (post vacation):
- [ ] Docker users that are upgrading from prior release aren't able to open the database because migration fails to trigger

PR #775, opened 2021-11-21
Horizontal Scroll

majora2007:
This PR has 2 critical issues that prevent it from being merged. I'm leaving it open in case anyone wants to fork and try and solve these.

# Added
- Added: New reading mode called Horizontal Webtoon. This allows for you to scroll horizontally to read comics. Thanks to @motty-mio2 for the foundational work to make this possible.


majora2007:
This cannot be merged. It has the following issues:
- [ ] Continuous reading css needs to be tweaked to properly be rotated
- [ ] On loading an image that is greater than screen display width, the prefetching of the images before the images will cause the reader to scroll infinitely back to page 0 (triggering prefetches, which cause the scrolling). This seems to be a core issue, needs deep investigation.

eswiig:
Love the webtoon mode, only improvement request I have is a "manual zoom slider" (so the images are not full width), sometimes I want to read at a "distance" instead of having my eyes travel so far while retaining the infinite scroll mode.

image

majora2007:
Hi @eswiig, this is not the place for feature requests. 

Please open a feature request here and Ill will evaluate it at a later time. Thanks.

[https://feats.kavitareader.com](https://feats.kavitareader.com)

PR #827, opened 2021-12-05
Update README.md

allhailthetail:
# Added
- Added: Instructions should have a distinction between the pre-built binary/installer and building it from source.  Also, should list out build dependencies; according to build.sh, they are npm and dotnet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Username: demouser
 Password: Demouser64
 ```
 
-## Setup
+## Installation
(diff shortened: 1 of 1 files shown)


majora2007:
I'm fine with changing the Setup -> Installation (despite there being no actual install), but Build dependencies shouldn't be on the main page and should stay within Contributing.md.

majora2007 (changes requested):
Please take care of my comment

PR #832, opened 2021-12-06
Feature/local metadata more tags

majora2007:
# Added
- Added: AgeRating is now supported in Kavita. This is stored at both the chapter level and series level. Series level contains the highest (worst) rating of all the chapters.
- Added: ReleaseDate is now supported in Kavita. This is stored at both the chapter and series levels. At a series level, only the lowest year is stored representing the first release of the Series. 

# Changed
- Changed: When a card is selected for bulk actions, there is now an outline
- Changed: For ComicInfo.xml scraped from ComicVine, Kavita will interpret results differently. If the volume is set to 1, Kavita will set it to 0 forcing everything to be treated as an issue. This needs feedback! 

# Fixed
- Fixed: Fixed a critical issue where cover directory didn't exist (due to a bad previous build) and when trying to write a cover image would crash Kavita (develop)

### DIFF
--- a/API.Tests/Entities/ComicInfoTests.cs
+++ b/API.Tests/Entities/ComicInfoTests.cs
@@ -1,0 +1,36 @@
+﻿using API.Data.Metadata;
+using API.Entities.Enums;
+using Xunit;
+
+namespace API.Tests.Entities;
+
+public class ComicInfoTests
+{
+    #region ConvertAgeRatingToEnum
+
+    [Theory]
+    [InlineData("G", AgeRating.G)]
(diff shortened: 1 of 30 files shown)


kleisauke:
> Fixed a crash when NetVips tries to write a cover file and cover directory is not existing.
> 597e57a3e8f5ad8a688284960190981d98bb7190

FWIW, this crash will be fixed in the forthcoming v8.12.2 release of NetVips.Native, see PR https://github.com/libvips/libvips/pull/2571.

PR #867, opened 2021-12-22
Add middle-click functionality to various cards

acsellers:
# Added
- Added: Implemented a middle click event and handlers on the obvious places where users might want to open something in a new tab.

### DIFF
--- a/UI/Web/src/app/cards/card-item/card-item.component.ts
+++ b/UI/Web/src/app/cards/card-item/card-item.component.ts
@@ -63,6 +63,10 @@ export class CardItemComponent implements OnInit, OnDestroy {
    */
   @Output() clicked = new EventEmitter<string>();
   /**
+   * Event emitted when item is middle-clicked
+   */
+  @Output() middleclicked = new EventEmitter<string>();
+  /**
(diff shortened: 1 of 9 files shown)


majora2007:
This is too much code duplication. Instead, we should look at using anchors with target _blank. This way we can utilize existing code and url routes.

majora2007 (changes requested):
This is too much code duplication. Instead, we should look at using anchors with target _blank on the cards or handling it internally in the card-item component. This way we can utilize existing code and url routes.

majora2007:
Not sure if you're going to come back and update it. I'll put this in a future release and tackle it then with the way I purposed.

Dimava:
Rather then do *anything* with middle-clicks it may be better to just do nothing and leave handling of everything except non-modified click to the browser default handlers - like <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>LMB</kbd> which has the default behaviour to open a link in a new foreground tab in most of browsers

snickers2k:
never used a "middle" click for years :D how to even do it on a mac with trackpad? O_o

PR #871, opened 2021-12-23
Added check to see if mount folder is empty

### LINKED ISSUE #844: Scan still running when mount is unmounted and using Docker
I am experiencing a bug where my docker install still runs a scan if my rclone mount goes down, thereby deleting my library.

Mount --> Path in Docker Container
/home/localretard/library/Library --> /library

This path in the docker container will still exist if the mount is unmounted. A check inside will need to be done to determine if it is empty.

Related code:
ScannerService, line 217
`        // Check if any of the folder roots are not available (ie disconnected from network, etc) and fail if any of them are
        if (library.Folders.Any(f => !_directoryService.IsDriveMounted(f.Path)))
        {
            _logger.LogError("Some of the root folders for library are not accessible. Please check that drives are connected and rescan. Scan will be aborted");
            return;
        }`

therobbiedavis:
# Fixed
- Fixed: Fixed a bug where running a scan in a docker container with a disconnected mount would still run. (Fixes #844 )

### DIFF
--- a/API/Services/DirectoryService.cs
+++ b/API/Services/DirectoryService.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -32,6 +32,7 @@ namespace API.Services
         void CopyFileToDirectory(string fullFilePath, string targetDirectory);
         int TraverseTreeParallelForEach(string root, Action<string> action, string searchPattern, ILogger logger);
         bool IsDriveMounted(string path);
+        bool IsDirectoryEmpty(string path);
         long GetTotalSize(IEnumerable<string> paths);
         void ClearDirectory(string directoryPath);
         void ClearAndDeleteDirectory(string directoryPath);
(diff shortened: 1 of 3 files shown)


API/Services/DirectoryService.cs
```diff
@@ -259,7 +260,12 @@ public bool IsDriveMounted(string path)
            return FileSystem.DirectoryInfo.FromDirectoryName(FileSystem.Path.GetPathRoot(path) ?? string.Empty).Exists;
        }
 
-       public string[] GetFilesWithExtension(string path, string searchPatternExpression = "")
+        public bool IsDirectoryEmpty(string path)
```
majora2007, line 263: Add a unit test for this please

majora2007 (changes requested):
After you update the message based on what was discussed in discord and write the unit test, we are good to go.

API/Services/DirectoryService.cs
therobbiedavis: Added in https://github.com/Kareadita/Kavita/pull/871/commits/a0fa4fb2800b5881b97000c4b73a59de3de68d7d

Resolving request.

majora2007 (approved):

PR #935, opened 2022-01-13
Fix image aspect ratio when image has inline style

andrewmackrodt:
# Fixed
- Fixed: Fixed a rendering issue in book reader where an image would loose its aspect ratio if it has an inline style, e.g. `height: (x)vh`

I've created a fiddle which demonstrates the fix https://jsfiddle.net/02od8bw4/
Issue discussed on Discord, affects RE:Zero Light Novel Vol. 1 (Amazon ID B01CO4B14I).


majora2007:
LGTM, thanks for the contribution.

PR #937, opened 2022-01-13
Performance Enhancements

therobbiedavis:
# Changed
- Changed: (Performance) Added a short circuit in CanOpen to check for rar files and force SharpCompress for deflating. This avoids an IO and speeds up processing.
- Changed: (Performance) Added more Split Queries for many queries that are JOINs, speeding them up significantly

### DIFF
--- a/API/Services/ArchiveService.cs
+++ b/API/Services/ArchiveService.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
--- a/API/Services/ArchiveService.cs
+++ b/API/Services/ArchiveService.cs
@@ -57,12 +57,8 @@ namespace API.Services
         {
             if (string.IsNullOrEmpty(archivePath) || !(File.Exists(archivePath) && Parser.Parser.IsArchive(archivePath) || Parser.Parser.IsEpub(archivePath))) return ArchiveLibrary.NotSupported;
 
-            try
-            {
-                using var a2 = ZipFile.OpenRead(archivePath);
-                return ArchiveLibrary.Default;
-            }
-            catch (Exception)
+            var ext = _directoryService.FileSystem.Path.GetExtension(archivePath);
+            if (ext == ".cbr" || ext == ".rar")
             {
                 try
                 {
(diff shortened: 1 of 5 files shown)


API/Services/ArchiveService.cs
```diff
@@ -57,12 +57,8 @@ public virtual ArchiveLibrary CanOpen(string archivePath)
...
             if (string.IsNullOrEmpty(archivePath) || !(File.Exists(archivePath) && Parser.Parser.IsArchive(archivePath) || Parser.Parser.IsEpub(archivePath))) return ArchiveLibrary.NotSupported;
 
-            try
-            {
-                using var a2 = ZipFile.OpenRead(archivePath);
-                return ArchiveLibrary.Default;
-            }
-            catch (Exception)
+            var ext = _directoryService.FileSystem.Path.GetExtension(archivePath);
+            if (ext == ".cbr" || ext == ".rar")
```
majora2007, line 61: Change this to if (ext == ".cbr" || ext == ".rar") return ArchiveLibrary.SharpCompress

majora2007: Also make sure you do a .toLower() (actually .toUpper() would be better) because some files are .RAR

therobbiedavis: Resolve in commit 731d078

API/Services/ArchiveService.cs
```diff
@@ -57,6 +57,9 @@ public virtual ArchiveLibrary CanOpen(string archivePath)
         {
             if (string.IsNullOrEmpty(archivePath) || !(File.Exists(archivePath) && Parser.Parser.IsArchive(archivePath) || Parser.Parser.IsEpub(archivePath))) return ArchiveLibrary.NotSupported;
 
+            var ext = _directoryService.FileSystem.Path.GetExtension(archivePath);
+            if (ext.ToUpper() == ".CBR" || ext.ToUpper() == ".RAR") return ArchiveLibrary.SharpCompress;
```
majora2007, line 61: This just needs to be .Equals()

majora2007 (changes requested):
Just one comment, then it's good

API/Services/ArchiveService.cs
therobbiedavis: Resolved with 81c0834

PR #962, opened 2022-01-19
Fix for 3rd Party Clients copy button

therobbiedavis:
# Fixed
- Fixed: Fixed an issue where the copy button wasn't working on the "3rd Party Clients" page.

### DIFF
--- a/UI/Web/src/app/user-settings/api-key/api-key.component.html
+++ b/UI/Web/src/app/user-settings/api-key/api-key.component.html
@@ -4,7 +4,7 @@
     <div class="input-group">
         <input #apiKey type="text" readonly class="form-control" id="api-key--{{title}}" aria-describedby="button-addon4" [value]="key" (click)="selectAll()">
         <div class="input-group-append" id="button-addon4">
-            <button class="btn btn-outline-secondary" type="button" (click)=copy(apiKey)><span class="sr-only">Copy</span><i class="fa fa-copy" aria-hidden="true"></i></button>
+            <button class="btn btn-outline-secondary" type="button" (click)=copy()><span class="sr-only">Copy</span><i class="fa fa-copy" aria-hidden="true"></i></button>
             <button class="btn btn-danger" type="button" (click)="refresh()" *ngIf="showRefresh"><span class="sr-only">Regenerate</span><i class="fa fa-sync-alt" aria-hidden="true"></i></button>
         </div>
     </div>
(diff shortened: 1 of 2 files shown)


UI/Web/src/app/user-settings/api-key/api-key.component.ts
```diff
@@ -43,8 +44,10 @@ export class ApiKeyComponent implements OnInit, OnDestroy {
     this.onDestroy.complete();
   }
 
-  async copy() {
-    await navigator.clipboard.writeText(this.key);
+  async copy(input: any) {
```
majora2007, line 47: First, don't use any types, this should be a string. Second, why are you passing it from the template and not just referencing this.key directly?

majora2007: Reading the copy code vs how it's invoked, I'm confused. It's supposed to be an HTMLInputElement type, but not sure how the template is getting a reference to this. 

I think the mode ideal way would be to setup a ViewChild to reference the input:

@ViewChild('keyInput') inputElem: ElementRef<HTMLInputElement>;

Then in copy, use this.inputElem.nativeElement to use the .select and setSelectionRange, etc.

majora2007 (changes requested):

UI/Web/src/app/user-settings/api-key/api-key.component.ts
therobbiedavis: Resolved in 9692ab3

UI/Web/src/app/user-settings/api-key/api-key.component.html
```diff
@@ -4,7 +4,7 @@
     <div class="input-group">
         <input #apiKey type="text" readonly class="form-control" id="api-key--{{title}}" aria-describedby="button-addon4" [value]="key" (click)="selectAll()">
         <div class="input-group-append" id="button-addon4">
-            <button class="btn btn-outline-secondary" type="button" (click)="copy()"><span class="sr-only">Copy</span><i class="fa fa-copy" aria-hidden="true"></i></button>
+            <button class="btn btn-outline-secondary" type="button" (click)=copy()><span class="sr-only">Copy</span><i class="fa fa-copy" aria-hidden="true"></i></button>
```
majora2007, line 7: Looks like you're missing the "" around the copy call. Please add them in.

majora2007 (changes requested):

UI/Web/src/app/user-settings/api-key/api-key.component.html
therobbiedavis: Resolved in 8922be2

majora2007 (approved):

PR #984, opened 2022-01-24
Library page resetting

therobbiedavis:
# Fixed
- Fixed: Fixed an issue where the pagination would reset to 1 on the library-details page when refreshing or navigating directly to a specific page number.


majora2007:
Can you check this in collection-detail page as well? it also shares the same code.

therobbiedavis:
> Can you check this in collection-detail page as well? it also shares the same code. 

Just checked, it works there as well. No issues.

PR #990, opened 2022-01-26
Fix for add to collection visibility issue

therobbiedavis:
# Fixed
- Fixed: Fixed a bug where users couldn't select a collection when they had a lot of them.

### DIFF
--- a/UI/Web/src/app/cards/_modals/bulk-add-to-collection/bulk-add-to-collection.component.html
+++ b/UI/Web/src/app/cards/_modals/bulk-add-to-collection/bulk-add-to-collection.component.html
@@ -1,46 +1,48 @@
 
-<div class="modal-header">
-    <h4 class="modal-title" id="modal-basic-title">Add to Collection</h4>
-    <button type="button" class="close" aria-label="Close" (click)="close()">
-    <span aria-hidden="true">&times;</span>
-    </button>
-</div>
-<form style="width: 100%" [formGroup]="listForm">
-<div class="modal-body">
-    <div class="form-group" *ngIf="lists.length >= 5">
-        <label for="filter">Filter</label>
-        <div class="input-group">
-            <input id="filter" autocomplete="off" class="form-control" formControlName="filterQuery" type="text" aria-describedby="reset-input">
-            <div class="input-group-append">
-                <button class="btn btn-outline-secondary" type="button" id="reset-input" (click)="listForm.get('filterQuery')?.setValue('');">Clear</button>
+<div id="add-to-collection">
+    <div class="modal-header">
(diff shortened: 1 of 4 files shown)


UI/Web/src/app/cards/_modals/bulk-add-to-collection/bulk-add-to-collection.component.html
```diff
@@ -1,46 +1,48 @@
...
-</div>
-<form style="width: 100%" [formGroup]="listForm">
-<div class="modal-body">
-    <div class="form-group" *ngIf="lists.length >= 5">
-        <label for="filter">Filter</label>
-        <div class="input-group">
-            <input id="filter" autocomplete="off" class="form-control" formControlName="filterQuery" type="text" aria-describedby="reset-input">
-            <div class="input-group-append">
-                <button class="btn btn-outline-secondary" type="button" id="reset-input" (click)="listForm.get('filterQuery')?.setValue('');">Clear</button>
+<div id="add-to-collection">
```
majora2007, line 2: Don't use ids

majora2007: In addition, modals have a fixed structure, keep the structure as stated, don't wrap with a div. See:
https://ng-bootstrap.github.io/#/components/modal/examples

majora2007 (changes requested):
Please try to solve this with the options already provided with ng-bootstrap's modal rather than rolling your own. 

Please test this on all devices plus other screens that have this type of thing. I'm pretty sure we just need to pass scrollable to the code that opens the modal.

UI/Web/src/app/cards/_modals/bulk-add-to-collection/bulk-add-to-collection.component.html
therobbiedavis: ID removed in abe0d9f

therobbiedavis:
> Please try to solve this with the options already provided with ng-bootstrap's modal rather than rolling your own.
> 
> Please test this on all devices plus other screens that have this type of thing. I'm pretty sure we just need to pass scrollable to the code that opens the modal.

Tried passing scrollable, however there is some css being added in the head that I can not find.
image

I reviewed the bootstrap modal documentation and found that `encapsulation: ViewEncapsulation.None,` needed to be added to the component in order for the styles to pick up. Please see the testing verification below:

Desktop:
Screenshot 2022-01-26 210209

Mobile:
unnamed

https://user-images.githubusercontent.com/7907265/151279265-1d96c476-065d-4282-85ee-064fc7fe20b5.mp4

majora2007 (approved):
Looks good and feels good, let's merge

PR #991, opened 2022-01-26
Fixed Parser.Parser.Normalize returns empty string

jeongsu816:
# Fixed
- Fixed: When normalizing series names, if the series has no alphanumeric characters (ie non-roman), return the original name back rather than empty string.


majora2007:
Can you give me some examples of when this breaks? Can you add Unit tests for this method that encapsulate your issue?

PR #1057, opened 2022-02-11
v0.5.1 Release

majora2007:
Another jam packed update is here and it lays out some important foundational work for future updates. The focus of this update was account management and search overhaul. Both are functions that were build out very early in the project and haven't had much love in quite some time. The key accomplishments that I wanted was making account registration easier and more streamlined and overhaul search to work against the newly added metadata from last release. 

First off, accounts now require emails to function. When you start v0.5.1 up, a one time migration will occur to have you confirm your email address. After that, you can now easily invite users to join your server and your user's can register their account from a unique link sent from your server. User's can now use forgot password as well to change their password, without you having to change and send to them manually. 

If you are concerned about your email, please note, that our email service (gmail) automatically delete emails that flow through them immediately and is only accessible by myself. If you are not satisfied with this, we offer a separate service for the email, so you can use your own SMTP server. You can find details [here](https://github.com/Kareadita/KavitaEmail). If you don't want that, you can either access and invite users when your server is not accessible to the web (aka localhost/internal ip) or change the email service to a bad url, which will always fail and use the fallback. 

Next up is search and I have to give a huge thanks to @therobbiedavis who really polished the UI up quite a bit. Our search now searches against the metadata fields, like people, genres, tags, it can search against collections and reading lists and is still blazing fast. The UX has been really polished and should work well on even the smallest devices. 

As an added bonus to this release, we have refresh tokens which keep you authenticated without you having to re-login and Tachiyomi progress syncing is done and a Pull Request is open to have Tachiyomi support us. 

This has been another great release. The next 2 releases are planned with a focus on UX and the readers. It will take some time, please be patient, but I think the results will speak for themselves. 

# Added
- Added: Added 'f' as keybind to toggle fullscreen on manga and book readers.
- Added: Added Refresh Tokens. Users are issued tokens that last 14 days, just before the 14 days, the UI will request a new token to avoid having to re-authenticate.
- Added: Changed the Dockerfile to include a health check to make sure Kavita is running smoothly.
- Added: Added support for Tachiyomi to sync progress
- Added: Added a new section to the dashboard that shows when new volumes and chapters are added to the library, providing more detail than just the Series itself (Recently Added Chapters)
- Added: Invite users to your server through email and they can setup their accounts from the emailed link
- Added: All accounts now must have an email assigned to their account and must verify said email.
- Added: Added ability to reset password via email from login page.
- Added: Implemented a new role 'Change Password'. This role allows you to change your own password. By default, all users will have it. A user can have it removed by an admin.
- Added: Added a new activity stream on the dashboard which shows Recently Updated Series with a number badge showing the number of chapter/volumes added.
- Added: Added the ability to search for metadata fields (people, genre, tags), series, collections, and reading lists. This also applies to OPDS where applicable.
- Added: Added data collection on 2 new fields: NumberOfLibraries which will help for some upcoming design work and HasBookmarks, to see if other users are using this functionality and the scope of changes I can make to stabilize them further.

# Changed
- Changed: When migrations occur, create a temp/migration/version/kavita.db backup, attempt to run migrations and if any fail, restore the temp/migration/version/kavita.db backup and exit immediately with verbose logging. 
- Changed: Admins should have ability to delete a user's reading list (api only)
- Changed: Moved calculation for getting the resume point to the backend
- Changed: Removed showing toasts when series are added to the server since cards will show automatically.
- Changed: Changed how Kavita finds the ComicInfo.xml to prefer top-level documents, which is the current way most tools work.
- Changed: Series detail now will have dedicated tabs for Volumes and Chapters/Issues, in addition to Storyline (which is the existing view). Storyline is selected by default and has side-by-side grouping of volumes and chapters. Volumes shows just Volumes, Chapters/Issues shows just those underlying chapters, especially good for users with chapters grouped into volumes.
- Changed: Added a spinner when epub reader is loading the book
- Changed: Server settings is now next to progress button in header.
- Changed: Admins can change an admin's username/role/library access.
- Changed: Admins can remove another user's admin role.
- Changed: Changed (you) to a star to indicate who you're logged in as.
- Changed: Editing a user's role/library access is now streamlined.
- Changed: (Performance) Put an optimization in refresh cover images to avoid 3 db calls.
- Changed: Brand spanking new UX for search screen. Search will now open up the results in a modal like window, the search bar will grow to eat the rest of the screen width.
- Changed: Tweaked the card design to help them standout better on dark mode.
- Changed: Users can now user their own email service (docker) to allow them to use their own SMTP servers. By default, Kavita will use it's email service, which auto-deletes email records after sending, even on failure. No emails leave the user's install until an invite occurs, giving the user the opportunity to use their own service.
- Changed: Changed the default margin on mobile devices for book reader to 5% to give a little padding from edge
- Changed: On Bookmarks tab in user settings, the series name will now show the bookmarks and where the edit button is now a button to open series.
- Changed: Changed how we delete a bookmark so the logic is more precise
- Changed: On super small screens (less than 300px width), the server settings button will move under account dropdown
- Changed: Webtoon reader will now hide all images until the first load is complete to avoid any layout shifting
- Changed: Adjusted the rendering on reading list page to be better on smaller screens (and worse on larger ones). This will be addressed in UX release.
- Changed: Bookmark file management now happens in real time for deleting files and no longer runs on a scheduled task. 

# Fixed
- Fixed: Fixed an issue where the copy button wasn't working on the '3rd Party Clients' page.
- Fixed: Fixed an issue where a user who's logged in but has download role removed and then tries to download a file, Kavita would allow the download as the claim was still present in their JWT. 
- Fixed: In directory picker, clear the filter when you load a folder
- Fixed: EPUB parsing was missing setting the Day metadata tag
- Fixed: Fixed the spelling of 'On Going' to 'Ongoing'
- Fixed: Removed filtering button on Reading List page since it doesn't make sense
- Fixed: Fixed a bug where the events widget was visible for non-admins
- Fixed: Fixed an issue where the pagination would reset to 1 on the library-details page when refreshing or navigating directly to a specific page number.
- Fixed: When normalizing series names, if the series has no alphanumeric characters (ie non-roman), return the original name back rather than empty string.
- Fixed: When no user exists for an api key, throw a 401.
- Fixed: Fixed a bug where users couldn't select a collection (via add to collection) when they had a lot of them.
- Fixed: Fixed a bug where some covers would generate with green overlays and offsets due to an underlying bug in libjpeg 
- Fixed: Fixed a bug where marking a chapter as unread was actually marking it as read
- Fixed: Fixed a regression when reading epubs, the last page wouldn't count it as totally read
- Fixed: When deleting folders in bookmark cleanup, delete empty folders correctly
- Fixed: When a new library is created, some cards wouldn't show the library name due to caching. Reworked the caching code to be lighter on memory and fix the issue as well.
- Fixed: Fixed a bug where some book style overrides were duplicating. New logic targets only elements that really need the tag.
- Fixed: Fixed a bug where Filter options were not in alphabetical order
- Fixed: Don't throw an exception when we are trying to save progress and there is nothing new to save in manga reader 
- Fixed: Fixed a bug where checking for update did no validation if current version is already ahead/same before telling UI to open modal
- Fixed: Fixed a bug where sort options were not working on all series page
- Fixed: Fixed an issue when downloading bookmarks, sometimes a filename conflict can occur. Filenames will be prepended by chapter Id
- Fixed: Fixed a bug where typeahead wasn't filtering as you type and was allowing multiple selection of same item 
- Fixed: Fixed an issue where download event wasn't sending consistently when downloading files leaving the progress widget showing something downloading indefinitely
- Fixed: Fixed a bug in the webtoon reader where we were using the wrong dimensions for fitting images to screen 
- Fixed: Fixed a bug in Get Next/Prev Chapter, where chapters were not correctly between volumes and specials. They now behave correctly. The reading order is volumes -> Chapters -> Specials.
- Fixed: Fixed a bug where tap to paginate in the book reader would prevent the user from clicking the scrollbar

# Removed
- Removed: Account creation without emails is no longer possible.
- Removed: Users can no longer disable authentication on their servers. All non-admins can use the default password on the wiki or Github to change their password to one of their choosing.
- Removed: Removed old manual migrations (config change, cover images) that are from many releases ago. If you're updating from over 10 releases ago, you may loose cover images, you can update to v0.5 then to v0.5.1 to avoid that.

# External Services
- Stats: Stats services is now collecting two new fields to help collect data for future reworks. HasBookmarks to help understand how many people are using this functionality and NumberOfLibraries, which is the number of libraries. Remember, all data is anonymized and the Stat service is Open Source. Please continue to opt into data collection to help us build a better service for your needs.


therobbiedavis:
> Both are functions that were build out very early in the project
*Built

PR #1099, opened 2022-02-20
Scanner event hub fix

therobbiedavis:
# Fixed
- Fixed: Fixed an issue where the scanner would error when adding a new series because the series didn't have a library name yet. (develop)

### DIFF
--- a/API/Services/Tasks/ScannerService.cs
+++ b/API/Services/Tasks/ScannerService.cs
@@ -149,7 +149,7 @@ public class ScannerService : IScannerService
         try
         {
             await _eventHub.SendMessageAsync(MessageFactory.NotificationProgress, MessageFactory.LibraryScanProgressEvent(library.Name, ProgressEventType.Started, series.Name));
-            await UpdateSeries(series, parsedSeries, allPeople, allTags, allGenres, library.Type, library);
+            await UpdateSeries(series, parsedSeries, allPeople, allTags, allGenres, library);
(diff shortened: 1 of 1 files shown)


API/Services/Tasks/ScannerService.cs
```diff
@@ -430,7 +430,7 @@ private async Task UpdateLibrary(Library library, Dictionary<ParsedSeries, List<
             foreach (var series in librarySeries)
             {
                 await _eventHub.SendMessageAsync(MessageFactory.NotificationProgress, MessageFactory.LibraryScanProgressEvent(library.Name, ProgressEventType.Started, series.Name));
-                await UpdateSeries(series, parsedSeries, allPeople, allTags, allGenres, library.Type);
+                await UpdateSeries(series, parsedSeries, allPeople, allTags, allGenres, library.Type, library);
```
majora2007, line 433: If you are altering the signature, don't keep library type and library. Remove the type and get directly from library

majora2007 (changes requested):
Please update code to remove library.type and replace it with just a library instance.

API/Services/Tasks/ScannerService.cs
therobbiedavis: Fixed with [28bef4b](https://github.com/Kareadita/Kavita/pull/1099/commits/28bef4b4e2a5d1a375456761c8291a7b07f316ec). Resolving comment.

PR #1223, opened 2022-04-19
Fixes #1222

### LINKED ISSUE #1222: Checking for null after usage
In ```DefaultParser.cs``` 

```
              if ((ret.Volumes.Equals(Parser.DefaultVolume) || string.IsNullOrEmpty(ret.Volumes)) && !parsedVolume.Equals(Parser.DefaultVolume))
              {
                ret.Volumes = parsedVolume;
              }
              if ((ret.Chapters.Equals(Parser.DefaultChapter) || string.IsNullOrEmpty(ret.Chapters)) && !parsedChapter.Equals(Parser.DefaultChapter))
              {
                ret.Chapters = parsedChapter;
              }
```
For these 2 if statements ```ret.Volumes``` and ```ret.Chapters``` are being compared first and then checked if they have null values.
I think the ```IsNullOrEmpty``` check should be first. 
If it's OK for you I can create a pull request.

mihaibargau:
# Developer
- Fixed: fixed a bug where ```ret.Volumes``` and ```ret.Chapters``` are used before checking for null. (Fixes #1222 )

### DIFF
--- a/API/Parser/DefaultParser.cs
+++ b/API/Parser/DefaultParser.cs
@@ -1,4 +1,4 @@
-﻿using System.IO;
+using System.IO;
 using System.Linq;
 using API.Entities.Enums;
 using API.Services;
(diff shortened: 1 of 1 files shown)


majora2007:
When you create a PR, you need to follow the template. Please fix this. In your case, since it's not user facing, you should use the Developer heading.

majora2007:
Just updated as it's not a user facing bug, so we use Developer heading. Looks good now.

PR #1269, opened 2022-05-21
Fix for book cover extending across multiple pages

therobbiedavis:
# Fixed
- Fixed: Fixed an issue where book covers were not respecting height limiting styles, cause the image to spill across multiple pages in column layout


majora2007 (approved):
Looks good, this will be a great improvement

PR #1307, opened 2022-06-06
bugfix/user-preferences-accordion

magujun:
# Fixed
- Fixed: Fixed a bug in the user-preferences within the BookReader accordion panel where missing brackets for the `id` prevented the accordion from functioning correctly.


majora2007:
You need to stick to the template the other PRs do. Please reword your message to be more descriptive as it's aimed for the end user changelog.

PR #1321, opened 2022-06-14
dynamic height for series detail

therobbiedavis:
# Fixed
- Fixed: Fixed a bug that caused double scrollbars for series with longer titles

### DIFF
--- a/UI/Web/src/app/series-detail/series-detail.component.html
+++ b/UI/Web/src/app/series-detail/series-detail.component.html
@@ -52,7 +52,7 @@
 </app-side-nav-companion-bar>
 
 
-<div (scroll)="onScroll()" class="main-container container-fluid pt-2" *ngIf="series !== undefined" #scrollingBlock>
+<div (scroll)="onScroll()" [style.height]="getHeight()" class="main-container container-fluid pt-2" *ngIf="series !== undefined" #scrollingBlock>
     <div class="row mb-3 info-container">
         <div class="col-md-2 col-xs-4 col-sm-6 d-none d-sm-block">
             <app-image maxWidth="300px" [imageUrl]="seriesImage"></app-image>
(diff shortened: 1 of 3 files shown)


UI/Web/src/app/series-detail/series-detail.component.html
```diff
@@ -52,7 +52,7 @@ <h4 class="offcanvas-title" id="offcanvas-basic-title">Page Settings</h4>
 </app-side-nav-companion-bar>
 
 
-<div (scroll)="onScroll()" class="main-container container-fluid pt-2" *ngIf="series !== undefined" #scrollingBlock>
+<div (scroll)="onScroll()" [style.height]="getHeight()" class="main-container container-fluid pt-2" *ngIf="series !== undefined" #scrollingBlock>
```
majora2007, line 55: Main container has the #scrollingBlock, which means we can use a ViewChild to get this on load and keep a reference to it. 

Also, you should use [ngStyle]="{'height': getHeight()}"

UI/Web/src/app/series-detail/series-detail.component.ts
```diff
@@ -261,6 +261,12 @@ export class SeriesDetailComponent implements OnInit, OnDestroy {
 		}
 	}
 
+	getHeight(): string {
```
majora2007, line 264: Instead of a getHeight, let's use a getter method:

get ScrollingBlockHeight() {
  return string here;
}

And like I mentioned in other comment, let's use a viewchild.

@ViewChild('scrollingBlock') scrollingBlockElement!: ElementRef<HtmlDivElement>;

When using viewchild, it will be selected on ngAfterViewInit, so you should always do a check if it's set.

majora2007 (changes requested):

majora2007 (approved):

PR #1332, opened 2022-06-26
Add KnownProxies configuration

tjarls:
# Added
- Added: Allow appending the ForwardedHeadersOptions property KnownProxies from appsettings.json. By default that property is set to just 127.0.0.1. Definiting additional ones allows trusting x-forwarded headers from a non local proxy, for example one running on the local router, and thus the proper detection of the protocol to use in user invites in such setups.

# Changed
- Changed: Removed redundant definition ForwardedHeadersOptions configuration instance.
- Changed: Removed options from UseForwardedHeaders() and let it be set by the configuration instance.

### DIFF
--- a/API/Startup.cs
+++ b/API/Startup.cs
@@ -53,8 +53,10 @@ namespace API
             services.AddControllers();
             services.Configure<ForwardedHeadersOptions>(options =>
             {
-                options.ForwardedHeaders =
-                    ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+                options.ForwardedHeaders = ForwardedHeaders.All;
+                foreach(var proxy in _config.GetSection("KnownProxies").AsEnumerable().Where(c => c.Value != null)) {
+                    options.KnownProxies.Add(IPAddress.Parse(proxy.Value));
+                }
             });
             services.AddCors();
             services.AddIdentityServices(_config);
@@ -112,13 +114,6 @@ namespace API
             });
 
             services.AddResponseCaching();
-
-            services.Configure<ForwardedHeadersOptions>(options =>
-            {
-                options.ForwardedHeaders =
-                    ForwardedHeaders.All;
-            });
-
 
             services.AddHangfire(configuration => configuration
                 .UseSimpleAssemblyNameTypeSerializer()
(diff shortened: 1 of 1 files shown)


API/Startup.cs
```diff
@@ -53,8 +53,10 @@ public void ConfigureServices(IServiceCollection services)
             services.AddControllers();
             services.Configure<ForwardedHeadersOptions>(options =>
             {
-                options.ForwardedHeaders =
-                    ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+                options.ForwardedHeaders = ForwardedHeaders.All;
+                foreach(var proxy in _config.GetSection("KnownProxies").AsEnumerable().Where(c => c.Value != null)) {
```
majora2007, line 57: Will this code fail if KnowProfixes isn't set in the appsettings.json?

tjarls: It works when the section "KnownProxies" is missing. GetSection() returns an empty IConfigurationSection and enventually an empty enumerable, so no action is then taken.

tjarls: Also, I tested that case, just to be sure.

majora2007:
Do the users need to have KnownProxies set to get the benefits of the forwarded headers? 

I want to make sure we arent removing any functionality from users, since this is a pretty neiche add.

tjarls:
> Do the users need to have KnownProxies set to get the benefits of the forwarded headers?
> 
> I want to make sure we arent removing any functionality from users, since this is a pretty neiche add.

The current experience of users relying on forwarded headers is unchanged. That functionality is currently limited to local proxies because the property 'KnownProxies' in 'ForwardedHeadersOptions' which is used to config the ForwardedHeaders middleware is initialized as a list with just the loopback address. The proposed new setting allows to append additional IP addresses to that list, still keeping the 127.0.0.1 address. By doing so, the app will trust the forwarded headers in traffic originating coming from any additional IP thus set, in addition to trusting them in traffic that originated locally as before.

So if the new proposed app setting is not set or empty then the middleware property would be unchanged and set just as before:
```
["127.0.0.1"]
```

If, for example, one uses:
```
"KnownProxies": [ "192.168.1.1" ]
```
Then the middleware property becomes:
```
["127.0.0.1", "192.168.1.1"]
```

PR #1349, opened 2022-07-01
Bugfix for overflow on long filenames with no spaces

### LINKED ISSUE #403: Bug: Long filenames with no spaces are causing overflow on Info/Edit modal for Volume.
Potentially use the readmore component for thsi?

magujun:
# Fixed
- Fixed: Added word breaking for long filenames with no spaces on headers, titles, and lists which would cause overflow.
- Fixed: Titles on many screens and in modals will now word break if they are too long.  (Closes #403)

### DIFF
--- a/UI/Web/src/styles.scss
+++ b/UI/Web/src/styles.scss
@@ -70,7 +70,7 @@ app-root {
   scrollbar-color: var(--primary-color-scrollbar);
 }
 
-.justify-content-between, .modal-title {
+.justify-content-between, .modal-title, .list-group {
   word-break: break-all;
 }
 
--- a/UI/Web/src/styles.scss
+++ b/UI/Web/src/styles.scss
@@ -70,10 +70,6 @@ app-root {
   scrollbar-color: var(--primary-color-scrollbar);
 }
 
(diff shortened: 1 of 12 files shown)


UI/Web/src/styles.scss
```diff
@@ -70,6 +70,10 @@ app-root {
   scrollbar-color: var(--primary-color-scrollbar);
 }
 
+.justify-content-between, .modal-title {
```
majora2007, line 73: Don't add styles to the global css. This needs to be copied to the places where it's used.

majora2007 (changes requested):

UI/Web/src/styles.scss
magujun: Oh! I've just read this recommendation...
So with this fix we deal with the issue more broadly without having to add the style to all places individually. It shouldn't break anything else because it's just 'breaking' long strings as needed...
If you say it's not the way to go, I'll look into some other fix.

majora2007: Applying on .justify-content-between has unintended consequences. Please duplicate the style for now.

PR #1377, opened 2022-07-16
Parser improvements

tjarls:
# Changed
- Changed: Replace underscore with spaces prior to parsing to simplify file name regex
- Changed: Reuse some common regex subexpressions
- Changed: Merge some regex

### DIFF
--- a/API/Parser/Parser.cs
+++ b/API/Parser/Parser.cs
@@ -201,11 +201,11 @@ namespace API.Parser
         {
             // Tintin - T22 Vol 714 pour Sydney
             new Regex(
-                @"(?<Series>.+?)\s?(\b|_|-)\s?((vol|tome|t)\.?)(?<Volume>\d+(-\d+)?)",
+                @"(?<Series>.+?)\s?(\b|-)\s?(vol|tome|t)\.?\d",
                 MatchOptions, RegexTimeout),
             // Invincible Vol 01 Family matters (2005) (Digital)
             new Regex(
-                @"(?<Series>.+?)(\b|_)((vol|tome|t)\.?)(\s|_)(?<Volume>\d+(-\d+)?)",
+                @"(?<Series>.+?)\b(vol|tome|t)\.?\s\d",
                 MatchOptions, RegexTimeout),
             // Batman Beyond 2.0 001 (2013)
             new Regex(
@@ -213,7 +213,7 @@ namespace API.Parser
                 MatchOptions, RegexTimeout),
             // 04 - Asterix the Gladiator (1964) (Digital-Empire) (WebP by Doc MaKS)
             new Regex(
-            @"^(?<Volume>\d+)\s(-\s|_)(?<Series>.*(\d{4})?)( |_)(\(|\d+)",
+            @"^(?<Volume>\d+)\s(-\s)(?<Series>.*(\d{4})?) (\(|\d)",
                 MatchOptions, RegexTimeout),
             // 01 Spider-Man & Wolverine 01.cbr
             new Regex(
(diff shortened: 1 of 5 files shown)


API/Parser/Parser.cs
```diff
@@ -857,8 +753,6 @@ public static string CleanTitle(string title, bool isComic = false)
 
             title = RemoveEditionTagHolders(title);
 
-            title = isComic ? RemoveComicSpecialTags(title) : RemoveMangaSpecialTags(title);
```
tjarls, line 860: Duplicate code. It gets done in the next condition

API/Parser/Parser.cs
```diff
@@ -64,43 +64,43 @@ public static class Parser
         /// <summary>
         /// Recognizes the Special token only
         /// </summary>
-        private static readonly Regex SpecialTokenRegex = new Regex(@"SP\d+",
+        private static readonly Regex SpecialTokenRegex = new Regex(@"(SP|HS)\d+",
             MatchOptions, RegexTimeout);
 
+        private const string seriesExpr = $@"(?<Series>{balanced}?)";
```
majora2007, line 70: You need to use proper naming convention. Constant variables need to start with a capital letter in C#.

API/Parser/Parser.cs
```diff
@@ -64,43 +64,43 @@ public static class Parser
...
+        private const string chapterPrefixes = @"(c|chapter[-\s]+|issue\s?#?|#|(n°|n\.|Ep\.?)[-\s]*)";
+
 
         private static readonly Regex[] MangaVolumeRegex = new[]
         {
-            // Dance in the Vampire Bund v16-17
+            // vol_001-1.cbz for MangaPy default naming convention
             new Regex(
-                @"(?<Series>.*)(\b|_)v(?<Volume>\d+-?\d+)( |_)",
+                @"^(vol\s)(?<Volume>\d+(\.\d)?)-\d+$",
```
majora2007, line 87: I'm not sure about the end of string, $, as usually these examples have other tags and strings. Dance in the Vampire Bund {Special Edition} v01-02 (2019) (Digital) (KG Manga).cbz is an example of a very common use case.

API/Parser/Parser.cs
```diff
@@ -136,213 +136,121 @@ public static class Parser
         {
             // Grand Blue Dreaming - SP02
             new Regex(
-                @"(?<Series>.*)(\b|_|-|\s)(?:sp)\d",
+                $@"{seriesExpr}(\b|-|\s)(?:sp)\d",
                 MatchOptions, RegexTimeout),
-            // [SugoiSugoi]_NEEDLESS_Vol.2_-_Disk_The_Informant_5_[ENG].rar, Yuusha Ga Shinda! - Vol.tbd Chapter 27.001 V2 Infection ①.cbz
```
majora2007, line 141: Please make sure this case is covered in the code
[SugoiSugoi]_NEEDLESS_Vol.2_-_Disk_The_Informant_5_[ENG].rar

API/Parser/Parser.cs
```diff
@@ -136,213 +136,121 @@ public static class Parser
...
+                $@"^{seriesExpr} Vol\.?tbd\b",
                 MatchOptions, RegexTimeout),
+            // [SugoiSugoi] NEEDLESS Vol.2 - Disk The Informant 5 [ENG].rar
             // Mad Chimera World - Volume 005 - Chapter 026.cbz (couldn't figure out how to get Volume negative lookaround working on below regex),
             // The Duke of Death and His Black Maid - Vol. 04 Ch. 054.5 - V4 Omake
-            new Regex(
-                @"(?<Series>.+?)(\s|_|-)+(?:Vol(ume|\.)?(\s|_|-)+\d+)(\s|_|-)+(?:(Ch|Chapter|Ch)\.?)(\s|_|-)+(?<Chapter>\d+)",
-                MatchOptions,
-                RegexTimeout),
-            // Ichiban_Ushiro_no_Daimaou_v04_ch34_[VISCANS].zip, VanDread-v01-c01.zip
```
majora2007, line 151: Same here, multiple cases removed, I don't see a comment for their case. Please ensure there is a unit test matching all these cases as well.

majora2007: Especially since I don't see any MangaParserTests in this PR.

majora2007 (changes requested):
I haven't pulled the code down, this is just a preliminary look. 

Looks dangerous due to the amount of changes and changing how I debug my regex (aka it's not easy to copy and paste the regex into regex101). 

I'll pull this down next week and give it a through shakeout against my library.

API/Parser/Parser.cs
tjarls: I have not added new patterns to the manga regex, hence why there's no addition to MangaParserTests.
Wasn't the coverage good enough before?

API/Parser/Parser.cs
tjarls: That's covered by the regex at line 147. Please see comment at line 145.

API/Parser/Parser.cs
tjarls: The cases matching regex that have been removed are matched by other regex that have become more generic. I have moved the the comments describing use cases to the new regex in order to track where they are being matched now.

API/Parser/Parser.cs
tjarls: Done!

API/Parser/Parser.cs
tjarls: That's the exact naming convention of MangaPy. I wanted to make sure that was as narrow as possible.

It is unrelated to "Dance in the Vampire Bund v16-17". That example is not an edge case. It's a pretty standard format where the volume explicitely marked with the prefix v. Hence it matches:
<series names><volume prefix><volume number/range>

API/Parser/Parser.cs
```diff
@@ -67,40 +67,39 @@ public static class Parser
...
+        private const string SeriesExpr = $@"(?<Series>{Balanced}?)";
+        private const string VolumeExpr = @"(?<Volume>\d+(\.\d)?(-\d+(\.\d)?)?)";
+        private const string ChapterExpr = @"(?<Chapter>\d+(\.\d)?(-\d+(\.\d)?)?)";
+        private const string BalancedParen = @"(?:[^()]|(?<open>\()|(?<-open>\)))*?(?(open)(?!))";
+        // non greedy matching of a string where parenthesis and brackets are balanced
+        private const string Balanced = @"(?:[^()\[\]]|(?<open>\()|(?<-open>\))|(?<openbracket>\[)|(?<-openbracket>\]))*?(?(openbracket)(?!))(?(open)(?!))";
+        // French Volume: tome, t, t. -> volume
+        private const string VolumePrefixes = @"(v|t|(v\.|vol\.?|volume|t\.|tome)\s?)";
+        // n°,n. -> numero sign
+        private const string chapterPrefixes = @"(c|chapter[-\s]+|issue\s?#?|#|(n°|n\.|Ep\.?)[-\s]*)";
```
majora2007, line 79: This needs PascalCase

API/Parser/Parser.cs
```diff
@@ -67,40 +67,39 @@ public static class Parser
         private static readonly Regex SpecialTokenRegex = new Regex(@"SP\d+",
             MatchOptions, RegexTimeout);
 
+        private const string SeriesExpr = $@"(?<Series>{Balanced}?)";
+        private const string VolumeExpr = @"(?<Volume>\d+(\.\d)?(-\d+(\.\d)?)?)";
+        private const string ChapterExpr = @"(?<Chapter>\d+(\.\d)?(-\d+(\.\d)?)?)";
+        private const string BalancedParen = @"(?:[^()]|(?<open>\()|(?<-open>\)))*?(?(open)(?!))";
```
majora2007, line 73: Can you explain what open here means and why we use it?

API/Parser/Parser.cs
```diff
@@ -1074,5 +960,9 @@ public static bool HasComicInfoSpecial(string comicInfoFormat)
         {
             return FormatTagSpecialKeywords.Contains(comicInfoFormat);
         }
+
+        private static string replaceUnderscores(string filename) {
```
majora2007, line 964: Should be PascalCase matching C# style

majora2007:
I'm testing this out, I see unit tests are failing. Please ensure all are passing.

API/Parser/Parser.cs
majora2007, line 79: I think it would be beneficial to have unit tests for these, so when adding a new case, it can be easily tested without worrying about breakage.

API/Parser/Parser.cs
majora2007: I'm not getting it. When I test it out, (?:[^()\[\]]|(?<open>\()|(?<-open>\))|(?<openbracket>\[)|(?<-openbracket>\])) is worse that just using . for Series group matching. I don't understand the regex nor the benefit of adding such complexity.

API/Parser/Parser.cs
tjarls: They are strings that are reused in regex. This is meant to ensure that regex can reuse the same format for a chapter, etc.  Just like the regex themselves get tested with the parsing functions, these pieces of strings get tested with them.

PR #1431, opened 2022-08-13
Gestures (swipe / manga reader)

magujun:
# Added

- Added: Added the ability to swipe left or right to change pages (manga reader).

### DIFF
--- a/UI/Web/src/app/_services/gestures.service.ts
+++ b/UI/Web/src/app/_services/gestures.service.ts
@@ -1,0 +1,23 @@
+let swipeCoord = [0, 0];
+let swipeTime = new Date().getTime();
+
+export function swipe(e: TouchEvent, when: string): string | undefined {
+
+    const coord: [number, number] = [e.changedTouches[0].clientX, e.changedTouches[0].clientY];
+    const time = new Date().getTime();
+
+    if (when === 'start') {
+        swipeCoord = coord;
+        swipeTime = time;
+    } else if (when === 'end') {
+        const direction = [coord[0] - swipeCoord[0], coord[1] - swipeCoord[1]];
+        const duration = time - swipeTime;
+
+        if (duration < 1000 //
+            && Math.abs(direction[0]) > 30 // Long enough
+            && Math.abs(direction[0]) > Math.abs(direction[1] * 3)) { // Horizontal enough
+            return (direction[0] < 0 ? 'right' : 'left');
+        }
+    }
+    return;
+}
--- a/UI/Web/src/app/manga-reader/manga-reader.component.html
+++ b/UI/Web/src/app/manga-reader/manga-reader.component.html
@@ -33,66 +33,68 @@
         </div>
     </ng-container>
 
-    <div (click)="toggleMenu()" class="reading-area" 
-            [ngStyle]="{'background-color': backgroundColor, 'height': readerMode === ReaderMode.Webtoon ? 'inherit' : 'calc(var(--vh)*100)'}" #readingArea>
-        <ng-container *ngIf="readerMode !== ReaderMode.Webtoon; else webtoon">
-            <div class="image-container" [ngClass]="{'d-none': !renderWithCanvas }" [style.filter]="'brightness(' + generalSettingsForm.get('darkness')?.value + '%)'">
-                <canvas #content class="{{getFittingOptionClass()}}"
-                    ondragstart="return false;" onselectstart="return false;">
-                </canvas>
-            </div>
+    <div (touchstart)="onSwipe($event, 'start')" (touchend)="onSwipe($event, 'end')">
+        <div (click)="toggleMenu()" class="reading-area" 
+                [ngStyle]="{'background-color': backgroundColor, 'height': readerMode === ReaderMode.Webtoon ? 'inherit' : 'calc(var(--vh)*100)'}" #readingArea>
+            
+                <ng-container *ngIf="readerMode !== ReaderMode.Webtoon; else webtoon">
+                <div class="image-container" [ngClass]="{'d-none': !renderWithCanvas }" [style.filter]="'brightness(' + generalSettingsForm.get('darkness')?.value + '%)'">
+                    <canvas #content class="{{getFittingOptionClass()}}"
+                        ondragstart="return false;" onselectstart="return false;">
+                    </canvas>
+                </div>
 
-            <div class="pagination-area">
-                <!-- Pagination controls and screen hints-->
-                <div class="{{readerMode === ReaderMode.LeftRight ? 'left' : 'top'}} {{clickOverlayClass('left')}}" (click)="handlePageChange($event, 'left')" 
-                    [ngStyle]="{'height': (readerMode === ReaderMode.LeftRight ? ImageHeight: '25%')}">
-                    <div *ngIf="showClickOverlay">
-                        <i class="fa fa-angle-{{readingDirection === ReadingDirection.RightToLeft ? 'double-' : ''}}{{readerMode === ReaderMode.LeftRight ? 'left' : 'up'}}"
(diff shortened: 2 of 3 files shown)


UI/Web/src/app/manga-reader/manga-reader.component.html
```diff
@@ -33,66 +33,68 @@
...
 
-    <div (click)="toggleMenu()" class="reading-area" 
-            [ngStyle]="{'background-color': backgroundColor, 'height': readerMode === ReaderMode.Webtoon ? 'inherit' : 'calc(var(--vh)*100)'}" #readingArea>
-        <ng-container *ngIf="readerMode !== ReaderMode.Webtoon; else webtoon">
-            <div class="image-container" [ngClass]="{'d-none': !renderWithCanvas }" [style.filter]="'brightness(' + generalSettingsForm.get('darkness')?.value + '%)'">
-                <canvas #content class="{{getFittingOptionClass()}}"
-                    ondragstart="return false;" onselectstart="return false;">
-                </canvas>
-            </div>
+    <div (touchstart)="onSwipe($event, 'start')" (touchend)="onSwipe($event, 'end')">
```
majora2007, line 36: Why do we need another div layer? Will this work on fullscreen since we fullscreen on the readingArea?

UI/Web/src/app/_services/gestures.service.ts
```diff
@@ -0,0 +1,23 @@
+let swipeCoord = [0, 0];
```
majora2007, line 1: This shouldn't be gestures.service as that is for an angular service. This swipe method should be exposed under Utility service.

UI/Web/src/app/_services/gestures.service.ts
```diff
@@ -0,0 +1,23 @@
...
+let swipeTime = new Date().getTime();
+
+export function swipe(e: TouchEvent, when: string): string | undefined {
```
majora2007, line 4: Make when type: 'start' | 'end'

UI/Web/src/app/_services/gestures.service.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    const coord: [number, number] = [e.changedTouches[0].clientX, e.changedTouches[0].clientY];
```
majora2007, line 6: This needs index check to ensure there is at least 1 event (even if always there, better to write safe code than assume)

UI/Web/src/app/_services/gestures.service.ts
```diff
@@ -0,0 +1,23 @@
...
+    const time = new Date().getTime();
+
+    if (when === 'start') {
+        swipeCoord = coord;
+        swipeTime = time;
+    } else if (when === 'end') {
+        const direction = [coord[0] - swipeCoord[0], coord[1] - swipeCoord[1]];
+        const duration = time - swipeTime;
+
+        if (duration < 1000 //
```
majora2007, line 16: Does this swiping handle up/down to handle the up/down mode of the manga reader?

majora2007 (commented):
I left some code comments for you. I would need to test this out myself, before any merging. It seems like you've only handled one direction and for merge, I would want a more robust solution. 

CSS Image transitions is something also that should be experimented with on this PR. I'm not sure what benefit swiping adds without the css transition.

UI/Web/src/app/manga-reader/manga-reader.component.html
magujun: Wrapping the whole reading area with the event handler seemed to be more effective and less intrusive. Tested working on fullscreen.

UI/Web/src/app/_services/gestures.service.ts
magujun: Changed parameter to a strong, strict, static type.

UI/Web/src/app/_services/gestures.service.ts
magujun: Added guard clause check for e.changedTouches.length < 1

UI/Web/src/app/_services/gestures.service.ts
magujun: Right now it strictly handles horizontal left/right (prev/next?).
What is the up/down (top/bottom?) mode and how does it work?

UI/Web/src/app/_services/gestures.service.ts
magujun: Refactored to expose swipe function under utility.service

magujun:
Swiping (at least left/right) adds a more natural feel to the UX.
We could add some css image transition but choosing one can be subjective, the main purpose for now is functionality as per [feature request](https://feats.kavitareader.com/posts/120/swipe-to-change-page). Better than a simple slide transition would be implementing a solution similar to [turn.js](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwjg-Mrf08X5AhWGjYkEHZUlAnwQFnoECAIQAQ&url=http%3A%2F%2Fwww.turnjs.com%2F&usg=AOvVaw1NNs0lmYytQgufZY5XqCHc) (needs investigating, though).

UI/Web/src/app/_services/gestures.service.ts
majora2007: Click the reading mode button once on the menu and you'll see. We need to ensure this doesn't run when we are in webtoon mode as well.

magujun: Right now it's tested working with up/down reading mode, but still using the horizontal axis.
I believe we can extend the coordinate logic to switch to vertical when on this reading mode.

Webtoon scrolling and functionality doesn't seem to be affected by this, even though page numbers do update in the slider.
I'll add a check for webtoon mode before enabling the swipe gesture.

majora2007: I think it would be better to return early on the handler if webtoon mode is active.

magujun:
I thought that doing the check within the div tag would be better as it
wouldn't even fire the onSwipe handler. ;)

UI/Web/src/app/_services/gestures.service.ts
majora2007: You replied in a comment rather than on this thread, but the reason why doing it in the handler is that the code is much cleaner than loading up the template. You can already see the complexity that is added due to the template being overloaded with complex logic.

UI/Web/src/app/_services/utility.service.ts
```diff
@@ -0,0 +1,23 @@
+let swipeCoord = [0, 0];
+let swipeTime = new Date().getTime();
+
+export function swipe(e: TouchEvent, when: 'start' | 'end'): ('right' | 'left') | undefined {
```
majora2007, line 4: You misunderstood me when I said put this on the Utility serivce. I expect this detectSwipeDirection would exist as a method on the Utility Service. Not an exposed function, but should be used and exposed like all other methods on the UtilityService.

majora2007: See bulk-selection Service  for an example of how services are used in regards to events like this.

majora2007:
Can you validate this doesn't fire a page change when the user pinch zooms on mobile and then swipes to navigate over the image?

magujun:
https://user-images.githubusercontent.com/75567460/185685750-aaba42ed-d637-41e7-9caf-9fa0f490af02.mp4

No issues with the swipe gesture.

There's one thing though, not related to the gesture, as the next and previous page *clicking* zones seem to get enlarged with the image and end up taking a lot of the screen space and getting triggered inadvertently. You might want to validate that.

majora2007:
I have plans to implement this, however I'm going a different approach. I'm going to close this out for now. Thanks for allowing me the insight into a simple swipe without animations.

PR #1444, opened 2022-08-17
Collation tests (Discussion)

tjarls:
These are only experimental changes submitted for discussion

# Changed
- Changed: Sort series ignoring case, diacritics, kana (styles)


majora2007:
Doesn't seem like there was much discussion on this after this PR. I'm going to close it. If users want this, a feature hub can be created and upvoted.

Issue #933, opened 2022-01-13
Some Covers are writing with distoritons


majora2007:
Bleach Digital Colored has some jpeg images writing incorrectly.

The issue can be read about here: https://github.com/kleisauke/net-vips/issues/155

This requires a NetVips (issue is with libvips) update.

PR #1479, opened 2022-08-25
Fixes Cover Image issues on older CPUs

### LINKED ISSUE #1423: Docker - Covers will not apply (automatically or manually
Using docker image with version 0.5.4.0. Host system is Manjaro Linux (Arch base) on an older home server.

Installed with:

docker run --name kavita -p 5000:5000      -v /home/gyjo/Storage4TB/Manga/:/manga     -v /home/gyjo/Storage4TB/Manga/kavitaconfig/:/kavita/config     --restart unless-stopped     -d kizaing/kavita:latest

Everything works fine except every time I run "Refresh Covers" it hangs at 0% completion, and if I try to click on a page like "Home" or a library, it seems to crash the docker container and restart it. Covers don't get applied. Manually attempting to add a cover also crashes the container. All other settings, logs, cache seems to write to the disk normally. 

I have tried the following:
- reinstalling with clean config folder
- restarting docker containers
- restarting server
- chmod -R 777 config folder
- chmod -R 777 Manga folder
- changing location of config folder to different drive

No changes

Kizaing:
# Fixed
- Fixed: Fixed issue where users on older CPUs could not generate or change cover images. (Fixes #1423)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,12 @@ COPY --from=copytask /files/wwwroot /kavita/wwwroot
 
 #Installs program dependencies
 RUN apt-get update \
-  && apt-get install -y libicu-dev libssl1.1 libgdiplus curl\
+  && apt-get install -y libicu-dev libssl1.1 libgdiplus curl \
+  && apt-get install -y libvips --no-install-recommends \
   && rm -rf /var/lib/apt/lists/*
+
+#Removes the libvips.so.42 file to fix the AVX CPU requirement issue
+RUN rm /kavita/libvips.so.42
 
 COPY entrypoint.sh /entrypoint.sh
 


majora2007 (approved):

kleisauke:
Note that I cannot recommend this, since it will downgrade libvips to 8.9.1, causing issue #933 to occur again. Ubuntu's APT packages also contain untrusted loaders, and `NetVips.BlockUntrusted` will not work in this case.

BTW, the comment mentions that AVX is a hard requirement, which is incorrect. The next release of `NetVips.Native` will provide prebuilt binaries for Linux that require the x64 v2 microarchitecture, which is approximately equivalent to the Intel Nehalem family of CPUs (slightly older than Westmere).

majora2007:
Thanks for the comment Klies, I think we will build a separate image for users with this problem as a temporary measure so they can at least have cover images, while our mainline will use NetVips.Native out of the box. 

I will sync back with our users on the next release of NetVips.Native and see if it will work for them.

prosthetichead:
this was still an issue for me on an AMD Phenom(tm) II X6 1090T Processor.
I used the work around here https://github.com/Kareadita/Kavita/issues/1423 which fixed it for me but of course will need to be reapplied after any Docker pulls.

The issue does not just affect covers also, Manga and Comics that are made up of images (not a PDF) will not open and crashes the docker image unless the work around is applied also just the same as cover generation does. 

Happy to test anything you guys need on my system.

kleisauke:
> AMD Phenom II X6 1090T Processor

This CPU is from 2010 and the prebuilt Linux binaries from NetVips require the x64 v2 microarchitecture, which is approximately equivalent to the Intel Nehalem family of CPUs.

PR #1487, opened 2022-08-27
Sort series by chapter number only when some chapters have no volume

### LINKED ISSUE #1486: Incorrect chapter sorting
**If this is a feature request, request [here](https://feats.kavitareader.com/) instead. Feature requests will be deleted from Github.**

Please put as much information as possible to help me understand your issue. OS, browser, version are very important!

**Describe the bug**
Chapter sorting doesn't sort from lowest number to highest number

**To Reproduce**
1. Have Manga library with files with the naming in the pastebin https://pastebin.com/1ZA54Cdw
2. Scan library
3. Open series and click on Chapters
4. See incorrect sorting

**Expected behavior**
Chapters should be sorted from lowest to highest

**Screenshots**
image

 - OS: Docker
 - Browser Brave
 - Version 0.5.5.15

**Smartphone (please complete the following information):**

**Additional context**

tjarls:
# Fixed
- Fixed: Sort series by chapter number only when some chapters do not include a volume number (Fixes #1486)

### DIFF
--- a/API/Services/SeriesService.cs
+++ b/API/Services/SeriesService.cs
@@ -493,7 +493,7 @@ public class SeriesService : ISeriesService
             if (v.Number == 0) return c;
             c.VolumeTitle = v.Name;
             return c;
-        }).OrderBy(c => float.Parse(c.Number), ChapterSortComparer.Default));
+        }).OrderBy(c => float.Parse(c.Number), ChapterSortComparer.Default)).ToList();
 
         foreach (var chapter in chapters)
         {
@@ -518,7 +518,8 @@ public class SeriesService : ISeriesService
         var storylineChapters = volumes
             .Where(v => v.Number == 0)
             .SelectMany(v => v.Chapters.Where(c => !c.IsSpecial))
-            .OrderBy(c => float.Parse(c.Number), ChapterSortComparer.Default);
+            .OrderBy(c => float.Parse(c.Number), ChapterSortComparer.Default)
+            .ToList();
 
         // When there's chapters without a volume number revert to chapter sorting only as opposed to volume then chapter
         if (storylineChapters.Any()) {
--- a/API/Services/SeriesService.cs
+++ b/API/Services/SeriesService.cs
@@ -50,8 +50,8 @@ public class SeriesService : ISeriesService
     /// <returns></returns>
     public static Chapter GetFirstChapterForMetadata(Series series, bool isBookLibrary)
     {
-        return series.Volumes.OrderBy(v => v.Number, new ChapterSortComparer())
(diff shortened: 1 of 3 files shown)


API/Services/SeriesService.cs
```diff
@@ -520,6 +520,11 @@ public async Task<SeriesDetailDto> GetSeriesDetail(int seriesId, int userId)
             .SelectMany(v => v.Chapters.Where(c => !c.IsSpecial))
             .OrderBy(c => float.Parse(c.Number), new ChapterSortComparer());
 
+        // When there's chapters without a volume number revert to chapter sorting only as opposed to volume then chapter
+        if (storylineChapters.Any()) {
+            retChapters = retChapters.OrderBy(c => float.Parse(c.Number), new ChapterSortComparer());
```
majora2007, line 525: Let's move the new comparator to a private member to avoid recreating each API call.

tjarls: That's a good idea. We can even go further and implement the Comparer Default static property pattern.

majora2007:
Testing this now

majora2007:
This is not working

image

API/Services/SeriesService.cs
```diff
@@ -493,7 +493,7 @@ public async Task<SeriesDetailDto> GetSeriesDetail(int seriesId, int userId)
             if (v.Number == 0) return c;
             c.VolumeTitle = v.Name;
             return c;
-        }).OrderBy(c => float.Parse(c.Number), new ChapterSortComparer()));
+        }).OrderBy(c => float.Parse(c.Number), ChapterSortComparer.Default));
```
majora2007, line 496: You need .ToList() on this, as it will be enumerated on multiple times.

API/Services/SeriesService.cs
```diff
@@ -518,7 +518,12 @@ public async Task<SeriesDetailDto> GetSeriesDetail(int seriesId, int userId)
         var storylineChapters = volumes
             .Where(v => v.Number == 0)
             .SelectMany(v => v.Chapters.Where(c => !c.IsSpecial))
-            .OrderBy(c => float.Parse(c.Number), new ChapterSortComparer());
+            .OrderBy(c => float.Parse(c.Number), ChapterSortComparer.Default);
```
majora2007, line 521: You need .ToList() on this, as it will be enumerated on multiple times.

majora2007 (changes requested):

majora2007:
Please add the 2 comments then I can merge this.

tjarls:
Ah! Right! The queries gets evaluated multiple times overwise. How about the test case you showed before? That wouldn't the failure to sort you observed, would it?

majora2007:
The failure was not real. It was due to checking out old code. So it's good to go.

PR #1502, opened 2022-09-02
Tachiyomi Tracking Bugfix

ThePromidius:
# Changed
Changed: userHasProgress now properly returns True when PagesRead = SeriesPages


ThePromidius:
Fixed in cae072a192d05662926d38d04f2498678a920a43

PR #1511, opened 2022-09-03
Metadata Bugfixes

### LINKED ISSUE #1354: Series inferred from epub books are not sorted correctly
I have noticed that series inferred from epub metedata use the wrong "Sort Name" and hence are often incorrectly sorted in the library view:
image

This is due to some logic in the scanner task that fills SeriesSort with TitleSort even for books that are part of series as opposed to one shot books where the title is promoted to series name. It'd be more accurate that, when an epub metadata provides an element with a series name, to use it as SeriesSort as well. As far as I can tell, calibre does not have a sorting key for series name, so there seems to be no property "file-as" associated with the "calibre:series" element.

### LINKED ISSUE #1510: Problem with Metada (Again) V3
Hi with the new version Kavita 0.5.6, in spanish when i scan the library have a lot of FAIL like this:
 
> fail: API.Services.Tasks.Scanner.ProcessSeries[0]
      There was some issue when updating chapter's metadata
      System.FormatException: String '10/19/2021' was not recognized as a valid DateTime.
         at API.Services.Tasks.Scanner.ProcessSeries.UpdateChapterFromComicInfo(Chapter chapter, ComicInfo info) in C:\Users\josep\Documents\Projects\KavitaOrg\Kavita\API\Services\Tasks\Scanner\ProcessSeries.cs:line 0

When in windows i put regional format in English (EEUU) this issue not show and metadata are well

Maybe for this in spanish some metadata is not showed, in USA the time is set to MONTH/DAY/YEAR but in spain by default is DAY/MONTH/YEAR.

The previus issue #1411 

Pd: other thing i was find but not know if is a bug, if in library you select a folder with only the epubs inside it is not recognised, but if you put it in a folder work well

### LINKED ISSUE #1512: XML deserialization exception
Some ComicInfo.xml content triggers exceptions during the deserialization. In two cases I investigated I have found that it was being triggered by an empty Count or Day element.
For example in the following sequence, the empty Count element triggers the bug, because it deserializes into a type `int` property and the deserialization has no default value to use, while the similarly empty element Number works fine because its target property is of type string.
```
<ComicInfo>
  <Title>The Right Hand of Doom</Title>
  <Series>Hellboy</Series>
  <Number />
  <Count />
  <Volume>4</Volume>
```
In order to fix this, all elements that can be empty (minOccurs="0") and that deserialize into a type int property should have a default value.

tjarls:
# Changed
- Changed: When parsing epub files, ensure SeriesSort tag is used when defined (Fixes #1354)
# Fixed
- Fixed: Fixed an assumption that the environment always uses US time format which could cause ReleaseYear parsing of ComicInfo to fail. (Fixes #1510)
- Fixed: Fixed a ComicInfo XML de-serialization bug with empty elements of xs:int type into ints (Fixes #1512)


API/Services/BookService.cs
```diff
@@ -609,15 +613,10 @@ public ParserInfo ParseInfo(string filePath)
                             FullFilePath = filePath,
                             IsSpecial = false,
                             Series = series.Trim(),
+                            SeriesSort = series.Trim(),
                             Volumes = seriesIndex
                         };
 
-                        // Don't set titleSort if the book belongs to a group
```
majora2007, line 615: Can you help me understand the change here?

API/Data/Metadata/ComicInfo.cs
```diff
@@ -54,6 +62,8 @@ public class ComicInfo
         public string StoryArc { get; set; } = string.Empty;
         public string SeriesGroup { get; set; } = string.Empty;
         public string AlternateNumber { get; set; } = string.Empty;
+        // This should really be -1 according to anansi-project spec
```
majora2007, line 65: Please remove these comments. I understand the spec, but there is no need to have -1 for this application. No one validates against the spec of the ComicInfo.

API.Tests/Services/ArchiveServiceTests.cs
```diff
@@ -291,7 +291,25 @@ public void CanParseComicInfo()
                 Web = "https://www.comixology.com/BTOOOM/digital-comic/450184"
             };
 
+            // XXX: This is not doing anything useful
             Assert.NotStrictEqual(expected, actual);
+            // XXX: This also works
+            Assert.NotStrictEqual(null, actual);
```
majora2007, line 297: Explain why this isn't doing anything useful, when we are validating what was parsed matches what we expect?

API.Tests/Services/ArchiveServiceTests.cs
```diff
@@ -291,7 +291,25 @@ public void CanParseComicInfo()
...
+        }
+
+        [Fact]
+        public void CanParseComicInfo2()
```
majora2007, line 301: Please make this more expressive, to help understand what this is adding versus the existing test. Like CanParseComicInfo_DefaultNumberIsBlank

majora2007 (changes requested):

API.Tests/Services/ArchiveServiceTests.cs
tjarls: First of all `NotStrictEqual` makes no sense here. If the prerequisite conditions were met, the test should have been:
```
Assert.StrictEqual(expected, actual);
```
It would have asserted that the objects types matched and the values matched instead of the opposite as it current stands.

The later check (value equality) relies on the class (here ComicInfo) implementing the Equals method, for checking equality of the "value". Because an explicit one is not defined for ComicInfo, the test here falls back to the one implemented on Object (https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-6.0#system-object-equals(system-object)). The Object implementation is only true when the identifiers point to the same instance of an object and here `actual` and `expected` are 2 different instances of ComicInfo that happen to have the same values.

There are two ways to fix this test:
 1) implement an Equal() method on object ComicInfo and use `StrictEqual`.
 2) Check individually the field on the two objects like I did on the next test.
 
Please let me know how you'd rather see this addressed?

API/Services/BookService.cs
tjarls: The following code is dead code because string.IsNullOrEmpty(seriesIndex) is already known to be false as it's a condition for this block.

The rest of the code ensures that when a seriesSort sort is filled with the value of series. Note that this value can be an empty string. This final value is actually set in `ReadingItemService.cs` (which make any attempt to have a similar logic here redundant also). The code here affects the final outcome (in `ReadingItemService.cs`) so:
- When `Series` is set and non-empty, `SeriesSort` is set to the same value and that will be used to sort the series.
- When `Series` is empty, TitleSort is used instead.

Please check issue #1354 for details and examples of the issue this is trying to solve.

Long term, instead of copying Series into SeriesSort, we should instead transform or/and normalize it. There's two mechanisms I can think of that would be beneficial:
- Unicode normalization to remove accents, etc
- Language specific transform as part of localization (I noticed that's somewhere in the future plans). Here we'd drop leading articles like "the" in English. Ideally this is language dependend and only triggered when a language is set in the metadata.

API.Tests/Services/ArchiveServiceTests.cs
tjarls: I have added a partial implementation of `Equals()` to illustrate the issue with incorrect usage of NotStrictEqual.

On second though, implementing a correct `Equals()` method seems to be far too expensive to be justified by only a couple of unit tests. In particular, multiple values fields like Characters, etc would need something like being split and sorted before being compared.

API/Data/Metadata/ComicInfo.cs
```diff
@@ -125,4 +125,18 @@ public static void CleanComicInfo(ComicInfo info)
     }
 
 
+    // TODO: Example of Equals() implementation. This currently only
```
majora2007, line 128: Please finish out the implementation for Equals, implementing a half implementation won't get this PR merged as it opens potential bugs.

API.Tests/Services/ArchiveServiceTests.cs
majora2007: But we don't use any equality checks for ComicInfo anywhere in the code, so I don't think it would be that bad. Instead, we can rewrite the Unit test to be more like how you wrote yours, with individual Assert statements.

majora2007 (approved):
Tested and looks fine to me.

Issue #1104, opened 2022-02-22
Name Parser strips out 'special' from words that contain this as part of their own word.


ocgineer:
**Describe the bug**

The name parser (manga library) strips out the literal part 'special' of words that have this part of their own word.

For example;
```
Dr. Ramune - Mysterious Disease Specialist v01 (2020) (Digital) (danke-Empire).cbz
Dr. Ramune - Mysterious Disease Specialist v02 (2020) (Digital) (danke-Empire).cbz
```

Is parsed by Kavita with the series title of 
`Dr. Ramune - Mysterious Disease ist` 

**Expected behavior**

Only the exact match 'Special' with bounding whitespace should be parsed out if you want to consider a title 'special' without the use of comicinfo.xml. Any other word containing 'special' as part of its word, should not be parsed out. 

**Screenshots**

![image](https://user-images.githubusercontent.com/12483966/155059111-a49bf1c1-ce52-4fb0-99cf-282add993d00.png)

**Desktop (please complete the following information):**
 - OS: Unraid Docker 
 - Browser Chrome
 - Version v0.5.1.0

majora2007:
This is something I can't solve unfortunately, the only solution is to embed the ComicInfo.xml, which will not do special stripping or rename the series in the UI, which will then stick and still match correctly.

Issue #1513, opened 2022-09-05
Scan can't handle 2 series with {omnibus edition} edition difference


ocgineer:
**Describe the bug**

Seems Kavita can't handle two of the same series anymore, with a {omnibus edition} difference that was added to the filename to denote an edition to allow both editions to co-exist in Kavita.

```
|- Magic Knight Rayearth {Omnibus Edition}
|    |- Magic Knight Rayearth {Omnibus Edition} v01 (2011) (Digital) (BlurPixel-Empire).cbz
|    |- Magic Knight Rayearth {Omnibus Edition} v02 (2012) (Digital) (BlurPixel-Empire).cbz
| 
|- Magic Knight Rayearth
     |- Magic Knight Rayearth v01 (2019) (Digital) (danke-Empire).cbz
     |- Magic Knight Rayearth v02 (2019) (Digital) (danke-Empire).cbz
     |- Magic Knight Rayearth v03 (2019) (Digital) (danke-Empire).cbz
```

After Scan `Magic Knight Rayearth {Omnibus Edition}` shows up as `Magic Knight Rayearth` instead of _as expected_ `Magic Knight Rayearth { Edition}` similar how {Special Edition} works. Also, the 2 volumes have the wrong covers, from the non-omnibus edition!

Renamed this to `Magic Knight Rayearth (Omnibus Edition)` then did a scan to have the non-omnibus show up as `Magic Knight Rayearth` properly. I didn't do anything with the wrong covers yet on the omnibus edition.

After the scan, I tried refreshing the covers of the omnibus edition but Kavita promptly deleted the whole series!

Scanned the library again, it removed volume 3 under `Magic Knight Rayearth` and replaced v1 and v2 by with the 2 omnibus versions with wrong cover again (I check details to see it is the omnibus version by filename). Refresh Covers worked now as it still had the wrong covers.

So to work around this issue we have to add the {omnibus edition} first, scan, rename, add the other edition, and scan.

**Expected behavior**
Both series can be added correctly at the same time before a scan, and the {omnibus edition} series name will still include the {edition} (or part of it if it contains a keyword such as omnibus like in this case), similar to using {Special Edition}.

**Desktop (please complete the following information):**
 - OS: Unraid Docker
 - Browser Chrome
 - Version 0.5.6.0

ocgineer:
[...]

PR #1531, opened 2022-09-14
Parser optimization part1

### LINKED ISSUE #1529: False "special" positives in ComicSpecialRegex
The current special regexp does not use word boundaries on the keywords used to recognise special issues/chapters. As a result it can falsely identify those words when part of a larger word. Some examples of false positives:
```
Mazebook 001
G.I. Joe - A Real American Hero Yearbook 004 Reprint (2021)
```
Using word boundaries (\b) in the regex fixes this issue.

### LINKED ISSUE #1530: Please add some common variants of some of the exiting "special" keywords
Currently the parser recognises `OneShot` and `One-Shot` as a keyword for a special issue. The variant `One Shot` is also commonly found and should be added.

Similarly `Hors Série` is a recognised keyword. Its canonical form is actually `Hors-Série`. Variants where the e is missing the accute accent (eg `Hors Serie`) are also commonly found.

Adding all those variants would increase the use cases where this mechanism is useful.

tjarls:
# Changed
- Changed: Optimized some of the parser methods. CleanTitle is up to 10 times faster.
- Changed: Support additional variants for special keywords "One Shot", "Hors-Série" (Closes #1530 )

# Fixed
- Fixed: Special regexps are now word bound which reduces false positives (Fixes #1529 #1104)
- Fixed: Cases where CleanTitle would erase the entire file name when it starts and ends with parenthesis


API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -938,34 +844,10 @@ public static string CleanTitle(string title, bool isComic = false)
...
-            title = title.Substring(1);
-        }
+        title = title.Trim(SpacesAndSeparators);
 
         title = EmptySpaceRegex.Replace(title, " ");
 
-        return title.Trim();
-    }
-
-    private static string RemoveReleaseGroup(string title)
```
tjarls, line 958: RemoveReleaseGroup has been dropped entirely since RemoveEditionTagHolders already clears those tags.

API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -840,73 +793,26 @@ public static string ParseComicChapter(string filename)
...
-                    title = title.Replace(match.Value, string.Empty).Trim();
-                }
-            }
-        }
+        title = MangaEditionRegex.Replace(title, string.Empty);
 
         return title;
     }
 
     private static string RemoveMangaSpecialTags(string title)
```
tjarls, line 803: Those are now one-liners. I left them as their own methods just to make the PR easier to read but we could just add their code into CleanTitle instead and avoid the risk of function call overhead albeit the compiler should optimize those away.

API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -70,6 +70,12 @@ public static class Parser
     private const string Number = @"\d+(\.\d)?";
     private const string NumberRange = Number + @"(-" + Number + @")?";
 
+    // Some generic reusage regex patterns:
+    // - non greedy matching of a string where parenthesis are balanced
+    private const string BalancedParen = @"(?:[^()]|(?<open>\()|(?<-open>\)))*?(?(open)(?!))";
+    // - non greedy matching of a string where parenthesis are balanced
```
tjarls, line 76: I need to fix that comment. square brackets instead of parentheis

API/Services/Tasks/Scanner/Parser/DefaultParser.cs
```diff
@@ -85,10 +85,10 @@ public ParserInfo Parse(string filePath, string rootPath, LibraryType type = Lib
             ret.Edition = edition;
         }
 
-        var isSpecial = type == LibraryType.Comic ? Services.Tasks.Scanner.Parser.Parser.ParseComicSpecial(fileName) : Services.Tasks.Scanner.Parser.Parser.ParseMangaSpecial(fileName);
+        bool isSpecial = type == LibraryType.Comic ? Services.Tasks.Scanner.Parser.Parser.IsComicSpecial(fileName) : Services.Tasks.Scanner.Parser.Parser.IsMangaSpecial(fileName);
```
majora2007, line 88: Please use var

API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -938,34 +844,10 @@ public static string CleanTitle(string title, bool isComic = false)
...
-        if (title.EndsWith("-") || title.EndsWith(","))
-        {
-            title = title.Substring(0, title.Length - 1);
-        }
-
-        if (title.StartsWith("-") || title.StartsWith(","))
-        {
-            title = title.Substring(1);
-        }
+        title = title.Trim(SpacesAndSeparators);
```
majora2007, line 847: I like this

API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -1150,4 +1032,9 @@ public static bool HasComicInfoSpecial(string comicInfoFormat)
     {
         return FormatTagSpecialKeywords.Contains(comicInfoFormat);
     }
+
+    private static string ReplaceUnderscores(string name) {
+        if (string.IsNullOrEmpty(name)) return name;
+        return name.Replace("_", " ");
```
majora2007, line 1038: Couldn't we do name?.Replace() to avoid the if statement?

majora2007 (changes requested):
I need to give this some dedicated testing and exploration, I will look into this after my current branch is done. 

So far, the changes don't look too bad. Please keep in mind if you're moving forward on other Regex that I plan to replace most Regex with source generators in .net 7, which brings even deeper performance improvements.

majora2007:
Please add unit tests for all the new individual regex cases you've added, to make maintenance easier. For example, balanced paren, if we can have unit tests for that, it will not only help with maintenance, but also understanding of the code.

tjarls:
> Please also commit up the Benchmark you're using. I'm sure you're using the Benchmark project I already have setup. We can leave them in and build more on top of them.

For my tests I kept a copy of old `Parser.cs` where I renamed the namespace to in order to perform the comparison. Not exactly what we want long term. I'll post it in a separate branch so that you can take a look before I add the commit here.

Ideally, the bench results would be collected as artifacts during builds within the CI/CD and rely on an external tool to do the comparison between versions. I have no idea what framework would be available to do so within github, unfortunately.

majora2007:
Before we merge, to avoid conflicts, can you add a case for #1513? The issue is that CleanTitle strips {} out and since we are now using that as a custom tag in Kavita (and the likelihood of finding it from scene releases is low), we shouldn't strip from Series anymore.

We just need to add 
```
[InlineData("Magic Knight Rayearth {Omnibus Edition} v01 (2011) (Digital) (BlurPixel-Empire)", "Magic Knight Rayearth {Omnibus Edition}")]
```
to our Manga and Comic Parser Tests for Series Name.

tjarls:
I've added the benchmarking for CleanTitle. The data file is on the large side and may be a concern. Please advise if it should be much smaller.

majora2007:
@tjarls I'm just waiting for you to reduce the size of the benchmark data to something like 100 lines and then I think we are good to merge.

majora2007:
I'm doing the final round of validation on this then we should be able to merge it today.

PR #1551, opened 2022-09-21
Fallback to other locations when ComicInfo.xml not at root of archive

tjarls:
# Changed
- Changed: Try using ComicInfo in other location when not present at the root of the comic archive, as the spec describes.


API/Services/ArchiveService.cs
```diff
@@ -328,12 +328,11 @@ public bool IsValidArchive(string archivePath)
         return false;
     }
 
-    private static bool ValidComicInfoArchiveEntry(string fullName, string name)
+    private static bool BackupComicInfoArchiveEntry(string fullName, string name)
```
majora2007, line 331: The method name is not very clear, it should be IsComicInfoArchiveEntry() which indicated that it's checking an entry to be valid and in the caller, we see the context of the backup logic.

API/Services/ArchiveService.cs
```diff
@@ -356,7 +355,8 @@ private static bool ValidComicInfoArchiveEntry(string fullName, string name)
                 {
                     using var archive = ZipFile.OpenRead(archivePath);
 
-                    var entry = archive.Entries.FirstOrDefault(x => ValidComicInfoArchiveEntry(x.FullName, x.Name));
+                    var entry = archive.Entries.FirstOrDefault(x => x.FullName == ComicInfoFilename) ??
```
majora2007, line 358: From what I've seen, FullName might not always be populated, hence why I had:
(string.IsNullOrEmpty(fullName) && name.Equals(ComicInfoFilename)).

majora2007 (changes requested):
What was the difference in terms of memory and performance btw? Just curious what you saw.

API/Services/ArchiveService.cs
tjarls: Do you have an example I can use to reproduce that? I'd like to understand and get something like it for the unit tests. I had assumed that this was meant for rar archives where Path.GetDirectoryName would return null (which now that I have read on that method is indeed unlikely).

tjarls:
> What was the difference in terms of memory and performance btw? Just curious what you saw.

Sorry it was late, and did not report on that. The difference is within the statistical noise, so too small to be conclusive. So typically within 1 us, while the whole operation depends on the size of the archive from around 50 us for the zips from the unit tests to 400 with examples from my libraries. My server, where I ran the tests is unfortunately too busy to be precise enough and in some of my tests the order was sometimes even inverted.
One example:
|            Method |     Mean |    Error |   StdDev |
|------------------ |---------:|---------:|---------:|
| TestGetComicInfo2 | 53.72 us | 0.517 us | 0.596 us |
|  TestGetComicInfo | 53.14 us | 0.470 us | 0.523 us |

tjarls:
I have updated the benchmarks to compare cases where
1) comicInfo is in the correct location. This is slightly faster than with current version.
2) comicInfo is in duplicated. This would have been the same as 1) but it's slower because the archive is larger
3) comicInfo is not in root but present in subdir. This should be slightly slower.
The size of the archive is the main factor in timings and memory usage. The lookup of the right ComicInfo file, regardless of the scenario, is just too small to have any impact.

tjarls:
Here's the results. 1) is baseline, 2) is "duplicate" and 3) is "outside_root":
|                        Method |     Mean |    Error |   StdDev | Ratio | Rank |   Gen0 |   Gen1 |   Gen2 | Allocated | Alloc Ratio |
|------------------------------ |---------:|---------:|---------:|------:|-----:|-------:|-------:|-------:|----------:|------------:|
|     TestGetComicInfo_baseline | 49.72 us | 0.437 us | 0.503 us |  1.00 |    1 | 0.4272 | 0.3662 |      - |   37.2 KB |        1.00 |
|    TestGetComicInfo_duplicate | 52.02 us | 0.308 us | 0.355 us |  1.05 |    2 | 0.4272 | 0.3662 | 0.0610 |  38.77 KB |        1.04 |
| TestGetComicInfo_outside_root | 52.11 us | 0.141 us | 0.162 us |  1.05 |    2 | 0.4272 | 0.3662 | 0.0610 |  38.17 KB |        1.03 |

API/Services/ArchiveService.cs
tjarls: The real concern here is that when FullName is absent, we cannot really tell if ComicInfo is in the root directory of the archive. This needs to be understood.

majora2007:
I have some examples as it also caught me off guard. Let me find them and I'll shoot over via discord.

majora2007 (approved):
Looks good, thanks.

PR #1583, opened 2022-10-11
New Series Relation - Edition

### LINKED ISSUE #1533: New Relation option for 'different' editions of the same series.
Created issue as per request via Discord.

For people that keep the different published editions and like to have them both in Kavita, to link them together with relationships my proposal would be to add an additional relationship like `Alternate Edition`, `Other Edition`,  or simply `Edition`. 

Preferably the relationship should **not** show 'Parent' as there is no real parent here. It would be best for users to set-up the other edition similar to prequel/sequel (if automatic does not allow this) so that whenever the user checks either of the edition the user can see other edition always in the relations tab.

With this extra relation option, users can use this for, but not limited to;
- Normal published <> Omnibus published
- Special Edition Anniversary Edition
- Expired License <> New License (re)published
- Black & White <> Color Edition
- Official <> Scanlation

The current available relation option do not convey that it is a different edition. `Alternate Version` which is the closest would be more used to link Manga with the Light Novel, as the format is different but basically the same story. With different Editions only slight changes are made, like translation or colorization while the story is exactly the same.

ThePromidius:
# Added
- Added: Added a new Series relation called Edition. (A generic relation with no Parent relation) (Closes #1533 )


API.Tests/Services/SeriesServiceTests.cs
```diff
@@ -1119,4 +1137,176 @@ public void GetFirstChapterForMetadata_NonBook_ShouldReturnVolume1_WhenFirstChap
...
+        var relationDto = InstantiateRelationsDto(series1);
+        relationDto.Adaptations.Add(2);
+
+        await _seriesService.UpdateRelatedSeries(relationDto);
+        // Expected is only one instance of the relation (hence not duping)
+        Assert.Equal(2, series1.Relations.Single(s => s.TargetSeriesId == 2).TargetSeriesId);
+    }
+
+    [Fact]
+    public async Task AddRelation_EditionPrequelSequel_ShouldNotHaveParent()
```
majora2007, line 1249: This should be GetRelatedSeries_EditionPrequelSequel_ShouldNotHaveParent()

API.Tests/Services/SeriesServiceTests.cs
```diff
@@ -1119,4 +1137,176 @@ public void GetFirstChapterForMetadata_NonBook_ShouldReturnVolume1_WhenFirstChap
     }
 
     #endregion
+
+    #region SeriesRelation
+    [Fact]
+    public async Task UpdateRelatedSeries_ShouldAddAllRelations_And_DeleteAllRelations()
```
majora2007, line 1143: I would copy/paste the code and do the Delete in a separate test to keep it clean.

majora2007 (changes requested):
Looks pretty good. just need a few small things

majora2007:
I'll pull this tomorrow

API.Tests/Services/SeriesServiceTests.cs
majora2007: @ThePromidius still waiting on the comments to be taken care of

majora2007:
Once you take care of my comments, I'm good to merge this.

API.Tests/Services/SeriesServiceTests.cs
ThePromidius: I completely missed this. Will do

majora2007 (approved):

PR #1649, opened 2022-11-13
Run Kavita as non-root user in Docker container

mikeymakesit:
# Added
- Added: Ability to provide environment variables PUID, PGID, and KAVITAUSER which will set the uid, gid and optionally the username under which Kavita will be started in a Docker container.  The defaults are set to 1000, 1000, "kavita".  The user and group will be created in the container if they don't already exist.  Execution of "./Kavita" will be done as that user.


mikeymakesit:
Will submit a separate PR with defaults to run as root.

Issue #1662, opened 2022-11-19
Update CI/CD


majora2007:
- Update based on deprecation warnings
- Update [setup-dotnet@3](https://github.com/actions/setup-dotnet/releases)
- Update setup-checkout@3
- Do develop and stable actually need to rely on build? Since cross jobs, there are no files shared? And in order to hit those flows, a build/test would already run. 
- Add a new check if PR body has any quotes we don't support

PR #1675, opened 2022-12-06
Add github action CI with first pass of build script

adamhathcock:
Starts to address [1662](https://github.com/Kareadita/Kavita/issues/1662)

- Adds Github Action CI for Angular and C#
- Introduces C# based build script using Bullseye https://github.com/adamralph/bullseye
- Start of Directory.props for common csproj params
- Executes Angular Tests but not C# tests (couldn't get them to work locally for the moment)

I just came across this project and wanted to help a bit.  

Suggested next steps:
- Make tests work with script
- ~~Intro warnings as errors and fix all~~
- Introduce csharpier to have style https://github.com/belav/csharpier
- There was talk of running the Dockerfile build, which I can do.  I use the C# script inside a build container then copy to the as in https://github.com/gothinkster/aspnetcore-realworld-example-app/blob/master/Dockerfile (needs updating I know)

Let me know if what, if any, direction to head down.  Thanks.


.github/workflows/ci-dotnet.yml
```diff
@@ -0,0 +1,21 @@
+name: Kavita - CSharp
```
majora2007, line 1: Why do we need this when we are already building in the main workflow? This would increase our build to 3 times in a single PR, which is the opposite of the ask in the issue (builds are too slow)

build/build.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+      <OutputType>Exe</OutputType>
+      <AnalysisMode>Recommended</AnalysisMode>
+      <WarningsAsErrors>true</WarningsAsErrors>
+      <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+      <CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+      <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+      <Nullable>enable</Nullable>
```
majora2007, line 10: I already have this enabled on a separate branch and dealing with the warnings. Let's keep this and something like WarningAsErrors off. Keep the changes simple and progressively move towards better code.

feature/net7.0

majora2007:
Hi Adam, thank you for the PR and for SharpCompress. I'm not understanding the need of adding Bullseye or how it adds enhancement from the existing build system we have in build.sh? I don't see the ability to trigger one build flavor or all with the new script nor how it will fit into the overall build strategy? 

As for angular tests, I don't think I have any to be honest. It would be nice, but there is another effort going to use Cypress and build out a testing matrix. 

Please let me know. PS, I like the Build.Props, I didn't know about that (as I learned .NET 2 years ago with the start of this project).

.github/workflows/ci-dotnet.yml
adamhathcock: The idea was to split out different things into different jobs for faster feedback.  Something like this would be "faster" while there would be a main build once things are merged that would do a docker build and other stuff if needed.

Maybe I misread the requirement for the issue but I thought this was what was being asked for with the usage of certain github actions

build/build.csproj
adamhathcock: These extra params in this csproj keep this project locked down with more settings until they can be graduated to the Directory.Build.props for all projects.  

The idea would be to put these in each csproj progressively and when they all match move to Directory.Build.props.  How many steps depends on how big a single PR for fixing warnings, etc should be :)

majora2007:
I can't fully visualize Bullseye in action, but I think there is potential in this PR to streamline our build system. 

First let me explain some necessary items that must exist:
1. We must do sonarcloud scans on all PRs. Currently, Github secrets don't extend to forks, so this would be great to be a separate yml, so everything else can run on fork PRs except that.
2. I need the ability to build individual and all variants of platforms RIDs, as you'll see in build.sh
3. I need the ability to build a package outlined in build.sh. I'm not sure if Bullseye can handle that.
4. In the original issue, the goal is to reduce speed (PRs to build can take 20 mins) and provide additional checks, like checking for ` or " in the PR body, as one of our actions fails when in there (I believe due to sending to discord). 

Let me know how I can help support you on this effort. If you're in discord, you can reach out to me on the Kavita discord https://discord.com/invite/b52wT37kt7

majora2007:
No problem and no rush, my next release window is next year with the holiday approaching soon, so we have lots of time to do this.

majora2007:
While doing this, feel free to come up with a creative solution for the swagger openapi generation. I need to generate this for the openapi spec that we host, but the GA would need to commit it from the run. This was something I was going to circle around back to, as right now I'm committing manually and it's extremely annoying.

build/build.csproj
majora2007: Just for reference, the PR is up (but not building on GA) for nullability and .net 7. As you can see, it's quite large.
https://github.com/Kareadita/Kavita/pull/1677

build/Program.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        Run("cp", $"{outputFolder}/API.exe {outputFolder}/Kavita.exe");
+    }
+    else
+    {
+        Run("mv", $"{outputFolder}/API {outputFolder}/Kavita");
+    }
+
+    Console.WriteLine("Copying appsettings.json");
+    Run("cp", $" API/config/appsettings.json {outputFolder}/config/appsettings.json");
```
majora2007, line 66: This step might not be needed anymore given your, what I believe is a fix, in the csproj to copy the appsettings.json over on build.

adamhathcock:
> While doing this, feel free to come up with a creative solution for the swagger openapi generation. I need to generate this for the openapi spec that we host, but the GA would need to commit it from the run. This was something I was going to circle around back to, as right now I'm committing manually and it's extremely annoying.

Do you need it as a checked-in file?  I usually have been using swashbuckle to host the definition that's generated as well as a SwaggerUI.  Not sure what the openapi file is used for so maybe my suggestion is useless.

build/Program.cs
adamhathcock: Removed

adamhathcock:
Hopefully, I've got openapi checking in.  All the actions need testing.

The `build_target.sh` script looks outdated.  The update version function in the old `build.sh` looks unused as well.

Need to go through `docker-build.sh` but I think that's unused too?

majora2007:
@Kizaing can you remind me if anything except build.sh is needed in our build scripts?

Kizaing:
Yeah only monorepo-build.sh and build.sh are used, the latter is for local builds. The other two aren't needed any more as far as I know

PR #1713, opened 2022-12-30
Allow changing listening ip addresses

Kupferhirn:
# Added
- Added: Allow changing listening IP addresses from server settings to one or multiple addresses. Defaults to 0.0.0.0,::. This is not available for Docker users.


Kavita.Common/Configuration.cs
```diff
@@ -140,4 +146,54 @@ private static int GetPort(string filePath)
...
+    {
+        if (new OsInfo(Array.Empty<IOsVersionAdapter>()).IsDocker)
+        {
+            return;
+        }
+
+        try
+        {
+            var currentIpAddresses = GetIpAddresses(filePath);
+            var json = File.ReadAllText(filePath).Replace("\"IpAddresses\": \"" + currentIpAddresses, "\"IpAddresses\": \"" + ipAddresses);
```
Kupferhirn, line 162: Changing the IP address with the UI will not work on existing installations because the string replace does nothing if the appsettings.json does not contain the setting. 
Same is true for the port if it was manually changed in the appsettings.json to 'Port:5000' (no space after : ).
I would change the config set methods to use json de-/serialization instead if this is okay for you?

Kupferhirn:
Screenshot of the new UI.
grafik

majora2007:
I'm away on holiday but I'll give this a look over afterwards.

Kavita.Common/Configuration.cs
majora2007: If you know how, I would love it. My config code is very hacky.

Kupferhirn:
@therobbiedavis 
The listening address for docker installs can be changed with the "ports" setting of docker run/docker-compose.
Changing this in the UI will not work for docker install.

majora2007:
I'm ready to test this out and get it merged, please resolve the one conflict and ping me.

Kupferhirn:
Hi @majora2007 
I resolved the conflict and also updated the the set methods to use json serialization.

majora2007:
Took a look, here is my thoughts:

1) Ip should be IP. 
2) Tooltip should just say "This does not apply to Docker"
3) you need to remove BOM on the files you touched. Your IDE has applied BOM (byte order mark) which is causing issues in my IDE (Rider)
4) What are some valid inputs? Can we provide inline pattern validation before we let the user save? 
5) We need a reset button on the UI to reset to the default.

Kupferhirn:
@majora2007 
> 4. What are some valid inputs? Can we provide inline pattern validation before we let the user save?

I think valid are all IPv4 and IPv6 addresses with and without subnet.
I can definitely add validation.

> 5. We need a reset button on the UI to reset to the default.

You mean only for this input field?

majora2007:
Correct, only on that field, like how we have for email URL.

API/Controllers/SettingsController.cs
```diff
@@ -147,6 +148,14 @@ public async Task<ActionResult<ServerSettingDto>> UpdateSettings(ServerSettingDt
 
             if (setting.Key == ServerSettingKey.IpAddresses && updateSettingsDto.IpAddresses != setting.Value)
             {
+                // Validate IP addresses
+                foreach (var ipAddress in updateSettingsDto.IpAddresses.Split(','))
+                {
+                    if (!IPAddress.TryParse(ipAddress.Trim(), out _)) {
```
majora2007, line 154: This is good, was actually thinking of just doing this in the UI layer.

UI/Web/src/app/admin/manage-settings/manage-settings.component.ts
```diff
@@ -41,7 +41,7 @@ export class ManageSettingsComponent implements OnInit {
       this.settingsForm.addControl('bookmarksDirectory', new FormControl(this.serverSettings.bookmarksDirectory, [Validators.required]));
       this.settingsForm.addControl('taskScan', new FormControl(this.serverSettings.taskScan, [Validators.required]));
       this.settingsForm.addControl('taskBackup', new FormControl(this.serverSettings.taskBackup, [Validators.required]));
-      this.settingsForm.addControl('ipAddresses', new FormControl(this.serverSettings.ipAddresses, [Validators.required]));
+      this.settingsForm.addControl('ipAddresses', new FormControl(this.serverSettings.ipAddresses, [Validators.required, Validators.pattern(/^(\s*((([12]?\d{1,2}\.){3}[12]?\d{1,2})|(([\da-f]{0,4}\:){0,7}([\da-f]{0,4})))\s*\,)*\s*((([12]?\d{1,2}\.){3}[12]?\d{1,2})|(([\da-f]{0,4}\:){0,7}([\da-f]{0,4})))\s*$/i)]));
```
majora2007, line 44: Refactor this into a const called ValidIpAddress.

Kupferhirn:
I finished all requested changes.

I added server side validation to keep client validation simple.
Checking every invalid address would make the regex very long and unreadable.
This regex checks the basics but still allows some invalid addresses like 10.10.10.299


>     3. you need to remove BOM on the files you touched. Your IDE has applied BOM (byte order mark) which is causing issues in my IDE (Rider)

It was actually the other way around. My IDE removed the BOM. Should be fine now.

majora2007:
Validation looks good, let me give this another run tomorrow and if all looks good, I'll merge in for the v0.7 release.

majora2007 (approved):
Thank for for the PR. Everything looked good for me.

PR #1776, opened 2023-02-06
Feature/expose progress last modified date

### LINKED ISSUE #1700: Expose the last modified date of the reading progress in the API
Please expose the last modified date of the reading progress in the API for each chapter.

Reason:
It would be useful to have this information in case you use several devices in offline mode, it would allow to keep the most recent reading status when you sync back.

Api concerned:
/api/Reader/get-progress
/api/Reader/progress
/api/Series/volume
/api/Series/volumes
/api/Series/chapter
/api/Series/chapter-metadata
/api/Series/series-detail

hgourvest:
# Added
- UTC creation and update dates
- Expose the date of last modification of the reading status in UTC (Fixes #1700)

# Changed
- ProgressDto.LastModified renamed to LastModifiedUtc

# Fixed
- Fixed: SeriesController.GetChapter() now also update PagesRead

### DIFF
--- a/API/Controllers/ReaderController.cs
+++ b/API/Controllers/ReaderController.cs
@@ -467,6 +467,7 @@ public class ReaderController : BaseApiController
             progressBookmark.SeriesId = progress.SeriesId;
             progressBookmark.VolumeId = progress.VolumeId;
             progressBookmark.PageNum = progress.PagesRead;
+            progressBookmark.LastModifiedUtc = progress.LastModifiedUtc;
             progressBookmark.BookScrollId = progress.BookScrollId;
         }
         return Ok(progressBookmark);
--- a/API/DTOs/ChapterDto.cs
+++ b/API/DTOs/ChapterDto.cs
@@ -43,6 +43,10 @@ public class ChapterDto : IHasReadTimeEstimate
     /// </summary>
     public int PagesRead { get; set; }
     /// <summary>
+    /// Last time the progress was synced from UI or external app
+    /// </summary>
+    public DateTime ProgressLastModifiedUtc { get; set; }
+    /// <summary>
     /// If the Cover Image is locked for this entity
     /// </summary>
     public bool CoverImageLocked { get; set; }
--- a/API/DTOs/ProgressDto.cs
+++ b/API/DTOs/ProgressDto.cs
@@ -21,7 +21,7 @@ public class ProgressDto
     /// </summary>
     public string BookScrollId { get; set; }
     /// <summary>
-    /// Last time the progress was synced from UI or external app
+    /// Last time in Utc the progress was synced from UI or external app
     /// </summary>
-    public DateTime LastModified { get; set; }
+    public DateTime LastModifiedUtc { get; set; }
 }
--- a/API/Data/Migrations/20230205222635_EntityTimestampsUtc.cs
+++ b/API/Data/Migrations/20230205222635_EntityTimestampsUtc.cs
@@ -1,0 +1,236 @@
+﻿using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace API.Data.Migrations
+{
+    public partial class EntityTimestampsUtc : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<DateTime>(
+                name: "CreatedUtc",
+                table: "Volume",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "LastModifiedUtc",
+                table: "Volume",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "CreatedUtc",
+                table: "SiteTheme",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "LastModifiedUtc",
+                table: "SiteTheme",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "CreatedUtc",
+                table: "Series",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "LastModifiedUtc",
+                table: "Series",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "CreatedUtc",
+                table: "ReadingList",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "LastModifiedUtc",
+                table: "ReadingList",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "CreatedUtc",
+                table: "MangaFile",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "LastModifiedUtc",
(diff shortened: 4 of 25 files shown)


API/DTOs/ProgressDto.cs
```diff
@@ -21,7 +21,7 @@ public class ProgressDto
     /// </summary>
     public string BookScrollId { get; set; }
     /// <summary>
-    /// Last time the progress was synced from UI or external app
+    /// Last time in Utc the progress was synced from UI or external app
     /// </summary>
-    public DateTime LastModified { get; set; }
+    public DateTime LastModifiedUtc { get; set; }
```
majora2007, line 26: Do not cause breaking changes. Add the additional field, do not break my implementations

API/DTOs/ChapterDto.cs
```diff
@@ -43,6 +43,10 @@ public class ChapterDto : IHasReadTimeEstimate
     /// </summary>
     public int PagesRead { get; set; }
     /// <summary>
+    /// Last time the progress was synced from UI or external app
```
majora2007, line 46: This comment doesn't make sense for within Kavita. It would be more correct to say, Last time Chapter was read in Utc.

API/DTOs/ProgressDto.cs
```diff
@@ -21,7 +21,7 @@ public class ProgressDto
     /// </summary>
     public string BookScrollId { get; set; }
     /// <summary>
-    /// Last time the progress was synced from UI or external app
+    /// Last time in Utc the progress was synced from UI or external app
```
majora2007, line 24: This comment doesn't make sense for within Kavita. It would be more correct to say, Last time Chapter was read in Utc.

API/Controllers/ReaderController.cs
```diff
@@ -467,6 +467,7 @@ public async Task<ActionResult<ProgressDto>> GetProgress(int chapterId)
             progressBookmark.SeriesId = progress.SeriesId;
             progressBookmark.VolumeId = progress.VolumeId;
             progressBookmark.PageNum = progress.PagesRead;
+            progressBookmark.LastModifiedUtc = progress.LastModifiedUtc;
```
majora2007, line 470: Remove this. IEntityDate does this for you already.

API/Data/Migrations/20230205222635_EntityTimestampsUtc.cs
```diff
@@ -0,0 +1,236 @@
...
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
+
+            migrationBuilder.AddColumn<DateTime>(
+                name: "LastModifiedUtc",
+                table: "Volume",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
```
majora2007, line 24: You will need to setup a converter for the Utc Fields so that when they are read in, the DateTimeKind is set.

See https://stackoverflow.com/questions/4648540/entity-framework-datetime-and-utc/73154546#73154546

API/Data/Repositories/ChapterRepository.cs
```diff
@@ -247,4 +248,13 @@ public async Task<IList<MangaFile>> GetFilesForChaptersAsync(IReadOnlyList<int>
...
             .ToListAsync();
     }
+
+    public async Task AddChapterProgress(int userId, ChapterDto chapter)
+    {
+        var progress = await _context.AppUserProgresses.FirstOrDefaultAsync(x =>
+            x.AppUserId == userId && x.ChapterId == chapter.Id);
+        if (progress == null) return;
+        chapter.PagesRead = progress.PagesRead;
+        chapter.ProgressLastModifiedUtc = progress.LastModifiedUtc;
```
majora2007, line 258: If you are adding progress, add the non Utc field as well to keep consistent.

API/Data/Repositories/ReadingListRepository.cs
```diff
@@ -218,6 +218,7 @@ public async Task<IEnumerable<ReadingListItemDto>> GetReadingListItemDtosByIdAsy
             if (progressItem == null) continue;
 
             progressItem.PagesRead = progress.PagesRead;
+            progressItem.ProgressLastModifiedUtc = progress.LastModifiedUtc;
```
majora2007, line 221: Add the non Utc field as well to keep consistent.

API/Data/Repositories/ReadingListRepository.cs
```diff
@@ -241,8 +242,11 @@ public async Task<IEnumerable<ReadingListItemDto>> AddReadingProgressModifiers(i
 
         foreach (var item in items)
         {
-            var progress = userProgress.Where(p => p.ChapterId == item.ChapterId);
-            item.PagesRead = progress.Sum(p => p.PagesRead);
+            var progress = userProgress.SingleOrDefault(p => p.ChapterId == item.ChapterId);
+            if (progress == null) continue;
+
+            item.PagesRead = progress.PagesRead;
```
majora2007, line 248: Why are you removing the Sum?

API/Data/Repositories/ReadingListRepository.cs
```diff
@@ -241,8 +242,11 @@ public async Task<IEnumerable<ReadingListItemDto>> AddReadingProgressModifiers(i
 
         foreach (var item in items)
         {
-            var progress = userProgress.Where(p => p.ChapterId == item.ChapterId);
-            item.PagesRead = progress.Sum(p => p.PagesRead);
+            var progress = userProgress.SingleOrDefault(p => p.ChapterId == item.ChapterId);
+            if (progress == null) continue;
```
majora2007, line 246: You can skip the null check and use null coalescing instead. 

item.ProgressLastModifiedutc = progress?.LastModifiedUtc;

Add the non Utc field to be consistent.

API/Data/Repositories/VolumeRepository.cs
```diff
@@ -218,7 +219,11 @@ private async Task AddVolumeModifiers(int userId, IReadOnlyCollection<VolumeDto>
         {
             foreach (var c in v.Chapters)
             {
-                c.PagesRead = userProgress.Where(p => p.ChapterId == c.Id).Sum(p => p.PagesRead);
+                var progress = userProgress.SingleOrDefault(p => p.ChapterId == c.Id);
+                if (progress == null) continue;
+
+                c.PagesRead = progress.PagesRead;
+                c.ProgressLastModifiedUtc = progress.LastModifiedUtc;
```
majora2007, line 226: You can skip the null check and use null coalescing instead.

c.ProgressLastModifiedutc = progress?.LastModifiedUtc;

Add the non Utc field to be consistent.

majora2007 (changes requested):

API/DTOs/ProgressDto.cs
hgourvest: this is not a breaking change IMHO since this field was actually not filled with any data.

majora2007: ProgressDto is built via AutoMapper and since AppUserProgress has this field, the data will be mapped. So it is a breaking change as you're changing what is being mapped and hence the UI which may rely on that field is now not getting the data.

hgourvest: ok, I understand

API/Controllers/ReaderController.cs
hgourvest: I checked, it doesn't, probably because progressBookmark is instantiated directly instead of being read from the database.

API/Data/Repositories/ReadingListRepository.cs
hgourvest: Because I thought a user can't have multiple reading status for the same chapter and I also needed to read the date. This is not the case?

API/DTOs/ProgressDto.cs
hgourvest: ok

API/Data/Migrations/20230205222635_EntityTimestampsUtc.cs
hgourvest: ok

API/Data/Repositories/ChapterRepository.cs
hgourvest: ok

API/Data/Repositories/ReadingListRepository.cs
hgourvest: ok

API/Data/Repositories/ReadingListRepository.cs
hgourvest: ok

API/Data/Repositories/VolumeRepository.cs
hgourvest: ok

API/Data/Repositories/ReadingListRepository.cs
majora2007: This is true, but there are some old versions that had a bug and may have multiple records, hence why it's done with the sum.

hgourvest: I do not see how the sum can solve this problem since you will have a page number which can exceed the number of pages of the chapter. I'm thinking of using the maximum page number instead. I can't sum dates!

API/DTOs/ChapterDto.cs
hgourvest: ok

API/Data/Repositories/ReadingListRepository.cs
hgourvest: I think the correct solution would be to remove the duplicates from the database and add a uniqueness constraint.

majora2007: Your solution is outside the scope of the PR. You don't need to sum dates, you can just use .Max() to get the latest date. Let's not change any extra code unless it's necessary.

hgourvest: ok

hgourvest:
Is that ok with you, or do you need me to improve something further?

majora2007:
I need to pull it and test it out, just haven't had the time yet. It's on the list for this week.

PR #1824, opened 2023-02-21
Base Url implementation

### LINKED ISSUE #195: Implement Base Url Support
For reverse proxy setups, people might want to use a url route rather than subdomain. 

Implement a server setting which allows setting base url and handle the angular ui to support this (without rebuilding the app).

gmahomarf:
# Added
- Added: Base Url support  (Closes #195)


Kavita.Common/Configuration.cs
```diff
@@ -200,10 +206,60 @@ private static string GetIpAddresses(string filePath)
     }
     #endregion
 
+    #region BaseUrl
+    private static string GetBaseUrl(string filePath)
+    {
+        if (new OsInfo(Array.Empty<IOsVersionAdapter>()).IsDocker)
```
gmahomarf, line 212: I trust that you had a reason for this, so I decided to leave it.

API/Startup.cs
```diff
@@ -351,6 +366,26 @@ public void ConfigureServices(IServiceCollection services)
         });
     }
 
+    private static void UpdateBaseUrlInIndex(string baseUrl)
+    {
+        var htmlDoc = new HtmlDocument();
```
majora2007, line 371: Haha I was thinking of doing this when I was stuck as well, but seemed to be a bad idea.

majora2007:
Upon launching this in my dev environment, without doing any configuration, my env breaks.

http://localhost:4200/http://localhost:5000/api/library

API/Startup.cs
gmahomarf: Hahaha my first thought was to edit the html response in transit every time it was requested, and then I took a step back and realized that it only needed to be changed at startup.

This was the easiest way I could find for everything to work seamlessly. Angular needed to know the `baseUrl` from the start, and there's no simple way to fetch it from the API.

gmahomarf:
> Upon launching this in my dev environment, without doing any configuration, my env breaks.
> 
> http://localhost:4200/http://localhost:5000/api/library

I'll take a look. I was so focused on making it work with a reverse proxy I forgot to check that it worked during development

gmahomarf:
OK, I've made changes to essentially disable `baseUrl` when running in dev mode, as it makes no sense to me to have it enabled. Let me know what you think

majora2007:
I'm about to go on a holiday till end of next week, so I will check this out then. @ThePromidius or @therobbiedavis can one of you give this a spin through docker?

ThePromidius:
i can give this a try on docker.
Any special config or anything i should be looking for?

majora2007:
You just need to set it in the settings page and restart the container to test. Biggest part are ebooks, manga reader, PDF, etc.

majora2007:
If you run this and there is a port conflict, it will write null into the json, whereas if it runs successfully, nothing is written. When null is written, it breaks the install until you remove it.

I'm unable to get this working locally. I setup the server and via Caddy attempted to proxy into it with /kavita/ base url, but assets would not load to bootstrap the app:

```
Loading module from “https://mydomain/kavita/main.414a635bf92adc45.js” was blocked because of a disallowed MIME type (“”).
```

gmahomarf:
> If you run this and there is a port conflict, it will write null into the json, whereas if it runs successfully, nothing is written. When null is written, it breaks the install until you remove it.

Which value will be null in what JSON? Also, if you just run the API on a particular port and open url in that port, does it not serve the index.html and redirect to `/basepath/`? 

> I'm unable to get this working locally. I setup the server and via Caddy attempted to proxy into it with /kavita/ base url, but assets would not load to bootstrap the app:
> 
> ```
> Loading module from “https://mydomain/kavita/main.414a635bf92adc45.js” was blocked because of a disallowed MIME type (“”).
> ```

How are you setting up the proxying? I'm using my branch on my nginx setup to serve on a `/kavita/` base path.

majora2007:
> Which value will be null in what JSON? Also, if you just run the API on a particular port and open url in that port, does it not serve the index.html and redirect to /basepath/?

The baseUrl was null in the JSON (appsettings.json)

No, it does not redirect. It doesn't load at all. 

Using Caddy, my config is:
```
mydomain/kavita {
        encode gzip
        tls myemail
        reverse_proxy /* localhost:5010 192.168.1.83:5010
}
```

gmahomarf:
>

gmahomarf:
The null thing is fixed in my latest commit. As for the other thing, your Caddyfile is incorrect. Per the last paragraph in [this section of the caddy docs](https://caddyserver.com/docs/caddyfile/matchers#path-matchers), your config will only match `mydomain/kavita` **exactly**. The correct config is:

```Caddyfile
mydomain/kavita* { # or mydomain/kavita/*, depending on your needs 
        encode gzip
        tls myemail
        reverse_proxy /* localhost:5010 192.168.1.83:5010
}
```

majora2007 (approved):
Everything looks good after I sorted my config out and you've patched the null issue. @therobbiedavis can you give this a spin on a docker container for me? I want to cover all bases before we merge.

gmahomarf:
@majora2007 Do you have a preferred method for resolving the current merge conflicts? (e.g., rebasing onto develop vs merging develop into my branch)

majora2007:
I'm just waiting on a go ahead from a docker user from the team, then I can merge this.

API/Startup.cs
```diff
@@ -273,6 +274,17 @@ public void ConfigureServices(IServiceCollection services)
 
         app.UseForwardedHeaders();
 
+        var basePath = !Configuration.BaseUrl.StartsWith("/")
```
majora2007, line 277: Put all this logic in the setter, so that it always is set this way. you can also put in the getter to ensure it comes out this way as well in Configuration.

API/Startup.cs
```diff
@@ -333,6 +345,8 @@ public void ConfigureServices(IServiceCollection services)
             endpoints.MapFallbackToController("Index", "Fallback");
         });
 
+        Console.WriteLine("Starting with base url as " + basePath);
```
majora2007, line 348: Use Logger and don't use string concatenation

API/Startup.cs
```diff
@@ -349,6 +363,26 @@ public void ConfigureServices(IServiceCollection services)
         });
     }
 
+    private static void UpdateBaseUrlInIndex(string baseUrl)
+    {
+        var htmlDoc = new HtmlDocument();
+        string indexHtmlPath = Path.Combine(Directory.GetCurrentDirectory(), "wwwroot", "index.html");
```
majora2007, line 369: Use var when possible.

API/Startup.cs
```diff
@@ -349,6 +363,26 @@ public void ConfigureServices(IServiceCollection services)
...
+        htmlDoc.Load(indexHtmlPath);
+
+        var baseNode = htmlDoc.DocumentNode.SelectSingleNode("/html/head/base");
+        if (baseNode != null)
```
majora2007, line 373: We shouldn't need this. index will always have base tag as we control it.

Kavita.Common/Configuration.cs
```diff
@@ -200,10 +206,61 @@ private static string GetIpAddresses(string filePath)
...
+                var baseUrl = tokenElement.GetString();
+                return String.IsNullOrEmpty(baseUrl) ? "/" : baseUrl;
+            }
+        }
+        catch (Exception ex)
+        {
+            Console.WriteLine("Error reading app settings: " + ex.Message);
+        }
+
+        return "/";
```
majora2007, line 234: "/" is hardcoded in many places, extract to a Default variable, like done with IpAddresses. This will allow us to provide a reset, like we did with IpAdresses.

UI/Web/src/app/admin/manage-settings/manage-settings.component.html
```diff
@@ -33,6 +33,19 @@
...
 
+        <div class="mb-3">
+            <label for="settings-baseurl" class="form-label">Base Url (WARNING: Requires server restart)</label>&nbsp;<i class="fa fa-info-circle" placement="right" [ngbTooltip]="baseUrlTooltip" role="button" tabindex="0"></i>
+            <ng-template #baseUrlTooltip>Use this if you want to host Kavita on a base url ie) yourdomain.com/kavita (WARNING: Requires server restart)</ng-template>
+            <span class="visually-hidden" id="settings-cachedir-help">Use this if you want to host Kavita on a base url ie) yourdomain.com/kavita (WARNING: Requires server restart)</span>
+            <input id="settings-baseurl" aria-describedby="settings-baseurl-help" class="form-control" formControlName="baseUrl" type="text"
+            [class.is-invalid]="settingsForm.get('baseUrl')?.invalid && settingsForm.get('baseUrl')?.touched">
+            <div id="baseurl-validations" class="invalid-feedback" *ngIf="settingsForm.dirty || settingsForm.touched">
+              <div *ngIf="settingsForm.get('baseUrl')?.errors?.pattern">
+                  Base URL must start and end with /
```
majora2007, line 44: Per logic, it must start and end in a /, shouldn't it?

UI/Web/src/app/admin/manage-settings/manage-settings.component.html
```diff
@@ -33,6 +33,19 @@
             </div>
         </div>
 
+        <div class="mb-3">
+            <label for="settings-baseurl" class="form-label">Base Url (WARNING: Requires server restart)</label>&nbsp;<i class="fa fa-info-circle" placement="right" [ngbTooltip]="baseUrlTooltip" role="button" tabindex="0"></i>
+            <ng-template #baseUrlTooltip>Use this if you want to host Kavita on a base url ie) yourdomain.com/kavita (WARNING: Requires server restart)</ng-template>
```
majora2007, line 38: Remove the  (WARNING: Requires server restart) and instead add it to the top where it mentions Port change, etc needs server restart.

majora2007 (changes requested):
Seems you've made some enhancements, for which I have some comments on. I've gotten a sign on docker, so please take care of the comments and do another shakeout. I look forward to this addition.

UI/Web/src/app/admin/manage-settings/manage-settings.component.html
gmahomarf: I don't understand what you mean by this

majora2007: In Program.cs, you have code to ensure base url when fetched from appsettings.json starts and ends with /. In that case, might as well put that same logic in your inline validation.

gmahomarf: Ah, got it. It's already, in the validation, using a RegExp. See https://github.com/Kareadita/Kavita/pull/1824/files#diff-d63d0019b68d4b64af3d4280d3d19857e793b8cd7e1b13a29c864b98f9e178c0R50

majora2007: Correct, so don't you think the user should understand that as well? Hence why I want the message updated.

gmahomarf: Oh. I thought "Base URL must start and end with /" was already a good message. Suggestions to make it clearer?

majora2007: Okay my bad, I totally misread that as "Base URL must end with /". Nevermind, so we are good.

majora2007 (approved):

majora2007:
Users are reporting issues with Fetching data from Github for updates, can you please validate that flow?

sulisu:
> Hi, I changed the base url setting in setting pages, but it seems not work.
> 
> I use docker version on nightly-0.7.1.28 which was released on 20230319.
> 
> A fresh new installation. Start with docker-compose, create first user, change base url from "/" to "/kavita/", restart the container.
> 
> The server still works on "/" not "/kavita". If i access ip:5000/kavita/, it will redirect me to "/". The log confirms that the base url setting has not changed, which showed "[Information] API.Startup Starting with base url as / ". I checked the setting in web page, it still shows the right one "/kavita/".
> 
> In data/appsettings.json, there is "BaseUrl": null. I stopped the container, change it to "BaseUrl": "/kavita/". But it also doesn't work after restarting service.
> 
> I have read through this issue, but can not find more instruction on how to setting base url. Did I make some mistakes in the procdeure?

It seems this commit will not work in docker, there is some check in the code. I managed to build a win-x64 version and run it on my win server. It works well now.

Why to place this limitation? I think base url is an applicatioin layer feature and have nothing to do whether the app is running in a container or not.

sulisu:
When reading EPUB, URLs with book-resource dont use the base url.

for example: 
/api/book/3/book-resources?file=..%2fimages%2f00001.jpeg

While there is no problem for URL with book-page and pdf reader is not affected.

majora2007:
@sulisu open appsettings.json and change the baseUrl to "/" or whatever you set yours as. The null was an error that occurred in one nightly and was patched later.

sulisu:
@majora2007 

I had notied the discussions about Null value in appsettings.json. The dockerhub image I tested is nightly-0.7.1.28 which was released around 0320, while this commit is merged on Mar 11.  So i guessed the dockerhub image had this Null patch included. If the nightly image is not based on nightly code, I will try it in docker later.

But I am concerned with this code in develop/Kavita.Common/Configuration.cs

```

public const string DefaultBaseUrl = "/";

...

    #region BaseUrl
    private static string GetBaseUrl(string filePath)
    {
        if (new OsInfo(Array.Empty<IOsVersionAdapter>()).IsDocker)
        {
            return DefaultBaseUrl;
        }
```

Doesn't it mean that if the app find it is running in docker enviroment, it will use DefaultBaseUrl "/" and ignore baseurl setting?

majora2007:
That is correct. I can't recall why I originally had that code, but I did validate via @Kizaing and @therobbiedavis that the code works for Docker instances behind a reverse proxy. 

Please test out the latest nightly (after updating your baseurl like I mentioned) and let me know.

Yidna:
@majora2007 

Hello, I'm here to tack on my confusion/bug report. Like @sulisu I am running the docker image `nightly-0.7.1.28`

Steps taken:
- kavita fresh install, all configs nuked, previous images nuked
- set up admin account, navigated to settings page, and set the `Base Url` to `/kavita/`
- restarted container, nothing seems to have changed
- noted that the `Base Url` field is not supported on Docker using non-root user
- verified that everything is configured via the root user
- edited `appsettings.json` to change `BaseUrl` to `/kavita/`
- restarted container, nothing seems to have changed
- noted the following line on the logs `[Information] API.Startup Starting with base url as /`
- edited `appsettings.json` to change `BaseUrl` to `/kavita`
- restarted container, nothing seems to have changed
- noted the following line on the logs `[Information] API.Startup Starting with base url as /`

`mydomain.com/kavita` continues to not work and `192.168.x.x:5000/kavita` continues to redirect to `192.168.x.x:5000/`

Yidna:
I spun up the repository locally and built two Docker images - the `develop` branch as-is (`kavita:latest`), and the `develop` branch with the conditional removed (`kavita:test`)

Interestingly enough, I seem to still be having the issue on both images, the only difference being that the `kavita:test` log now properly shows `[Information] API.Startup Starting with base url as /kavita/`

🤔

sulisu:
@Yidna
On my working win-x64 build, It is not necessory to edit appsettings.json by hand. Just change Base url setting in web gui, value of base url in appsettings.json will change accordingly.

therobbiedavis:
> * kavita fresh install, all configs nuked, previous images nuked
> * set up admin account, navigated to settings page, and set the `Base Url` to `/kavita/`
> * restarted container, nothing seems to have changed
> * noted that the `Base Url` field is not supported on Docker using non-root user
> * verified that everything is configured via the root user
> * edited `appsettings.json` to change `BaseUrl` to `/kavita/`
> * restarted container, nothing seems to have changed
> * noted the following line on the logs `[Information] API.Startup Starting with base url as /`
> * edited `appsettings.json` to change `BaseUrl` to `/kavita`
> * restarted container, nothing seems to have changed
> * noted the following line on the logs `[Information] API.Startup Starting with base url as /`
> 
> `mydomain.com/kavita` continues to not work and `192.168.x.x:5000/kavita` continues to redirect to `192.168.x.x:5000/`
> 
> I too suspect that https://github.com/gmahomarf/Kavita/blob/develop/Kavita.Common/Configuration.cs#L216 is an issue

@Yidna Do you have access to a root user to test? If so add this to your docker compose:
```
version: '3.9'
services:
    kavita:
        image: kizaing/kavita:latest
        **user: UID:GID**
...
```
Changing the UID and GID to their appropriate values.

For Docker run you can do:
```
docker run --user USER GROUP --name kavita -p 5000:5000 \
...
```

Yidna:
@therobbiedavis I think I figured it out!

@majora2007
Digging through the commit/PR history, it looks like the `OSInfo.IsDocker` check was added in initially for `PORT` - this makes sense given that you would rarely want to change the port of the application directly, when you could just change the port mapping on the container. The check was also added in for IP Address but I don't really understand why the setting would not have been applicable to Docker installs.
https://github.com/Kareadita/Kavita/pull/280
https://github.com/Kareadita/Kavita/pull/1713

For `BaseUrl` I think that `OSInfo.IsDocker` check qualifies more as a bug than anything - I see no reason why Kavita running in Docker should not be able to change the `BaseUrl`... as for why this works on Docker in Ubuntu, it turns out `OSInfo` detects the environment as Linux rather than Docker (did not dig into why, might be a bug)

After gutting all the conditionals on my Windows environment, everything was working after building another Docker image; I did not get around to testing native builds though. You actually merged in another conditional that wasn't part of this PR that I had to remove - I don't think your change was correct. It's not clear to me from the PR comments or commit what the issue you were trying to fix was though. Discussion can continue in that PR
https://github.com/Kareadita/Kavita/pull/1880

PR #1650, opened 2022-11-13
Docker nonroot

mikeymakesit:
# Added
- Added: Ability to provide environment variables PUID and PGID which allow running Kavita as a non-root user in a Docker container.  If you do not specify these variables, Kavita will run as root.  If specified, user and group 'kavita' will be created in the container with the specified UID and GID if they don't already exist. Kavita will then run as that user.

PR #1832, opened 2023-02-23
Revert ability to run docker as non-root due to issues with Synology.

majora2007:
# Changed
- Changed: You can no longer run the docker as non-root (Reverts functionality as it was breaking Synology installs)

Reverted #1650

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,18 +1,19 @@
 #! /bin/bash
 
-# Set default UID and GID for Kavita but allow overrides
-PUID=${PUID:-0}
-PGID=${PGID:-0}
-
-# Add Kavita group if it doesn't already exist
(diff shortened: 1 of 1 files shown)


LeoShivas:
Can you please provide me the last docker image ID before the revert ?

I want to understand why setting `PUID` and `PGID` causes issue in Synology.

I don't understand why [LinuxServer](https://github.com/linuxserver/docker-kavita) is able to make it work and not [Kareadita](https://github.com/Kareadita/Kavita).

majora2007:
Hi @LeoShivas, you can see the changes in this PR. Just take these and see. I don't own a Synology so I cannot help at all. If you are able to figure it out, please do let me know. The users would appreciate being able to use non-root containers.

PR #1871, opened 2023-03-09
Fixes bugs in vertical writing mode.

CKolle:
# Fixed
- Fixed: Fixed an issue with pagination in vertical writing style in the book reader. The issue  was that it scrolled using 'smooth' instead of 'auto'. Caused issues when clicking new page while smooth scrolling. It is also now consistent with horizontal writing style (develop)
- Fixed: Fixed an issue where the bottom bar would appear if you first were in horizontal style and an image overflowed, then changed to vertical style. Fixed by making sure the bottom bar never gets created if you are both in vertical  style and scrolling mode (develop)


majora2007:
I'm seeing a side effect and not sure if I should wait for your other PR, but I do see images on horizontal writing mode no longer centering, whereas before they would. Now they are left aligned.

CKolle:
I am not able to replicate images not being centered. The images you are seeing that are getting left aligned are probably svg images. Currently in stable svg images get left aligned, this PR doesn't change that behavior. Normal images that being images created with the "img" tag should still be center aligned.

majora2007:
Indeed it was via svg.

PR #1872, opened 2023-03-09
Reworks how the image sizes in the book reader gets handled

CKolle:
This is a fairly large PR of changes to how images are handled with the epub reader. Please test this out and report any issues you see, even if they are small.

# Added
- Added: If a page only contains a single image. It gets positioned to the center in column mode
- Added: Added a check for if the current page only contains a single image. Happens during loading as to not have the image pop in place. 

# Changed
- Changed: Previously the preserveAspectRatio got ignored due to lacking max-width. Meaning that whether an image's aspectRatio gets changed is decided by the book. Because of this most cover image's aspect ratio will no longer be preserved 
- Changed: The image sizes is no longer handled by the side effect of getters, instead a separate method.
- Changed: Image now only gets scaled through properties, and no longer looping through every image and setting max-height

# Fixed
- Fixed: Fixed a bug where images would not resize in column-mode vertically due the image escaping the container.
- Fixed: Fixed a bug where in scroll mode, the max-height of an image would not be unset again after returning from column-mode

### DIFF
--- a/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
+++ b/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
@@ -122,7 +122,10 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
    * TODO: See if continuousChaptersStack can be moved into reader service so we can reduce code duplication between readers (and also use ChapterInfo with it instead)
    */
   continuousChaptersStack: Stack<number> = new Stack();
-
+  /*
+   * The current page only contains an image. This is used to determine if we should show the image in the center of the screen.
+   */
+  isSingleImagePage = false;
   /**
    * Belongs to the drawer component
    */
@@ -372,17 +375,14 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
     if (this.layoutMode !== BookPageLayoutMode.Default || this.writingStyle === WritingStyle.Vertical) {
       // Take the height after page loads, subtract the top/bottom bar
       const height = this.windowHeight  - (this.topOffset * 2);
-      this.document.documentElement.style.setProperty('--book-reader-content-max-height', `${height}px`);
       return height + 'px';
     }
-
     return 'unset';
   }
 
   get VerticalBookContentWidth() {
-    if (this.layoutMode !== BookPageLayoutMode.Default && this.writingStyle !== WritingStyle.Horizontal ) {
+    if (this.layoutMode !== BookPageLayoutMode.Default) {
       const width = this.getVerticalPageWidth()
-      this.document.documentElement.style.setProperty('--book-reader-content-max-width', `${width}px`);
       return width + 'px';
     }
     return '';
@@ -592,7 +592,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
           this.libraryType = type;
         });
 
-        this.updateImagesWithHeight();
+        this.updateImageSizes();
 
 
         if (this.pageNum >= this.maxPages) {
@@ -637,6 +637,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
   onResize(){
     // Update the window Height
     this.updateWidthAndHeightCalcs();
+    this.updateImageSizes();
     const resumeElement = this.getFirstVisibleElementXPath();
     if (this.layoutMode !== BookPageLayoutMode.Default && resumeElement !== null && resumeElement !== undefined) {
       this.scrollTo(resumeElement); // This works pretty well, but not perfect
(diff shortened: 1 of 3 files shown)


UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -851,41 +856,78 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
-      maxHeight = (parseInt(this.ColumnHeight.replace('px', ''), 10) - (this.topOffset * 2));
-    } else if (this.layoutMode !== BookPageLayoutMode.Column2 && this.writingStyle === WritingStyle.Vertical) {
-      maxHeight = this.getPageHeight() - COLUMN_GAP;
-    } else if (this.layoutMode === BookPageLayoutMode.Column2 && this.writingStyle === WritingStyle.Vertical) {
-      maxHeight = this.getPageHeight() / 2 - COLUMN_GAP;
+  updateImageSizes() {
+    const isVerticalWritingStyle = this.writingStyle === WritingStyle.Vertical
+    const height = this.windowHeight - (this.topOffset * 2)
+    let maxHeight = 'unset'
+    let maxWidth = ''
```
majora2007, line 872: Please use semicolons in typscript.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -851,41 +856,78 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
-        const pageWidth = this.getVerticalPageWidth()
-        const maxImgHeight = Math.min(maxHeight, pageWidth / aspectRatio);
-        this.renderer.setStyle(img, 'max-height', `${maxImgHeight}px`);
-      }
-    });
+  }
 
+  checkSingleImagePage(safeHtml: SafeHtml) {
+    const parser = new DOMParser();
+    const doc = parser.parseFromString(safeHtml.toString(), "text/html");
```
majora2007, line 917: Use single quotes

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -851,41 +856,78 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    const images = doc.querySelectorAll("img");
+    const text = doc.querySelectorAll("p");
+
+
+    if (images.length > 1 || text.length > 1) {
+      return false
+    }
+    // A p tag could hold the image, and needs to be checked
+    if (text.length) {
```
majora2007, line 926: What if the image is within a div, we might want to expand the tags.

majora2007 (changes requested):
Still doing some testing, but looking good.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
CKolle: You raise a valid point. Forgot that there could be books that have their text inside of "div"s instead of "p"-tags or other kinds of elements. I will re-do this section to handle these scenarios

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -156,6 +156,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
   clickToPaginateVisualOverlay = false;
   clickToPaginateVisualOverlayTimeout: any = undefined; // For animation
   clickToPaginateVisualOverlayTimeout2: any = undefined; // For kicking off animation, giving enough time to render html
+  updateImageSizeTimeout: any = undefined;
```
majora2007, line 159: Don't use any

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -1373,7 +1373,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
 
   updateWritingStyle(writingStyle: WritingStyle) {
     this.writingStyle = writingStyle;
-    this.updateImageSizes()
+    setTimeout(() => this.updateImageSizes());
```
majora2007, line 1376: I'm assuming you want to do this? this.updateImageSizeTimeout =  setTimeout()?

CKolle: No I don't want to do it here. It doesn't need to get debounced. I needed to debounce the other one as it gets called numerous times once the page loads.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
majora2007: Once this is sorted, I'm ready to merge.

majora2007 (approved):

PR #1886, opened 2023-03-17
Donate styling

therobbiedavis:
# Changed
- Changed the styling and positioning of the donate button to still be visible but less intrusive and more cohesive to the application.


.github/workflows/sonar-scan.yml
```diff
@@ -15,7 +15,7 @@ jobs:
           uses: JJ/github-pr-contains-action@releases/v7
           with:
               github-token: ${{ secrets.GITHUB_TOKEN }}
-              bodyDoesNotContain: "[\"|`]"
+              bodyDoesNotContain: "[\"|\']"
```
majora2007: Let's unescape ` and just leave it for now. We might have to pull this action since it doesn't seem to work.

PR #1901, opened 2023-03-29
Bugfix/base url in docker

### LINKED ISSUE #1900: Bug: BaseUrl does not work in all Docker environments
**Describe the bug**
The `BaseUrl` config is disabled in Docker environments where `OsInfo` detects that it's in a Docker environment. The initial Docker check was meant for disabling port changes in Docker environments, which is not useful as you're meant to use port mappings to change the port. Since the majority of config in `Configuration.cs` had the check, it ended up being copied over into the `BaseUrl` getters and setters, even though it is not relevant.

**To Reproduce**
1. Spin up a Docker instance on a Windows/Unraid host
2. Note that the BaseUrl feature does not do anything
3. Spin up a Docker instance on an Ubuntu (22.04) host
4. Note that the BaseUrl feature works

**Expected behavior**
BaseUrl feature should work on any Docker image regardless of host

**Additional context**
Discussion in: https://github.com/Kareadita/Kavita/pull/1824
See also: https://github.com/Kareadita/Kavita/pull/1880

Possible tangential bug is that `OsInfo` detects Docker in Ubuntu as Linux rather than Docker

Yidna:
# Fixed
- Fixed: Fixed a bug where BaseUrl did not work in all Docker environments (Fixes #1900)

### DIFF
--- a/API/Startup.cs
+++ b/API/Startup.cs
@@ -391,7 +391,6 @@ public class Startup
     {
         try
         {
-            if (new OsInfo(Array.Empty<IOsVersionAdapter>()).IsDocker) return;
             var htmlDoc = new HtmlDocument();
             var indexHtmlPath = Path.Combine(Directory.GetCurrentDirectory(), "wwwroot", "index.html");
             htmlDoc.Load(indexHtmlPath);
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Setup guides, FAQ, the more information we have on the [wiki](https://wiki.kavit
 - HTML/Javascript editor of choice (VS Code/Sublime Text/Webstorm/Atom/etc)
 - [Git](https://git-scm.com/downloads)
 - [NodeJS](https://nodejs.org/en/download/) (Node 16.X.X or higher)
-- .NET 6.0+
+- .NET 7.0+
 - dotnet tool install -g --version 6.4.0 Swashbuckle.AspNetCore.Cli
 
 ### Getting started ###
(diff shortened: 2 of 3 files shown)


majora2007:
@Kizaing can you give this a once over and if it's good, we can merge it.

therobbiedavis:
> @Kizaing can you give this a once over and if it's good, we can merge it. 

Hey Chris, can you validate on unraid or an OS outside of Ubuntu? It seems like docker on ubuntu is showing up as Linux instead of docker.

Kizaing:
Yeah for sure, I'll validate it on Unraid and other non Ubuntu linux OS's and get back with my results

Yidna:
Any updates?

Kizaing:
Apologies on the long time, life got a bit busy haha

Everything seems to be working I tested on UNRAID, however OPDS appears to be broken when using Base URL. Tachiyomi works fine since it's using the API, but I tried regular OPDS with both Librera and Moon+ Reader and for both the /kavita/ base url was removed when connecting.

majora2007:
I'll take over the OPDS implementation, since it was missed in the original PR.

fawqsir:
I am using docker image kizaing/kavita:nightly-0.7.1.39 .  How can I change the base url so that I can access from example.com/kavita/? is there an environtment variable? is there a command line option? I cant seem to find any documentation.

Could line 12 in [Configuration.cs](https://github.com/Kareadita/Kavita/blob/develop/Kavita.Common/Configuration.cs)
`public const string DefaultBaseUrl = "/";`

`public const string DefaultBaseUrl = "$BASE_URL/";`

majora2007:
@fawqsir you just update it in the UI and restart

PR #1984, opened 2023-05-11
Updated Docker entrypoint

Kizaing:
# Changed
- Changed: Modified the docker entrypoint script to copy in the appsettings.json config rather than manually write a new one via echo


majora2007 (approved):
Thanks for taking care of this quickly.

PR #2034, opened 2023-06-05
Correct not-read badge position

Hobogrammer:
# Fixed
- Fixed: Fixed the not-read badge not being flush against the edge of the book cover in list-item page.

Before:
image

After:
image


majora2007:
I'm going to merge your fix into my feature branch. Thanks for the fix.

PR #2151, opened 2023-07-22
Change GoogleBooks icon to the real favicon

JShiesty-dev:
# Changed
- Changed: Changed the GoogleBooks ExternalServices icon to the real Google Books icon.


majora2007:
I personally like the current one. It looks better on light and dark themes.

PR #2188, opened 2023-08-07
A few bugfixes

majora2007:
# Changed
- Changed: Lots of localization changes with no user impact
- Changed: Last Added To -> Last Item Added to be more clear in Edit Series info section. 

# Fixed
- Fixed: (Kavita+) Fixed a case where when setting up initial rates for scrobbling, Kavita could print as if a user had rate limit
- Fixed: Tachiyomi should now work again with nightly releases (develop)
- Fixed: Fixed a build step which prevented minification and broke localization code in prod builds (develop)


majora2007:
Hold for:
https://github.com/ngneat/transloco/issues/685
https://github.com/ngneat/transloco/issues/684

PR #2228, opened 2023-08-21
[Docker] Make Kavita executable at Build time instead of Runtime 

### LINKED ISSUE #1889: Entryfile.sh tries to set chmod +x on every run
**Describe the bug**
Line 27 of the entryfile.sh tries to make the executable runnable every time the docker container starts. If the docker image just shipped with that change by default it wouldn't be necessary to do it at launch each time. Moving it to the dockerfile I think makes more sense. 

This would also resolve the issues with errors in the logs each time the container is started 
```
chmod: changing permissions of 'Kavita': Operation not permitted
```

**To Reproduce**
1. Launch the container in docker

**Expected behavior**
The docker image should ship with the file having the right permissions from the start. 

**Screenshots**
image

 - OS: Docker
 - Browser: Any
 - Version: v0.7.1.27

stavros-k:
# Fixed
- Fixed: Makes Kavita program and the entrypoint script executable, which removes the need to do that on every startup of the container, additionally is one step closer (still needs other changes) to potentially make the docker image run as rootless. (Fixes: https://github.com/Kareadita/Kavita/issues/1889)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/*
 
 COPY entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
 
 EXPOSE 5000
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ COPY _output/*.tar.gz /files/
 COPY UI/Web/dist /files/wwwroot
 COPY copy_runtime.sh /copy_runtime.sh
 RUN /copy_runtime.sh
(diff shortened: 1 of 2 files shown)


majora2007:
Please include a PR body like the template requests and information why this is needed.

stavros-k:
> Please include a PR body like the template requests and information why this is needed.

Done, Sorry, github.dev bypasses the template when opening PR from there.

majora2007:
Thanks. I'm on holiday but I'll take a look afterwards.

Kizaing:
I looked it over, looks good to me 👍

Kizaing (approved):

PR #2233, opened 2023-08-24
Add timezone and more comments to docker-compose.yml

ameer1234567890:
# Added
- Added: Added time zone in docker-compose.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,11 +1,16 @@
-version: '3'
+version: '3.9'
 services:
     kavita:
         image: kizaing/kavita:latest
         container_name: kavita
         volumes:
-            - ./manga:/manga
-            - ./config:/kavita/config
+            - ./manga:/manga            # Manga is just an example you can have the name you want. See the following
+            - ./comics:/comics          # Use as many as you want
+            - ./books:/books            #
+            - ./data:/kavita/config     # Change './data if you want to have the config files in a different place.
+                                        # /kavita/config must not be changed
+        environment:
+            - TZ=Your/Timezone
         ports:
             - "5000:5000"
         restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,13 +1,12 @@
-version: '3.9'
 services:
     kavita:
         image: kizaing/kavita:latest
         container_name: kavita
         volumes:
             - ./manga:/manga            # Manga is just an example you can have the name you want. See the following
-            - ./comics:/comics          # Use as many as you want
-            - ./books:/books            #
-            - ./data:/kavita/config     # Change './data if you want to have the config files in a different place.
+            - ./comics:/comics          # Optional
+            - ./books:/books            # Optional
+            - ./config:/kavita/config   # Change './config if you want to have the config files in a different place.
                                         # /kavita/config must not be changed
         environment:
             - TZ=Your/Timezone


majora2007:
Why are you removing port and restart lines? These comments are already found on the wiki and docker site to explain.

Kizaing:
You can actually remove the version: line entirely, the docker compose v2 spec no longer requires it. You can also keep ./config the same, it doesn't need to be renamed to ./data

therobbiedavis:
> Why are you removing port and restart lines? These comments are already found on the wiki and docker site to explain. 

They were not removed, Joe.

A better format would be to mark the additional lines as `#OPTIONAL`. Clear, succinct, and follows the format of basically every other compose.

docker-compose.yml
```diff
@@ -1,11 +1,16 @@
...
     kavita:
         image: kizaing/kavita:latest
         container_name: kavita
         volumes:
-            - ./manga:/manga
-            - ./config:/kavita/config
+            - ./manga:/manga            # Manga is just an example you can have the name you want. See the following
+            - ./comics:/comics          # Use as many as you want
+            - ./books:/books            #
+            - ./data:/kavita/config     # Change './data if you want to have the config files in a different place.
```
majora2007, line 10: Leave this as config instead of data. Comment is fine, but config will make it easier when supporting users.

docker-compose.yml
```diff
@@ -1,11 +1,16 @@
-version: '3'
+version: '3.9'
```
majora2007, line 1: Remove this as it's no longer needed

docker-compose.yml
```diff
@@ -1,11 +1,16 @@
...
 services:
     kavita:
         image: kizaing/kavita:latest
         container_name: kavita
         volumes:
-            - ./manga:/manga
-            - ./config:/kavita/config
+            - ./manga:/manga            # Manga is just an example you can have the name you want. See the following
+            - ./comics:/comics          # Use as many as you want
+            - ./books:/books            #
```
majora2007, line 9: Use OPTIONAL for the comics and books, like Robbie suggested.

majora2007 (changes requested):
Please take care of these comments then we can merge.

majora2007 (approved):

majora2007:
Okay I'm back from holiday. Why are you modifying this docker-compose.yml? This isn't used anywhere in the dev or production code. It seems like you'd want to modify our wiki instead.

majora2007:
That's not a good enough reason, I think we should in fact delete the compose file completely. It's not used as dev resource and I don't want to maintain extra files if there is no point. Users should not be coming to the source code to find files to learn how to run the software. @therobbiedavis and @Kizaing can I get your thoughts on this?

therobbiedavis:
@majora2007 Honestly it is common for open source products to contain the compose file to run within their repo since that is where a lot of people initially go for documentation. I don't see an issue with having it since it's like a one and done file, however if you do not want to maintain it then we should remove it entirely.

majora2007:
I've decided to decline this PR and delete the original file. This PR is testament to the compose file not being needed as timezone was added long ago on the wiki and all other documentation. I honestly didn't even notice we had a compose in the application, otherwise it would have been updated when we updated our documentation. 

I will take your comments and update the wiki documentation as it may help users understand how volume mounts work better. 

Thanks for raising the PR in the first place, enhancement was drawn from your effort, just not in the way initially expected.

PR #2244, opened 2023-08-29
Lots of Filtering Fixes & New Fields

majora2007:
# Added
- Added: You can now filter by Path, which is the highest folder for a Series. 
- Added: You can now filter by filepath, which is the full path of a file, including extension. This is much slower than Path, but still useable. 

# Changed
- Changed: Restored APIs needed by Tachiyomi and CDisplayEx, until more people upgrade to v0.7.8, in which these APIs will be removed. 
- Changed: (Performance) Removed single DB calls in many APIs.

# Fixed
- Fixed: Fixed a ton of filtering bugs, too many to list (develop)


majora2007:
Henri confirmed on-deck needs a v1 implementation. Don't merge until this is taken care of.

PR #2318, opened 2023-10-17
[skip ci] Reduce warnings during tests

senpai-notices:
Reduced warnings during tests, from 30-something to 10.

Some test cases are missing annotations - this had been done previously so the test is skipped.
As such, the remaining warnings relate to missing annotations:
- `X should be marked as a Theory`
- `X should be marked as a Fact`
- `X Test data attribute should only be used on a Theory`

# Developer
- Changed: Reduce warnings during tests

### DIFF
--- a/API.Tests/Parser/ComicParserTests.cs
+++ b/API.Tests/Parser/ComicParserTests.cs
@@ -115,7 +115,6 @@ public class ComicParserTests
     [InlineData("Batgirl Vol.2000 #57 (December, 2004)", "2000")]
     [InlineData("Batgirl V2000 #57", "2000")]
     [InlineData("Fables 021 (2004) (Digital) (Nahga-Empire).cbr", "0")]
-    [InlineData("Cyberpunk 2077 - Trauma Team 04.cbz", "0")]
     [InlineData("2000 AD 0366 [1984-04-28] (flopbie)", "0")]
     [InlineData("Daredevil - v6 - 10 - (2019)", "6")]
     [InlineData("Daredevil - v6.5", "6.5")]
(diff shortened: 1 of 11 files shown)


API.Tests/Parser/ComicParserTests.cs
```diff
@@ -115,7 +115,6 @@ public void ParseComicSeriesTest(string filename, string expected)
     [InlineData("Batgirl Vol.2000 #57 (December, 2004)", "2000")]
     [InlineData("Batgirl V2000 #57", "2000")]
     [InlineData("Fables 021 (2004) (Digital) (Nahga-Empire).cbr", "0")]
-    [InlineData("Cyberpunk 2077 - Trauma Team 04.cbz", "0")]
```
senpai-notices, line 118: Note: this is a duplicate InlineData

API.Tests/Services/CollectionTagServiceTests.cs
```diff
@@ -112,7 +112,7 @@ public async Task GetTagOrCreate_ShouldReturnNewTag()
         await SeedSeries();
         var tag = await _service.GetTagOrCreate(0, "GetTagOrCreate_ShouldReturnNewTag");
         Assert.NotNull(tag);
-        Assert.NotSame(0, tag.Id);
+        Assert.Equal(0, tag.Id);
```
majora2007, line 115: Doesn't this flip what the test is asserting?

API.Tests/Services/CollectionTagServiceTests.cs
```diff
@@ -121,7 +121,7 @@ public async Task GetTagOrCreate_ShouldReturnExistingTag()
         await SeedSeries();
         var tag = await _service.GetTagOrCreate(1, string.Empty);
         Assert.NotNull(tag);
-        Assert.NotSame(1, tag.Id);
+        Assert.Equal(1, tag.Id);
```
majora2007, line 124: Same here with assertion. Seems to be flipped.

majora2007 (commented):
Overall looks good, but a few areas the tests seem to be flipping the assertion.

API.Tests/Services/CollectionTagServiceTests.cs
senpai-notices: Yeah - we actually should be asserting Equal here since we are checking that the tag we just created has the correct Id we had passed through. It did pass with NotSame because `0` and `tag.Id` refer to different objects. Because value types were passed into a method that expects reference types, .NET will convert them into objects.

PR #2323, opened 2023-10-18
Fix `GetTagOrCreate_ShouldReturnExistingTag` test

senpai-notices:
Upon looking at the implementation of `GetTagOrCreate()`, it appears that asserting on `tag.Id` is not enough (introduced in #2318). This is because it doesn't tell us the object was newly created or an existing one. As such, I've added an additional `Assert.Equal` on `tag.Title`. This makes sure that `tag.Title` is equal to an existing tag (that was created during seeding).

# Developer
- Fixed: Fix GetTagOrCreate tests


majora2007:
I retargetted this into my current branch I'm working on. It's good to merge though. Just want to push it with my code.

PR #2398, opened 2023-11-05
fix: silence health checks

javabean:
# Changed
- Changed: Make health check silent by default (`-s` curl option). They will still output a message in case of error (`-S` curl option)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ EXPOSE 5000
 
 WORKDIR /kavita
 
-HEALTHCHECK --interval=30s --timeout=15s --start-period=30s --retries=3 CMD curl --fail http://localhost:5000/api/health || exit 1
+HEALTHCHECK --interval=30s --timeout=15s --start-period=30s --retries=3 CMD curl -fsS http://localhost:5000/api/health || exit 1
 
 ENV DOTNET_RUNNING_IN_CONTAINER=true
 


majora2007:
Just to understand, the current is logging out every interval and the change makes it so it logs out only on error? 

@therobbiedavis @DieselTech

javabean:
> Just to understand, the current is logging out every interval and the change makes it so it logs out only on error?


Absolutly. Current check logs un-necessary data: 

```
$ docker container inspect kavita-1
[snip]
            "Health": {
                "Status": "healthy",
                "FailingStreak": 0,
                "Log": [
                    {
                        "Start": "2023-11-05T17:22:32.61473432+01:00",
                        "End": "2023-11-05T17:22:32.782889167+01:00",
                        "ExitCode": 0,
                        "Output": "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100     2    0     2    0     0    250      0 --:--:-- --:--:-- --:--:--   250\nOk"
                    },

```

This patch removes the check output unless there is an error.

DieselTech:
Honestly I'd like to see the healthcheck moved to being defined in the compose file rather than in the dockerfile. That would align more with the current best practices giving the user control over the check and if they even want to run it to begin with.

```
    healthcheck:
      test: curl -fsS http://localhost:5000/api/health || exit 1
      interval: 30s
      timeout: 15s
      retries: 3
      start_period: 20s
```

Healthchecks that don't have any external monitoring happening are basically useless since docker doesn't take any corrective actions or even notify the user if the check fails. Moving it to the compose file lets users disable the check easier if they don't have monitoring infrastructure setup to handle it. 

```
      healthcheck:
        disable: true
```

majora2007:
@therobbiedavis please advise on what we should do. Since I do not use docker, either solution works for me.

majora2007:
So after talking with Robbie. I'll allow this to be merged as it's not hurting anything. I'm open to Diesel's idea if anyone wants to submit a PR for that.

PR #2420, opened 2023-11-08
[Bug-Fix] Fix last page skipping in 2 column mode

### LINKED ISSUE #2044: Pages getting skipped in 2 column layout for epub files
**Describe the bug**
If the final page of a chapter comes up on the right-hand side of the 2 columns it will occasionally be skipped entirely. Navigating backwards will bring up the skipped page, but it will skip a different page before to get back to the beginning of the chapter. 

This seems to have something to do with margin percentage. For the chapter I'm using it will only behave like this if the margin is 15% or greater, while turning it down prevents it from happening.

**To Reproduce**
* Open a epub file with chapters (it seems to be more prevalent for books with shorter length chapters)
* Set the layout mode to 2 columns and turn the margin to 15% or greater (you might need to play around with the percentages a bit). 
* Go forward in pages until a new chapter, after landing at the beginning of a chapter navigate one back and check if the right columns text has been skipped when moving forward. This is repeatable, going to the original chapter and moving forward until the next chapter will behave the same.

**Desktop**
 - OS: Docker (latest)
 - Browser: Able to repeat on up to date Firefox and Chrome

Hobogrammer:
# Fixed
- Fixed: Fixed book-reader skipping the end of chapters in 2 column mode. (Fixes #2044)

### DIFF
--- a/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
+++ b/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
@@ -1216,7 +1216,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
 
     const [scrollOffset, totalScroll] = this.getScrollOffsetAndTotalScroll();
     const pageSize = this.getPageSize();
-    const totalVirtualPages = Math.max(1, Math.round(totalScroll / pageSize));
+    const totalVirtualPages = Math.max(1, Math.ceil(totalScroll / pageSize));
     const delta = scrollOffset - totalScroll;
     let currentVirtualPage = 1;
 


Hobogrammer:
Use `Math.ceil` instead of `Math.round` to calculate total virtual pages. Rounding causes issues if there is less than half a page remaining. Using ceiling allows partial page amounts to be taken into account. 

I promised tests but it was taking longer than expected and I want to get this fix out for 2 column users.

Hobogrammer:
I promised tests but it was taking longer than expected and I want to get this fix out for 2 column users.

majora2007:
I'll review this soon. I have a few inflight things I'm testing out. Appreciate the PR and looks like a simple fix.

PR #2425, opened 2023-11-10
[Bug Fix] Fix dockerfile not copying correct folder to wwwroot 

TheDarkMage:
# Fixed
- Fixed: Fixed a bug where the dockerfile was copying the wrong folder to wwwroot.

After #2422 UI files are saved to dist/browser, dockerfile was not updated to reflect changes. This was causing the docker image to not be able to load the webpage with the error Could not find file '/kavita/wwwroot/index.html'


majora2007:
Oh this explains it. I changed everywhere BUT the dockerfile. Thanks.

PR #2442, opened 2023-11-19
.kavitaignore no more

### LINKED ISSUE #2394: Removal of .kavitaignore
Kavitaignore files are a complex system to allow the admin to use globbing syntax (which is already non-trivial) to filter out paths and file from the scanning system. 

Likely, users use this to ignore file types, like calibre's cover image in the directory. Folder filtering is likely not used very often. 

This system can be replaced by an advanced library option during setup/edit, which allows the user to have ALL or a mix of Image, Archive, Epub, Pdf (and any further extension support added in the future). 

This system is much easier than maintaining and editing files on their system, allows easier handling when new image types are added, etc. 

Globbing (exclude patterns) can even be added directly into this section as well, if there is a need for it.

majora2007:
image

# Added
- Added: Library Settings now has File types setting which allows you to turn on/off different types of file types from the scanner. 
- Added: Library Settings now has an exclude patterns control which allows you to ignore folders or file patterns. 

# Changed
- Changed: .kavitaignore files have been deprecated and will log a warning to the logs whenever one is found. They will currently still function but be removed in the v0.8 release. 

Closes #2394


janaxhell:
According to the wiki

`**/folder/* will ignore the folder at any depth in a path`

I have many series folders that contain an /Extra subfolder with HTML presentation full of gifs/jpegs and such. With .kavitaignore I used `Extra/*` and all seemed to work fine. Now I have tried the same pattern in the UI, but it doesn't work, so I tried the above example as `**/Extra/*`
and it doesn't work as well. I have manually removed the bogus "series", rescanned with that exclusion and they came back.
My hierarchy is
`/comics/country/publisher/series/Extra` 
with all the .cbz files at the same level of /Extra in
`/comics/country/publisher/series/`

image

I'm using latest 0.8.4.0

majora2007:
> According to the wiki
> 
> `**/folder/* will ignore the folder at any depth in a path`
> 
> I have many series folders that contain an /Extra subfolder with HTML presentation full of gifs/jpegs and such. With .kavitaignore I used `Extra/*` and all seemed to work fine. Now I have tried the same pattern in the UI, but it doesn't work, so I tried the above example as `**/Extra/*` and it doesn't work as well. I have manually removed the bogus "series", rescanned with that exclusion and they came back. My hierarchy is `/comics/country/publisher/series/Extra` with all the .cbz files at the same level of /Extra in `/comics/country/publisher/series/`
> 
> image
> 
> I'm using latest 0.8.4.0

This is not the correct place to bring up a bug. Please file a new issue. Otherwise this will not be worked on.

PR #2450, opened 2023-11-22
[skip ci] Update project link in README

Hobogrammer:
Updating project link in README to the classic version as that's whats being used. 

I'm honestly not certain why that first line is coming up as changed...


majora2007:
Thanks, I just coded this into a separate branch since it's so small.

PR #2464, opened 2023-11-30
[BugFix] Fix clicking on book-line-overlay turning pages

### LINKED ISSUE #2463: [Bug] EpubReader tap pagination activates if the book line overlay is open even with tap pagination disabled
**Describe the bug**
Tap pagination activates if the book line overlay is open even with tap pagination disabled.

**To Reproduce**
1. Open a book
2. Highlight a word so that the book line overlay appears
3. Click on the book line overlay in a tap pagination zone
4. Notice the page turns.

**Expected behavior**
Clicking on the book line overlay should not change pages

    OS: Docker
    Browser: Firefox, KiwiBrowser
    Version: 0.7.10.19

**Additional context**
Only happens if clicking a tap pagination area within the book line overlay.

Hobogrammer:
Quick fix for the book-line-overlay turning pages. Some time should probably be dedicated later to making tap pagination work correctly without having to blacklist classes.

# Fixed
- Fixed: Fix clicking on book-line-overlay turning pages (Fixes #2463)

### DIFF
--- a/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
+++ b/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
@@ -1633,7 +1633,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
   // Responsible for handling pagination only
   handleContainerClick(event: MouseEvent) {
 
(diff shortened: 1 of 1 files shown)


majora2007:
This isn't a good solution. I'll take up the bug today. I agree, Gazy's implementation of the new move page isn't ideal. I'll think if there is a better way to handle fixing the scroll under that they originally fixed with an implementation that doesn't require so many hacks.

PR #2611, opened 2024-01-15
feat: add flat library type

arition:
# Added
- Added: new library type: Flat

For some types like donjinshi and illustration collections, they don't have volumes and chapters, so it doesn't make sense to put them in a folder structure like manga. For this library type, all files will use file name or parent folder name (for loose images) as series name, and they will be assigned a volume number of 1.

### DIFF
--- a/API/Entities/Enums/LibraryType.cs
+++ b/API/Entities/Enums/LibraryType.cs
@@ -1,4 +1,4 @@
-﻿using System.ComponentModel;
+using System.ComponentModel;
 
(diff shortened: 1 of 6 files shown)


majora2007:
I'm not sure about this but at the same time I'm open to this. Can you open a discussion (feat request) and we can brainstorm. Another user mentioned custom regex, so I'd like to expand with the community and see if there is a good solution that meets more needs in one library type.

Issue #2139, opened 2023-07-17
Cover Aspect Ratio Discussion


kruegp:
## Description

This Github issue is to track further discussions related to the [Maintain Cover Aspect Ratio Feature Request](https://feats.kavitareader.com/posts/162/maintain-cover-aspect-ratio).
- This issue was created after recommendation from discord

## Considerations / Uses

- thumbnails for covers (series, volumes, chapters, issues, ...) when displaying in library, search results, or within a series
  - Types of aspect ratios
    -  portrait style
        - Within the [Dune Series in Google Play Books](https://play.google.com/store/books/series?id=av3XFwAAABC86M), the covers range from aspect ratios (`1:1.4` to `1:1.8`)
    - square-ish styles
        - see example of [Garfield Series in Google Play Books](https://play.google.com/store/books/series?id=yMnMGgAAABAF0M)
    - landscape style
        - see example of [Froggy Children's books in Google Play Books](https://play.google.com/store/search?q=froggy%20book&c=books)
  - Different genres (manga, western comics, newspaper comics, books, webtoons, ...) may all have different aspect ratios & notions of covers
      - ie: webtoons may have short height and long width strips as the first file
- thumbnails are used for generating combined image covers for collections

### Displaying thumbnails for assets

- komga
  - [Example](https://feats.kavitareader.com/static/images/attachments/ETIfe3CN572oPZY7ZOJs9gq7795bPHvedD53b0qnjASmk8w85uMPEq9tUya2Y7IX-komga.png?size=200)
  - as mentioned in the original [feature request](https://feats.kavitareader.com/posts/162/maintain-cover-aspect-ratio), komga seems to have a fixed height and width for the asset overlay, and the cover image is scaled to fit within the asset overlay (keeping the aspect ratio) (ie: containing)
    - additionally, the overlay background can be seen since the cover does not fill the overlay
    - additionally, the cover is centered
  - I believe this is similar to [object-fit: contain](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit)
- google search results
  - [Example for Dune Series](https://www.google.com/search?q=frank+herbert+dune+series) 
  - [Example for Garfield Series](https://www.google.com/search?q=garfield+book+series)
  - seems to have a fixed height and width for the asset overlay, and the cover image is scaled to fit the entire asset overlay (keeping the aspect ratio) (ie: cropping)
[...]

PR #2612, opened 2024-01-15
feat: better cover image generation

arition:
# Changed
- Changed: Resize and crop cover image instead of forcing it to resize to a specified resolution (which breaks its original aspect ratio).

### DIFF
--- a/API/Services/ImageService.cs
+++ b/API/Services/ImageService.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -132,7 +132,7 @@ public class ImageService : IImageService
         try
         {
             var dims = size.GetDimensions();
-            using var thumbnail = Image.Thumbnail(path, dims.Width, height: dims.Height, size: Enums.Size.Force);
+            using var thumbnail = Image.Thumbnail(path, dims.Width, height: dims.Height, size: Enums.Size.Both, crop: Enums.Interesting.Attention);
             var filename = fileName + encodeFormat.GetExtension();
             thumbnail.WriteToFile(_directoryService.FileSystem.Path.Join(outputDirectory, filename));
             return filename;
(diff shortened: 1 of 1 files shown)


majora2007:
Please attach before and afters for the cover types listed in #2139.

arition:
## Standard

IMG_001
image

## Square cover

img00116
image

## Another square cover

108942011_p0
image

## Horizontal

img00115
image

## Long

AXVUqdzi3rsC
image

majora2007:
I ran some tests with this and it doesn't behave well enough in my opinion. 

Left is new code, Right is old
image
image
image
image
image

arition:
Yeah, it is not perfect in every situation. At least I think we can give user an option to change between these two modes.

majora2007:
That's not an avenue I'd want to go down. Could you add a check if the file is a webtoon, aka height is at least 2/3 more than width, and use the attention crop code there only?

This would add benefit for webtoon users while keeping it working for comic and Manga users out of the box.

arition:
> That's not an avenue I'd want to go down. Could you add a check if the file is a webtoon, aka height is at least 2/3 more than width, and use the attention crop code there only?
> 
> This would add benefit for webtoon users while keeping it working for comic and Manga users out of the box.

I think it should to be "width 2/3 more than height", `width > height * 5 / 3`?

majora2007:
> I think it should to be "width 2/3 more than height", `width > height * 5 / 3`?

I can't understand the math you're proposing. The goal is to identify if the image is a long panel (webtoon) and apply the new logic based on that.

therobbiedavis:
> > I think it should to be "width 2/3 more than height", `width > height * 5 / 3`?
> 
> I can't understand the math you're proposing. The goal is to identify if the image is a long panel (webtoon) and apply the new logic based on that.

Very confused, would the conditional just be `if (width === (2/3 * height)) { //Width is 2/3 of height }else{ //Width is not 2/3 of height}` ?

arition:
> I can't understand the math you're proposing. The goal is to identify if the image is a long panel (webtoon) and apply the new logic based on that.

Got it. In my opinion, the new method also works well on wide images. Maybe we can apply it when `width > height * 5 / 3 || height > width * 5 / 3`?

majora2007:
Hey I'm planning to take this code into a test branch sometime and tweak the code. I think there can be some benefit to this new code with a little more prodding into the source cover/archive. I want to do some pretty heavy investigation to improve webtoon cover generation and test this against more user's servers. 

I have no plans when (as I'm in the middle of huge code refactors to support new features), but I wanted to share that I do plan to try this PR out in full for potential inclusion.

arition:
Thank you! @majora2007

majora2007:
I ended up taking a look at this and wrote my own logic along with a test suite to help tweak this in the future. Here are some of my results (original, old code, new code):
image
image
image

Issue #2624, opened 2024-01-19
Library Title can be "undefined"


ghost:
### What happened?

I created a new Library with the name "Manga"
Website Header, Site Title & Library Settings showed nothing or "undefined"

### What did you expect?

See the Title of the Library.

### Version

0.7.13.0

### What OS is Kavita being run on?

Docker

### If the issue is being seen on Desktop, what OS are you running where you see the issue?

Windows

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Chrome, Microsoft Edge

### If the issue is being seen on Mobile, what OS are you running where you see the issue?

Android

### If the issue is being seen on the UI, what browsers are you seeing the problem on?

Chrome

### Additional Notes

-

majora2007:
Image from support chat:
![image](https://github.com/Kareadita/Kavita/assets/735851/a6c602a3-5818-4bae-94d9-2ee5967d8f03)

PR #2638, opened 2024-01-20
Fix undefined title issue

charles7668:
# Fixed
- Fixed: Fixed the issue where the tab title is undefined when adding a new library and selecting it. (#2624)


majora2007:
This works, but is a workaround. It would be better to bust the cache of libraryNames so the next request resolves new information from the backend.

PR #2745, opened 2024-02-26
Foundational Rework

### LINKED ISSUE #2099: Reading continues past last chapter / volume and loops back to start
**Describe the bug**
When reading the last chapter or volume in a series, when you get to the very last page the reader will loop back around to chapter 1 again. 

**To Reproduce**
1. Go to the last page in a series
2. Click forward to advance the page
4. Be amazed your back at the start 

**Expected behavior**
In the past there was some kind of warning or popup, can't remember the wording exactly but it was like "You've reached the last page" or "No more pages" 

 - OS: Docker
 - Browser: Firefox 
 - Version: 0.7.3.15 (Edit by Joe: This is present in all versions until this issue is closed)

**Additional context**
So not only does it loop back around to the starting chapter, it starts marking the progress of that chapter as 'in progress" throwing off your resume.

I've also noticed that if your using Volumes and not chapters, that when it loops back to volume 1, that it will be on the last page of volume 1. I was reading Volume 9 and ended on page 150, clicked to advance the page and I was looped back into Volume 1 at page 150. 

Reproducible in /library/2/series/347

### LINKED ISSUE #2722: Preliminary Foundation Changes
In order to even think about supporting the Comic Rework, I need to do a ton of foundational work. 

Chapter:
- [x] Move title generation into DB (localization problems with this idea)
- [x] Add ChapterNumberDisplay field which will be the visual number representation (or symbol for issues), For specials this will be the title (did with Range)
-  [x] Add MinNumber and MaxNumber like Volume has, as floats 
-  [x] Clean up confusion around single epubs not being special (when they should be) so that we can fix up the code to not have code against 0 vs isSpecial
-  [x] entity-title component (lets move logic into pipe) (this is for list item layout)
- [x] Add Sort Order
  - [x] Allow Specials to write to Sort Order then we can render out Title (from comicinfo) over Range on the UI (Ties into ChapterNumberDisplay field)
  - [x] Ensure default Sort Order handles negative issues, issues with duplicates (ch 19, 19.HU, 19.BH, 20), specials, and floats
  - [x] Update all Reader/Series code to use SortOrder and fix unit tests
  - [x] Update Cover Code to respect SortOrder

Volume:
-  [x] New field for DisplayTitle (SeriesService.RenameVolumeName) (updated Name to now include ranges and if 1.0 -> 1) 
-  [x] Introduce a new Volume number that is for Specials only and leave Loose Leafs in another (Volume for Specials, Volume for Loose Leaf, Volume for Annuals)
- [x]  We need to look at all tests and ensure they are appropriately constructed with the new special volume.
  - [x] SeriesExtensionsTests
  - [x]  ReaderServiceTests
  - [x] VolumeListExtensionsTests
  - [x] SeriesServiceTests
  - [x] TachiyomiServiceTests
  - [x] SeriesExtensionsTests

Manual Migrations:
- [x] Migrate Range -> Min/Max Number
- [x] Migrate MangaFile to have FileName filled out
- [x] Migrate Chapter default number (0 -> new number)
- [x] Migrate Loose Leaf volume to new number
- [x] Migrate specials to a new volume and try to retain progress (Note: Users will have to run force scans after, but most progress migrated smoothly, <5% of data could be lost depending on if some bugs were present in book libraries)
- [x] Scrobble Events need to be reencoded with the new Special and Loose leaf volumes by updating any Volume 0 and Chapter 0 with their appropriate combos (scrobble will process as this, might want to leave as legacy)

Misc:
- [x] Scrobble Page needs a UI update to reflect new encoding and how it reports what Kavita is scrobbling up (Volume -100000 Chapter -100000 will not look good)
- [x] GetEstimatedChapterCreationDate Unit Tests
- [x] NextExpectedChapter ngIf needs updating
- [x] OPDS title handling for PDFs (specials). New encoding isn't taken care of

Apps:
- [ ] Kavya
- [ ] Aidoku
- [ ] Tachiyomi
- [ ] CDisplayEx
- [ ] Komf

Notes:
- Users will have to do a hard refresh on all libraries in order for things to work correctly after this PR

Prod Testing Phase Issues: 
- [x] 19.HU type issues aren't showing up from the scanner
- [x] Batman Beyond is showing a special that's duplicated and has no files (http://localhost:4200/library/2/series/1)
- [ ] This special is getting scanned as it's own series
image
- [x] Above special showing volume encoding on the UI 
- [ ]

majora2007:
This is a massive refactor handling (80 pure dev hours) how Loose Leaf Chapters and Specials are stored within Kavita and allowing any chapter number, including negatives. This also adds support for Sort Order, which will allow for future UIs to allow custom ordering of chapters, for comic users. 

This is being pushed to canary as there is a chance of data loss. Do not use this unless you are actively engaged in testing this. 

Use at your own risk. High chance of data loss. 

Please allow for all migrations to run on startup. They may take over 10 minutes. 

# Added
- Added: Chapters now sort by SortOrder. SortOrder is generated from chapter number, but can be overridden (in a future update). This allows for custom sorting for issues that aren't purely numerical. SortOrder will handle cases like 19, 19.HU, 19.BD, 20 without any user intervention. 
- Added: You can now have Chapter 0, -1, etc. 

# Changed
- Changed: Loose Leaf volumes are now encoded with the number -100000 instead of 0. This now means users can have a Volume 0 without issue.
- Changed: Specials are now stored in their own volume with number 100000. 
- Changed: Streamlined a lot of the code for generation of numbers for the UI
- Changed: Specials will now order by SP marker numbers, else fallback to natural sort

# Fixed
- Fixed: Fixed a bug with looping around to volume 1 at the end of a reading session when it shouldn't (Fixes #2099)

# API
- VolumeDto reverted the float on Number field back to int
- ChapterDto Range will be the chapter number. If a special, will be the special name. MinNumber can always be checked for the special encoding
- Encoding: Special (100000), Loose Leaf Volume (-100000). 
- stats/user/reading-history has a dto change, making ChapterNumber a float instead of a string.
- Tachiyomi routes now use their own Dto to avoid having to break apis there. 

Implements #2722


API/Data/ManualMigrations/ManualMigrateMixedSpecials.cs
```diff
@@ -0,0 +1,147 @@
...
+                foreach (var specialChapter in specialChapters)
+                {
+                    // Create a new progress event from the existing and store the ID of the existing progress event to delete it
+                    var newProgress = new AppUserProgress
+                    {
+                        AppUserId = distinctVolume.ProgressRecord.AppUserId,
+                        VolumeId = newVolume.Id,
+                        SeriesId = seriesId,
+                        ChapterId = specialChapter.Id,
+                        PagesRead = distinctVolume.ProgressRecord.PagesRead
```
majora2007: I need to migrate the old reading date information over as well.

majora2007:
I am now doing validation testing against my Prod DB for data loss and validation before this goes to canary for team testing. 

Edit: Everything looks fine, no observed data loss.

PR #2834, opened 2024-04-08
fix: don't translate bad-credentials messages

### LINKED ISSUE #2829: Login Form Username Enumeration Vulnerability
### What happened?

I noticed that the login form displays different error messages for incorrect usernames and passwords when the language is set to Japanese. This behavior might be leveraged by attackers to enumerate valid usernames.

In the Japanese setting, the error message "資格情報が正しくありません" is shown for a correct username with an incorrect password, while an English message is displayed for an incorrect username.

### What did you expect?

To mitigate the risk of username enumeration, consider using a generic error message for both incorrect username and password scenarios. For example:

"ユーザー名またはパスワードが正しくありません。"
(Translation: "Username or password is incorrect.")

By providing a consistent error message, it becomes more challenging for potential attackers to differentiate between valid and invalid usernames.

### Kavita Version Number - Don't see your version number listed? Then your install is out of date. Please update and see if your issue still persists.

0.7.14 - Stable

Docker (Dockerhub Container)

None

_No response_

None

_No response_

### Relevant log output

_No response_

### Additional Notes

_No response_

YodaDaCoda:
# Fixed
- Fixed: Avoid possible username enumeration in login or reset-password endpoints due to message translations (Fixes #2829)

### DIFF
--- a/API/Controllers/AccountController.cs
+++ b/API/Controllers/AccountController.cs
@@ -37,6 +37,8 @@ namespace API.Controllers;
 /// </summary>
 public class AccountController : BaseApiController
 {
+    private static readonly string BadCredentialsMessage = "Your credentials are not correct";
+
     private readonly UserManager<AppUser> _userManager;
     private readonly SignInManager<AppUser> _signInManager;
     private readonly ITokenService _tokenService;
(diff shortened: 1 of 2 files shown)


API/Controllers/AccountController.cs
```diff
@@ -37,6 +37,8 @@ namespace API.Controllers;
 /// </summary>
 public class AccountController : BaseApiController
 {
+    private static readonly string BadCredentialsMessage = "Your credentials are not correct";
```
majora2007, line 40: Can you add a comment here referencing the issue and mention avoiding localization which can cause multiple enumeration.

majora2007 (changes requested):
Please fix the one issue then I can merge this

PR #2838, opened 2024-04-10
Clear font settings before rendering the page

gozilla-paradise:
# Behavior
- When the EPUB has the following CSS style, the reader cannot override the font-size while reading. (Font-size won't change when scrolling the bar)
```
.p_normal {
     font-size: 26px;
}
```

# Purpose solution
- Clean the font-size, family before rendering the page

### DIFF
--- a/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
+++ b/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
@@ -931,6 +931,8 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
       this.updateSingleImagePageStyles()
       this.page = this.domSanitizer.bypassSecurityTrustHtml(content); // PERF: Potential optimization to prefetch next/prev page and store in localStorage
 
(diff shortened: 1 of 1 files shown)


majora2007:
This isn't a solution in my opinion as it will break more comics that help. Can you not just add important to the overrides?

gozilla-paradise:
image
The reader setting is already having important suffix, but it seems that the book's css font-size has higher priority.

I'm try moving pageLevelStyle font-size to elementLevelStyle and it seems to solve the issue. But I wonder, if this break anything ?
image

therobbiedavis:
You need to understand css specificity. https://web.dev/learn/css/specificity
When there are two competing rules on the same element, generally speaking, the style selector closest to the element will be applied and inline styles always take precedent. There is a scoring technique used, but that's generally the gist of it.

Now what you are trying to do may impact books that have decorative font stylings, such as the following example:
image

You can use this to test against: https://www.gutenberg.org/ebooks/23450

gozilla-paradise:
So in this case I think it probably be my book issue.

PR #2896, opened 2024-04-21
Fix images not loading in books when using a base url

Fesaa:
# Fixed
- Fixed: Fixes images not loading in books when using a custom base url

The way baseUrl was constructed did not take the configuration option into account, this caused all requests to fail. I didn't notice any style missing, but looking at the code; this would also fix that.

<details>
<summary> Before</summary>
image
</details>
<details>
<summary> After</summary>
image
</details>

### DIFF
--- a/API/Controllers/BookController.cs
+++ b/API/Controllers/BookController.cs
@@ -153,7 +153,7 @@ public class BookController : BaseApiController
         if (chapter == null) return BadRequest(await _localizationService.Translate(User.GetUserId(), "chapter-doesnt-exist"));
         var path = _cacheService.GetCachedFile(chapter);
 
-        var baseUrl = "//" + Request.Host + Request.PathBase + "/api/";
+        var baseUrl = "//" + Request.Host + Request.PathBase + Configuration.BaseUrl + "/api/";
 
         try
         {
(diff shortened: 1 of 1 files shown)


API/Controllers/BookController.cs
```diff
@@ -153,7 +153,9 @@ public async Task<ActionResult<string>> GetBookPage(int chapterId, [FromQuery] i
         if (chapter == null) return BadRequest(await _localizationService.Translate(User.GetUserId(), "chapter-doesnt-exist"));
         var path = _cacheService.GetCachedFile(chapter);
 
-        var baseUrl = "//" + Request.Host + Request.PathBase + "/api/";
+        string basePath = Configuration.BaseUrl;
```
majora2007: Please stick to coding style of the codebase. We use var and string.Empty. 

This line here is not needed. You would rewrite this as:
`var basepath = Configuration.BaseUrl == Configuration.DefaultBaseUrl ? string.Empty : Configuration.BaseUrl;`

majora2007 (changes requested):
I will test this out. Last time I had checked, PathBase has the base url string in it.

API/Controllers/BookController.cs
Fesaa: Done :)

Fesaa:
Hi, sorry for the trouble. Hope you haven't tried testing this yet, I did some dumb stuff in my nginx config which broke it. Fixing this, fixed it as well :pensive:

majora2007:
Haha no problem. Glad you got it sorted.

PR #1426, opened 2022-08-11
Lots of Bugfixes

majora2007:
# Added
- Added: Added Jump bar to reading list page

# Changed
- Changed: Downgraded Docnet to 2.4.0-alpha2 which is the version with our patches that enable pdfs to work on ARMx64 (Fixes #1393)
- Changed: Updated all but one API in collection control to admin only policy
- Changed: Updated some headers on Kavita to tighten security
- Changed: Tightened up cover upload to restrict more APIs to the admin
- Changed: Resetting your own password now requires you to pass your current password (if already authenticated). Admins can still change passwords for the user without this restriction.
- Changed: Removed an additional copy in the build script and changed a copy to use appsettings.json from build rather than development.json

# Fixed
- Fixed: Fixed a regression from hotfix where bookmarks were not able to load due to a missing [AllowAnonymous]
- Fixed: Fixed a bug where some config folders wouldn't be created on first load (Fixes #1408 )
- Fixed: Fixed a regression which broke Plugins (and Tachiyomi) from authenticating
- Fixed: Fixed up the caching for images when doing the cover reset flow. Chrome-based browsers do not respect cache-control headers and require a hard refresh for new cover images to show. (develop)
- Fixed: Fixed a bug where clicking on a Jump key wouldn't go exactly to the selected letter (Fixes #1404 )
- Fixed: Fixed a bug where . wasn't being included in # JumpKey
- Fixed: Fixed a bug where the add to reading list modal would overflow and make create button hidden
- Fixed: Remove from Want to Read wasn't working (Fixes #1424 )
- Fixed: Fixed some inconsistencies with clearing multiple series of bookmarks in one go by building a dedicated api (Fixes #1403 )
- Fixed: Fixed a bad naming case for reading list items where it was just a volume (Fixes #1396)
- Fixed: Fixed a bug where there could be duplicate scrollbars on pages with virtual scroller (Fixes #1401 )

PR #2958, opened 2024-05-16
Stylizing the theme viewer

therobbiedavis:
# Changed
- Changed: Changed how the Theme viewer looks and feels

Before:
image

After:
image


majora2007:
This will be taken care of in #1426

therobbiedavis:
Same here, should be #2960

PR #2980, opened 2024-06-05
Feature/manga scroll width changer

Marsimplodation:
# Added
- Added: (Manga Reader) Added the ability to manually override the width the image can scale to. Not applicable to double page layout modes. (Feature Request #2558)

the ability to change the width of manga pages (video 1)
current bugs with this: changing the width scrolls the pages due to the canvas size changing (video 2)

https://github.com/Kareadita/Kavita/assets/46057569/ec599e95-5be7-4a05-a0b4-57ba35f7fc4f

https://github.com/Kareadita/Kavita/assets/46057569/9a57681a-f601-41c1-b99f-6f1309ec405c


UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -193,7 +193,9 @@
                 <option value="full-height">{{t('height')}}</option>
                 <option value="full-width">{{t('width')}}</option>
                 <option value="original">{{t('original')}}</option>
+                <option value="manual">{{t('manual')}}</option>
               </select>
+              <input id="page-fitting-slider" type="range" min=0 max=100 class="form-range visually-hidden" formControlName="widthSlider">
```
majora2007: You're going to need to have a label that's invisible for this. You need "" around your values for attr="value".

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -513,6 +513,7 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
         autoCloseMenu: new FormControl(this.autoCloseMenu),
         pageSplitOption: new FormControl(this.pageSplitOption),
         fittingOption: new FormControl(this.mangaReaderService.translateScalingOption(this.scalingOption)),
+        widthSlider: new FormControl(70),
```
majora2007: I'd like to have this disabled by default.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -549,8 +550,31 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
         takeUntilDestroyed(this.destroyRef)
       ).subscribe(() => {});
 
+      this.generalSettingsForm.get('widthSlider')?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
+        //update the css width
+        const styleSheets = document.styleSheets;
```
majora2007: This isn't needed. You just need the width value from this. So:

```
this.widthOverride$ = this.generalSettingsForm.get('widthSlider')?.valueChanges.pipe(
map(val => val < -1 ? 'initial' : val + '%'),
takeUntilDestroyed(this.destroyRef)
);
```

This would give you back the string value. Then on the image you'd do style={{widthOverride$}}. 

But the thing is, the image isn't in this component. It is rendered out in the renderers. So while the control is here, it would need to populate the readerSettings and the above code would need to be in each renderer. 

You probably also need to disable the input for double page layout.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -549,8 +550,31 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+            if(!rule ||
+              !(rule.cssText.includes("img[_ngcontent-"))) continue;
+            rule.style.width = val + "%";
+          }
+        }
+      });
 
-
+      this.generalSettingsForm.get('fittingOption')?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
+          let slider = document.getElementById("page-fitting-slider") as HTMLInputElement;
```
majora2007: You don't need to do this type of thing in Angular. We should enable/disable only. We shouldn't hide it from the UI. 

Also keep the subscribe blocks empty. Put this in a tap() so that subscribe can be empty.

UI/Web/src/app/manga-reader/_models/reader-enums.ts
```diff
@@ -1,7 +1,8 @@
 export enum FITTING_OPTION {
     HEIGHT = 'full-height',
     WIDTH = 'full-width',
-    ORIGINAL = 'original'
+    ORIGINAL = 'original',
+    MANUAL = 'manual'
```
majora2007: I don't think we need this at all. I think we can just have a slider that applies an arbitrary width with the user is in WIDTH fitting.

majora2007 (changes requested):
Overall, the idea is pretty good but needs a level of polish and testing of course against all the different renderers and how it plays with things like swipe to paginate. 

If you're willing to do the work, I'm willing to merge this in for users.

Marsimplodation:
> Overall, the idea is pretty good but needs a level of polish and testing of course against all the different renderers and how it plays with things like swipe to paginate.
> 
> If you're willing to do the work, I'm willing to merge this in for users.

I'd love to the work, I will just need a while to make these changes. I will get back to you when they are done

Marsimplodation:
the last commit  changes the feature to only work on the infinite scroll manga-reader for now, while fixing the scrolling issue.
More renderers to come

Tested on mobile and desktop

https://github.com/Kareadita/Kavita/assets/46057569/92fa2f7c-e4b1-46ac-b442-04f3d7c2e498

Marsimplodation:
feature now works in all three single page modes (up/down, right/left, infinite scroll) and is disabled/does nothing in everything else, as it has no implementation there
for up/down, right/left the width changing will now work when the height fitting option is used, due to it defeating the point.
I did not experience an issue with swiping to scroll with this

Also are we sure we don't want to hide the slider per default and only show it for the supported modes? If yes the feature works on my part, let me know if there is a change that needs to be done

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
```diff
@@ -53,6 +55,11 @@ export class SingleRendererComponent implements OnInit, ImageRenderer {
   pageNum: number = 0;
   maxPages: number = 1;
 
+  /**
+   * Width overwrite for maunal width control
+  */
+  widthOverride$ : string = 'none';
```
majora2007: The $ sign means that the variable is an observable. The idea is to map it to the widthSlider valueChanges and use map to transform it to the style you need.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
```diff
@@ -67,6 +74,28 @@ export class SingleRendererComponent implements OnInit, ImageRenderer {
...
+      if (FITTING_OPTION.HEIGHT == scaling) return;
+
+      if(this.widthOverride$ == 'none') img.style.width = "";
+      else img.style.width = this.widthOverride$;
+    });
+    //reset manual width when fitting option changes
+    this.mangaReaderComponent.generalSettingsForm.get("fittingOption")?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
+      var img = this.document.getElementById('image-1');
+      if(!img) return;
+      img.style.width = "";
```
majora2007: Please always use single quotes in javascript for this project.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
```diff
@@ -67,6 +74,28 @@ export class SingleRendererComponent implements OnInit, ImageRenderer {
...
+      if(!img) return;
+
+      var scaling = this.mangaReaderComponent.generalSettingsForm.get("fittingOption")?.value;
+      if (FITTING_OPTION.HEIGHT == scaling) return;
+
+      if(this.widthOverride$ == 'none') img.style.width = "";
+      else img.style.width = this.widthOverride$;
+    });
+    //reset manual width when fitting option changes
+    this.mangaReaderComponent.generalSettingsForm.get("fittingOption")?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
```
majora2007: This isn't how you should code it. Don't pass the parent component into the child. You can get fittingOption from readerSettings$ and if you can't, update the code so it's packed in the ReaderSetting interface.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
```diff
@@ -67,6 +74,28 @@ export class SingleRendererComponent implements OnInit, ImageRenderer {
...
+      val = (val < 0) ? 'none' : val + "%";
+      this.widthOverride$ = val;
+
+      var img = this.document.getElementById('image-1');
+      if(!img) return;
+
+      var scaling = this.mangaReaderComponent.generalSettingsForm.get("fittingOption")?.value;
+      if (FITTING_OPTION.HEIGHT == scaling) return;
+
+      if(this.widthOverride$ == 'none') img.style.width = "";
```
majora2007: We don't need to set things on the image like this. This should map to a width value (percentage or none) and then that can be set directly on the  tag with style="width: {{widthOverride$ | async}}". This will then be "width: none" or "width: 20%"

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -194,6 +194,8 @@
                 <option value="full-width">{{t('width')}}</option>
                 <option value="original">{{t('original')}}</option>
               </select>
+              <label for="page-fitting-slider" class="form-label"></label>
```
majora2007: Label needs text inside it. I'm open to the label name.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -194,6 +194,8 @@
...
+              <input id="page-fitting-slider" type="range" min="-1" max="100" class="form-range" formControlName="widthSlider">
```
majora2007: What is the min of -1 for? Wouldn't 0 mean "none"?

majora2007 (changes requested):
Still needs a bit of work. Please also test against the double readers. The input needs to be disabled for those modes very likely.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
Marsimplodation: can you explain this further? I am not experienced in this area and when I try this like you say I get "cannot assign to 'style' because it is an import.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
Marsimplodation: so from the mangareader component updating the setting and in the renderers subscribing to the settings value to make changes?

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
majora2007: The idea is that in Angular we don't set things on DOM nodes directly, so instead of setting width like this through logic, instead we want to store the width value in an observable (widthOverride$) then in the template, directly use that value like I showed above. 

You can see something similar for imageFitClass$.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
majora2007: Yes, see imageFitClass$ in the double-renderer and my comment below.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
Marsimplodation: I see, as someone normally only doing procedural programming in C++ this whole webdev angular stuff is very confusing to me. Hopefully I get the code to your liking the next time
The logic itself should work fine, even deployed it on my own server.

majora2007:
One thing I also just thought about in addition to double page layout having it disabled, canvas renderer (which is used for SplitOptions.LTR/RTL) needs to disable the field as well. 

So to recap, the field should be disabled whenever:
SplitOption is LtR/RtL,
Renderer is any form of Double
Fit is for Height (and Original perhaps)

Marsimplodation:
Hey, did the rewrite in the style you wanted and this feature is only enabled for, specific combinations:
single renderer & Fit to screen splitting & width option
infinite scroll renderer & fit to screen splitting & width option
otherwise it sets the slider value to 0, which resets the width override and disables the slider

Hope this is more to your liking

UI/Web/src/app/manga-reader/_components/infinite-scroller/infinite-scroller.component.ts
```diff
@@ -232,6 +239,26 @@ export class InfiniteScrollerComponent implements OnInit, OnChanges, OnDestroy,
...
+      map(values => (parseInt(values.widthSlider) <= 0) ? '' : values.widthSlider + '%'),
+      takeUntilDestroyed(this.destroyRef)
+    );
+
+    //perfom jump so the page stays in view
+    this.widthSliderValue$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
+      this.currentPageElem= this.document.querySelector('img#page-' + this.pageNum);
+      if(!this.currentPageElem) return
+      var images = Array.from(document.querySelectorAll('img[id^="page-"]')) as HTMLImageElement[];
+      for(let i = 0; i < images.length; i++) {
```
majora2007: Please use newlines after `if (...) branch` and before for loops and return statements. Also new line after a closing `}`. It helps make reading the code much easier.

majora2007: You probably just want to use a foreach loop so we don't need ` var img = images[i];`

UI/Web/src/app/manga-reader/_components/infinite-scroller/infinite-scroller.component.ts
```diff
@@ -232,6 +239,26 @@ export class InfiniteScrollerComponent implements OnInit, OnChanges, OnDestroy,
...
+        var img = images[i];
```
majora2007: Don't use `var`. Always use `let` and `const`.

UI/Web/src/app/manga-reader/_components/infinite-scroller/infinite-scroller.component.ts
```diff
@@ -232,6 +239,26 @@ export class InfiniteScrollerComponent implements OnInit, OnChanges, OnDestroy,
...
+        this.renderer.setStyle(img, "width", val);
+      }
+      this.widthOverride$ = this.widthSliderValue$;
+      this.prevScrollPosition = this.currentPageElem.getBoundingClientRect().top;
+      this.currentPageElem.scrollIntoView();
```
majora2007, line 263: You need a cdRef.markForCheck() to tell Angular we've changed state variables and a re-render may be needed.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -194,6 +194,8 @@
                 <option value="full-width">{{t('width')}}</option>
                 <option value="original">{{t('original')}}</option>
               </select>
+              <label for="page-fitting-slider" class="form-label">Manual Width</label>
```
majora2007: You need localization on the label.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -194,6 +194,8 @@
...
+              <input id="page-fitting-slider" type="range" min="0" max="100" class="form-range" formControlName="widthSlider">
```
majora2007: I think we shouldn't allow 0. I think range from 1-100 is appropriate.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -549,8 +550,25 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
         takeUntilDestroyed(this.destroyRef)
       ).subscribe(() => {});
 
+      this.generalSettingsForm.get('pageSplitOption')?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
+        var fitting = this.generalSettingsForm.get('fittingOption')?.value;
+        if(PageSplitOption.FitSplit == val && FITTING_OPTION.WIDTH == fitting) {
+          this.generalSettingsForm.get('widthSlider')?.enable();
+        } else {
+          this.generalSettingsForm.get('widthSlider')?.setValue(0);
```
majora2007: Shouldn't this default to 100%?

majora2007 (changes requested):
I'm pulling this down to give it an actual test as well. Code is looking pretty good, just a few minor things.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
Marsimplodation: refering to the other comment as well, you suggested to let 0 be the value to turn it off, so if 0 is inputed the default behaviour of the width mode is restored. That is also why it defaults to 0 instead of 100%

majora2007:
Some follow up:
Let's move the slider down. Although it's tied to width, it doesn't look good in the current location. This will also give it padding (otherwise you should have added a mt-2)
image

When I switched from Fit to Hight to Width, the slider value was 0, but the functionality was 100%. 

I believe the slider can (but we might need to do with css) but it might be nice to have the 0% be disabled and 100% be default. 0% and 100% are honestly the same though, are they not? 

I'm seeing this disabled on the infinite scroller if I came from a mode in which it wasn't fit to width. I think we should have an override if we are in infinite scroller mode to enable it. 

I'm seeing the continuous reader (scroll into to change chapter) broken.

I'm seeing on next chapter load, the width is not being set immediately and needs a value change to reapply (this is in webtoon reader). 

These are preliminary results from testing. I'll do more in a bit.

UI/Web/src/assets/langs/en.json
```diff
@@ -1695,7 +1695,8 @@
         "image-scaling-label": "Image Scaling",
         "height": "Height",
         "width": "Width",
-        "manual_width": "Manual Width",
+        "width_override": "Width Override",
```
majora2007: You need to match the naming scheme of the other localization keys. override-width-label and off should be the keys.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -568,6 +572,13 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
         this.generalSettingsForm.get('widthSlider')?.setValue(0);
         this.generalSettingsForm.get('widthSlider')?.disable();
       }
+
+            this.widthOverrideLabel$ = this.readerSettings$?.pipe(
```
majora2007: Fix indentation

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -278,6 +276,13 @@
                      min="10" max="100" step="1" formControlName="darkness">
             </div>
 
+            <div class="col-md-6 col-sm-12">
+              <label for="page-fitting-slider" class="form-label">{{t('width_override')}}:
+                  {{ (widthOverrideLabel$ | async) ? (widthOverrideLabel$ | async) : t('width_override_off') }}
```
majora2007: We can probably re-write this completely as:
@If (widthOverrideLabel$ | async; as widthOverrideLabel) {
{{ widthOverrideLabel ? widthOverrideLabel : t('off') }}
}

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -278,6 +276,13 @@
...
+              </label>
+              <input id="page-fitting-slider" type="range" min="0" max="100" class="form-range" formControlName="widthSlider">
```
majora2007: let's rename these ids to be width-override-slider instead

majora2007:
After your last push, I see a few more issues:
Infinite Scroller: 
 - Continuous reader (next) doesn't trigger anymore 
 - You can now change the Fitting which previously didn't do anything, now can break the webtoon reader

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -549,9 +554,33 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
         takeUntilDestroyed(this.destroyRef)
       ).subscribe(() => {});
 
+      this.generalSettingsForm.get('pageSplitOption')?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
+        const fitting = this.generalSettingsForm.get('fittingOption')?.value;
+        if(PageSplitOption.FitSplit == val && FITTING_OPTION.WIDTH == fitting) {
```
majora2007: space after if. `If (` and also newline above if statements please

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -549,9 +554,33 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
         takeUntilDestroyed(this.destroyRef)
       ).subscribe(() => {});
 
+      this.generalSettingsForm.get('pageSplitOption')?.valueChanges.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(val => {
```
majora2007: Let's move the subscribe into a tap so that subscribe can be empty. 

In addition, move each pipe into a new line, like you see elsewhere. 

I would also get the control in a variable then re-use in the code, so:
const widthOverrideControl = this.generalSettingsForm.get('widthSlider')!;

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
majora2007: Let's add a bit of documentation about our decisions about when this should work. It will likely help me in the future when there is a bug.

UI/Web/src/app/manga-reader/_components/single-renderer/single-renderer.component.ts
```diff
@@ -53,6 +53,11 @@ export class SingleRendererComponent implements OnInit, ImageRenderer {
   pageNum: number = 0;
   maxPages: number = 1;
 
+  /**
+   * Width overwrite for maunal width control
```
majora2007: Width override*

majora2007 (changes requested):
A few more comments then I'm good to merge

Marsimplodation:
did these changes

majora2007 (approved):
Looks good. I'll merge this in and get feedback from nightly users. Appreciate the work.

Issue #2982, opened 2024-06-05
In certain situations, the night version "Exclude Pattern" function does not work


ytoaa:
### What happened?

I'm having trouble with the latest branch that came up about a few hours ago. I'm not having trouble with the branch from a week ago.

jmg/*

In the same grammar above, images and compressed files in the jmg subfolders should not be detected, but the problem of detecting occurs

Nightly Testing Branch

majora2007:
Can you attach logs from during the scan that shows `jmg/` directory being scanned?

majora2007:
I just tested on mine. Library root is /test/. 

Exclude pattern is "One Piece/*".

test/One Piece/My Series/My Series.cbz

and everything in One Piece is skipped.

ytoaa:
[Kavita] [2024-06-06 04:47:49.613 +09:00  51] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP POST /api/series/v2 responded 200 in 21.6353 ms
[Kavita] [2024-06-06 04:47:49.617 +09:00  41] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP POST /api/series/delete-multiple responded 200 in 53.9196 ms
[Kavita] [2024-06-06 04:47:49.648 +09:00  50] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP POST /api/series/v2 responded 200 in 21.7324 ms
[Kavita] [2024-06-06 04:48:02.424 +09:00  50] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP GET /api/users/has-library-access?libraryId=4 responded 200 in 2.6566 ms
[Kavita] [2024-06-06 04:48:02.433 +09:00  50] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP GET /api/users/has-library-access?libraryId=4 responded 200 in 2.3780 ms
[Kavita] [2024-06-06 04:48:02.463 +09:00  54] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP GET /api/library/jump-bar?libraryId=4 responded 200 in 1.9103 ms
[Kavita] [2024-06-06 04:48:02.464 +09:00  41] [Information] Serilog.AspNetCore.RequestLoggingMiddleware HTTP GET /api/filter responded 200 in 1.1152 ms
[...]

PR #2988, opened 2024-06-07
Fixing problems with pattern exclusion features in specific folder structures

ytoaa:
This is the action on the following issues.
# Fixed
-  #2982


majora2007:
Thanks, but as mentioned, I have already fixed this in my local branch. I'm just wrapping up 2-3 more bugs/enhancements then will have a new nightly out with this fixed.

PR #3020, opened 2024-06-26
Feature: Bring your own font

Fesaa:
# Added
- Added: user may provide their own fonts to be used in the epub reader

# THIS IS A DRAFT

I'm not done with this yet, quite a bit to do. But feel like it is time to ask for feedback on a few things. Opening a draft pull request, just so it's here as well. Will cross post in the discord after.

**Loading the fonts.** 

They're currently being loaded in the book-reader, this should ensure they're only being loaded when needed. I've also not noticed any higher loading times, but I've only had max 4 fonts loaded. We may want to try with a bunch more, at the very least the default ones, to see if it has any noticeable impact. 

**Design choices**

Need some feedback / greenlight for design choices as to call what what, and place what where. 

- Fonts are called `EpubFont`s
- Fonts are stored in `${pwd}/config/fonts`

If these are fine, how do we want to differentiate between the default fonts, and user fonts? Can we store them in the same folder, should we not? 

If these are not fine, suggestions as to changes to do.

**Link with AppUserPreferences**

Currently, the font preference for a user is saved, as the fonts name. It's still doing this currently, and _should_ work fine. However, we could change this to the ID of the font. Having it be the fonts name, does make it easier to load and reset them. As simple as changing it to default. 

**Default font**
The dropdown menu in the UI, currently, isn't displaying the default font, as it's not an EpubFont. Where should I best inject this option? In the UI, or the controller; instinctively I would do it in the controller, such that the UI doesn't have too much extra logic it has no business of owning. But wanted to ask for opinions anyway, maybe there is a better place? 

### Notes

- The controller currently has the wrong authorize annotations, still has to be changed. 
- The seed class doesn't have all default fonts in it yet. I only copied over one
- I'll probably squash commits before merging and give them a better name

### TODO's
Apart from "smaller" TODO's in the code. I think these are the big ones still to do

- Add update task in Admin UI
- Auto scan fonts
- Scan update communication with UI

I may have missed some


API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+    public FontController(IUnitOfWork unitOfWork, IFontService fontService, ITaskScheduler taskScheduler)
+    {
+        _unitOfWork = unitOfWork;
+        _fontService = fontService;
+        _taskScheduler = taskScheduler;
+    }
+
+    [ResponseCache(CacheProfileName = "10Minute")]
+    [AllowAnonymous]
+    [HttpGet("GetFonts")]
```
majora2007: This is not how API names are. Use either `list` or `all`

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+    public async Task<ActionResult<IEnumerable<EpubFontDto>>> GetFonts()
+    {
+        return Ok(await _unitOfWork.EpubFontRepository.GetFontDtos());
+    }
+
+    [AllowAnonymous]
+    [HttpGet("download-font")]
```
majora2007: I would leave this as HttpGet only. So api is /api/font.

This need authentication via API Key. We cannot have an exposed api to allow downloading fonts off the server.

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+    public async Task<IActionResult> GetFont(int fontId)
+    {
+        try
+        {
+            var font = await _unitOfWork.EpubFontRepository.GetFont(fontId);
+            if (font == null) return NotFound();
+            var contentType = GetContentType(font.FileName);
+            return File(await _fontService.GetContent(fontId), contentType, font.FileName);
```
majora2007: Use PhysicalFile like other areas an use enable Range processing.

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    [AllowAnonymous]
+    [HttpGet("download-font")]
+    public async Task<IActionResult> GetFont(int fontId)
+    {
+        try
+        {
+            var font = await _unitOfWork.EpubFontRepository.GetFont(fontId);
+            if (font == null) return NotFound();
```
majora2007: I like to have a newline above if statements and return statements to help keep code easy to consume.

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+        return Ok(await _unitOfWork.EpubFontRepository.GetFontDtos());
+    }
+
+    [AllowAnonymous]
+    [HttpGet("download-font")]
+    public async Task<IActionResult> GetFont(int fontId)
+    {
+        try
+        {
+            var font = await _unitOfWork.EpubFontRepository.GetFont(fontId);
```
majora2007: Although I don't follow always, having Async when the method returns a task is very helpful.

API/Controllers/FontController.cs
majora2007: Why do we need to pull the font from the DB then later re-pull it at the service layer? Just have the service return the appropriate information and return the file from it. See how I implemented cover image.

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (font == null) return NotFound();
+            var contentType = GetContentType(font.FileName);
+            return File(await _fontService.GetContent(fontId), contentType, font.FileName);
+        }
+        catch (KavitaException ex)
+        {
+            return BadRequest(ex.Message);
```
majora2007: This isn't needed because the Middleware will already catch and throw like this. It's only needed in some cases where localization needs to be done. (as middle ware is unaware of which user triggered an exception)

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    [AllowAnonymous]
```
majora2007: This cannot be AllowAnonymous. I'm not sure why you have all APIs as AllowAnonymous. The only one I can think of as perhaps being needed is GetFont perhaps due to the loading mechanism.

API/Controllers/FontController.cs
```diff
@@ -0,0 +1,72 @@
...
+    [HttpPost("scan")]
+    public IActionResult Scan()
+    {
+        _taskScheduler.ScanEpubFonts();
+        return Ok();
+    }
+
+    private string GetContentType(string fileName)
```
majora2007: First, check if MimeMap does this or not. If it doesn't, we already have an extension or method (see ImageService which uses) that can be supplemented.

API/Services/TaskScheduler.cs
```diff
@@ -431,6 +435,13 @@ public async Task CheckForUpdate()
         await _versionUpdaterService.PushUpdate(update);
     }
 
+    // TODO: Make this auto scan from time to time?
+    public void ScanEpubFonts()
+    {
+        _logger.LogInformation("Starting Epub Font scam");
```
majora2007: Move this into the actual font service.

API/Services/TaskScheduler.cs
```diff
@@ -431,6 +435,13 @@ public async Task CheckForUpdate()
         await _versionUpdaterService.PushUpdate(update);
     }
 
+    // TODO: Make this auto scan from time to time?
```
majora2007: I think it would be better to avoid this and instead allow the user to upload new fonts via some interface, like how I approached the theme rework in v0.8.2.

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public async Task<byte[]> GetContent(int fontId)
+    {
+        // TODO: Differentiate between Provider.User & Provider.System
+        var font = await _unitOfWork.EpubFontRepository.GetFont(fontId);
+        if (font == null) throw new KavitaException("Font file missing or invalid");
+        var fontFile = _directoryService.FileSystem.Path.Join(_directoryService.EpubFontDirectory, font.FileName);
+        if (string.IsNullOrEmpty(fontFile) || !_directoryService.FileSystem.File.Exists(fontFile))
+            throw new KavitaException("Font file missing or invalid");
```
majora2007: We need to throw with a localization key so we can translate to the actual user's language in the controller.

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+        return await _directoryService.FileSystem.File.ReadAllBytesAsync(fontFile);
+    }
+
+    public async Task Scan()
+    {
+        _directoryService.Exists(_directoryService.EpubFontDirectory);
+        var reservedNames = Seed.DefaultFonts.Select(f => f.NormalizedName).ToList();
+        var fontFiles =
+            _directoryService.GetFilesWithExtension(Parser.NormalizePath(_directoryService.EpubFontDirectory), @"\.[woff2|tff|otf|woff]")
```
majora2007: Put the supported font extensions in Parser

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+        // TODO: Differentiate between Provider.User & Provider.System
+        var font = await _unitOfWork.EpubFontRepository.GetFont(fontId);
+        if (font == null) throw new KavitaException("Font file missing or invalid");
+        var fontFile = _directoryService.FileSystem.Path.Join(_directoryService.EpubFontDirectory, font.FileName);
+        if (string.IsNullOrEmpty(fontFile) || !_directoryService.FileSystem.File.Exists(fontFile))
+            throw new KavitaException("Font file missing or invalid");
+        return await _directoryService.FileSystem.File.ReadAllBytesAsync(fontFile);
+    }
+
+    public async Task Scan()
```
majora2007: As you already have a TODO, we should have a SignalR message around Scanning Fonts and use the subtitle to indicate updates like adding a font/removing. 

However, I still think it might be worthwhile to remove the scan part and instead use a font manager system like I did with themes. This way we reduce the amount of filesystem work the user is doing and allow everything to be done via an UI interface.

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        var allFonts = (await _unitOfWork.EpubFontRepository.GetFonts()).ToList();
+        var userFonts = allFonts.Where(f => f.Provider == FontProvider.User).ToList();
+
+        foreach (var userFont in userFonts)
+        {
+            var filePath = Parser.NormalizePath(
+                _directoryService.FileSystem.Path.Join(_directoryService.EpubFontDirectory, userFont.FileName));
+            if (_directoryService.FileSystem.File.Exists(filePath)) continue;
+            allFonts.Remove(userFont);
```
majora2007: You're going to want a space after the if statement to help break it up, since it's a one liner branch if statement.

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+            // TODO: Send update to UI
+            _logger.LogInformation("Removed a font because it didn't exist on disk {FilePath}", filePath);
+        }
+
+        var allFontNames = allFonts.Select(f => f.NormalizedName).ToList();
+        foreach (var fontFile in fontFiles)
+        {
+            var nakedFileName = _directoryService.FileSystem.Path.GetFileNameWithoutExtension(fontFile);
+            // TODO: discuss this, using this to "prettyfy" the file name, to display in the UI
+            var fontName = Regex.Replace(nakedFileName, "[^a-zA-Z0-9]", " ");
```
majora2007: I think this is a good idea, but would like this to be a parser method with unit tests around it.

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+            // TODO: Send update to UI
+            _logger.LogInformation("Added a new font from disk {FontFile}", fontFile);
+        }
+        if (_unitOfWork.HasChanges())
+        {
+            await _unitOfWork.CommitAsync();
+        }
+
+        // TODO: Send update to UI
+        _logger.LogInformation("Finished FontService#Scan");
```
majora2007: Use human friendly information in here. "Font Scan completed"

API/Services/Tasks/FontService.cs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public async Task RemoveFont(EpubFont font)
+    {
+        // TODO: Default font? Ask in kavita discord if needed, as we can always fallback to the browsers default font.
+        _unitOfWork.EpubFontRepository.Remove(font);
```
majora2007, line 224: Don't allow system provided fonts to be removed.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -583,6 +583,15 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngOnInit(): void {
+    this.bookService.getEpubFonts().subscribe(fonts => {
+      fonts.forEach(font => {
+        const fontFace = new FontFace(font.name, `url(${this.bookService.baseUrl}Font/download-font?fontId=${font.id})`);
```
majora2007: This is where you need to use the api key to ensure the user is authenticated.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -583,6 +583,15 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        fontFace.load().then(loadedFace => {
+          (document as any).fonts.add(loadedFace);
```
majora2007: document should come from the injected Document instance, to allow SSR if I ever want to go that route.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -583,6 +583,15 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngOnInit(): void {
+    this.bookService.getEpubFonts().subscribe(fonts => {
+      fonts.forEach(font => {
+        const fontFace = new FontFace(font.name, `url(${this.bookService.baseUrl}Font/download-font?fontId=${font.id})`);
+        fontFace.load().then(loadedFace => {
```
majora2007: One thing I'm worried about is timing. There is no way to know if all the fonts are loaded or not. I also don't see setting the loaded fonts somewhere for the user to select from.

majora2007: Second pass, I think instead what we should do is move this load fonts to the book service. To make this most efficient, we can cache the loaded fonts and return the fonts from that. 

```js
bookService.getEpubFonts() {
 getFonts().pipe(tap(f => {
            const fontFace = new FontFace(font.name, `url(${this.bookService.baseUrl}Font/download-font?fontId=${font.id})`);
            fontFace.load().then(loadedFace => { // cache then put on the document }
            // ...
}
```

We just need to bust this when we get a fontScan complete (or if we go the theme manager route, book theme modified event). 

Thoughts?

UI/Web/src/app/book-reader/_services/book.service.ts
```diff
@@ -4,16 +4,19 @@ import { TextResonse } from 'src/app/_types/text-response';
...
-  /**
-   * The actual font face
-   */
-  family: string;
+export enum FontProvider {
+  System = 1,
+  User = 2,
+}
+
+export interface EpubFont {
```
majora2007: These should be in their own ts file.

UI/Web/src/app/book-reader/_services/book.service.ts
```diff
@@ -25,10 +28,8 @@ export class BookService {
 
   constructor(private http: HttpClient) { }
 
-  getFontFamilies(): Array<FontFamily> {
-    return [{title: 'default', family: 'default'}, {title: 'EBGaramond', family: 'EBGaramond'}, {title: 'Fira Sans', family: 'Fira_Sans'},
-    {title: 'Lato', family: 'Lato'}, {title: 'Libre Baskerville', family: 'Libre_Baskerville'}, {title: 'Merriweather', family: 'Merriweather'},
-    {title: 'Nanum Gothic', family: 'Nanum_Gothic'}, {title: 'RocknRoll One', family: 'RocknRoll_One'}, {title: 'Open Dyslexic', family: 'OpenDyslexic2'}];
+  getEpubFonts(): Observable<EpubFont[]>  {
```
majora2007: We don't need to specify the return on the method signature since it's now obv.

UI/Web/src/app/user-settings/user-preferences/user-preferences.component.ts
```diff
@@ -166,14 +166,18 @@ export class UserPreferencesComponent implements OnInit, OnDestroy {
...
 
     this.accountService.getOpdsUrl().subscribe(res => {
       this.opdsUrl = res;
       this.cdRef.markForCheck();
     });
 
+    this.bookService.getEpubFonts().subscribe(res => {
+      this.fontFamilies = res.map(f => f.name);
+      this.cdRef.markForCheck();
+    })
```
majora2007: missing semicolon.

majora2007 (changes requested):
Overall a great start. I didn't pull it down and test it, but from the code, I got the jist of it. I think this will be a great enhancement, but I do strongly think moving to a more UI centered loading system would be really smart. This will reduce friction by having to have admin's add their own fonts via their filesystem and instead rely completely on the UI. 

I would also argue that we could expand this functionality to non-admins (we would need a role for this). And if we really wanted to go to the next step, we could intake not only a file, but a Google Fonts url and download directly into the fonts directory. 

Let me know your thoughts as that expands the scope (but would be really fun to code).

API/Controllers/FontController.cs
Fesaa: Going through these while working on the UI stuff as well, resolving here so I can keep track better :)

API/Controllers/FontController.cs
Fesaa: I didn't really check, but just using the MimeMap method works so I suppose it does

majora2007:
As mentioned in Discord, I will be taking a look and working on this to help with the implementation and polishing.

majora2007:
Okay here's my plan of attack on this:
1. System Provided fonts (like I did with themes). This will allow the UI to mimic the theme manager. 
2. For the google font idea, we can do it by url. They need to give us the url https://fonts.google.com/specimen/Inter and we then call `https://fonts.google.com/download/list?family=Inter`. This will give us the download information. This will all happen on the backend. Likely we need a Role for this. 
3. Currently you only allow Woff, but not all fonts are in that format. We can expand this further to provide a better experience. 

There were quite a bit of bugs I found that prevented me from actually testing this. I will be working on your code today and pushing to a branch feature/font-manager. We can collaborate after the fact.

majora2007:
First pass of work is implemented, you can review and see my approach and the bugs fixed in feature/font-manager. 

I fixed up how loading takes place for System provided fonts (and fixed seeding as you were checking for fonts in the theme table) and updated the manager to match the UX of Theme manager. Lots of small improvements overall from your code. 

I put a few todos as well in there. I can return to this work later, but this should be enough if you want to start tackling and furthering what I did.

Fesaa:
Hi! I'll have a look tomorrow and try to do some TODO's 🎉

Fesaa:
Going to do this together with more book themes. Sadly, school is taking a lot of my time, so it'll be a while. I'll send a message in discord, when I'm working on it again. Still hoping to get this done at some point!

PR #3025, opened 2024-06-28
Koreader progress sync interface

MFDeAngelo:
# Added
- Added: Added native support to sync progress with Koreader. See [this issue](https://github.com/koreader/koreader/issues/11105) on the Koreader repository for more information.


MFDeAngelo:
This PR is not ready for a merge (if that isn't already apparent.) I wanted to show you the basic interface that I'd like to implement and get some feedback on it before diving in. Here are some relevant details that will help you understand the scope of the proposed change.

As it currently stands, you can read books on Koreader by using OPDS and stream the book from Kavita. Unfortunately, I have two issues with this. First, I haven't even gotten it to work yet. I've only tried a handful of times, but it always produces just the Koreader logo. It might simply be user error, but if you like, I can investigate further and open an issue if necessary. Second, what if I'm away from my LAN and want to sync my books? What if the power goes out? Well too bad. This feature should solve that issue.

Koreader has a plugin that ships with the install called 'progress sync'. It is on by default. It allows a user to connect to a Koreader sync server, to sync progress across Koreader devices. We can use this functionality within Koreader to sync progress between Koreader devices and Kavita. A user would only need to us OPDS to download the book onto their Koreader device, then authenticate to the server in the progress sync page, and poof! The Koreader device and Kavita can sync progress!

The user will have to enter the following url into their Koreader device: 'http://my-kavita-server.com/api/koreader/api-key/'. Then they will login using their username, and the password can be anything, as the authentication happens through the api-key. (I don't think it makes sense to use the user's actual password, as Koreader sends the MD5 hash of the password to the server. Since Kavita doesn't store passwords as an MD5, it would make implementing this much more difficult.) Alternatively we could require the user to use the api-key as the password. Your advice would be appreciated.

Koreader allows you to register as a user. I think it's in our best interest to leave that functionality out.

There are Put and Get endpoints that allow for syncing between the device and the server. Koreader uses the MD5 hash of the book (or filename as a secondary option) to identify what book is being synced. I'm not sure how to handle the device id stuff yet, but that seems like an implementation concern.

Let me know what you think. I'd love to implement this functionality.
Also, I'd like to thank you for putting together this wonderful piece of software. I've really enjoyed using it. It was the second thing I installed on my home server, and I'm glad I did!

majora2007:
First off, I would welcome a PR for this (and any needed PR for OPDS and Koreader). 

> (I don't think it makes sense to use the user's actual password, as Koreader sends the MD5 hash of the password to the server. Since Kavita doesn't store passwords as an MD5, it would make implementing this much more difficult.) Alternatively we could require the user to use the api-key as the password. Your advice would be appreciated.

I agree. I don't want to use passwords with external applications, hence why we have the api key in the first place. 

> Koreader allows you to register as a user. I think it's in our best interest to leave that functionality out.

Agreed. External user provisioning is not something I'm a fan of for Kavita. 


My main points of concern are this: 
1. Make sure you use Dto classes when appropriate with good comments
2. Try to write some unit tests on the actual sync (unless it's as simple as calling SaveProgress())
3. Let's make sure the code is separated into a Koreader controller (which you've already done) and any Koreader service if needed.
4. Please have in mind that documentation will be needed for the wiki as well once this is implemented. 

Let me know if you need anything either via Github or via discord.

MFDeAngelo:
Wonderful! I'll be working on this over the next couple of weeks. I've looked through code, and I think I have a general idea of how to accomplish it. Thanks for your help!

MFDeAngelo:
Not sure why this says it was closed by me. I had no intention of closing it yet.
Just wanted to give an update saying that I'm still working on it. I've been able to establish a connection from koreader to kavita, and now am working on converting the progress datatypes from one to another.

majora2007:
Awesome. I think AutoMapper can help with that as much as possible. It's okay if the PR gets closed, as long as you re-open once you're ready for some more review.

MFDeAngelo:
@majora2007 I've made some more changes and have it working locally. When reading in koreader, the progress is updated on kavita, and vice versa. However, I don't feel confident that I've accounted for every potential "bookScrollId". I wanted to know what you thought about releasing this feature in beta to start off with, with the expectation that it might break here and there.

I'll get this branch caught up on the develop branch in the next couple of days, improve the logging, and start preparing the Wiki entry. Could you perform a quick once over to give me any feedback on it?

majora2007:
Yeah, ill go over it and give you a review. I think we can probably get this into the current nightly, since we still have some good amount of time before the next stable goes out.

API/API.csproj
```diff
@@ -12,9 +12,6 @@
     <LangVersion>latestmajor</LangVersion>
   </PropertyGroup>
 
-  <Target Name="PostBuild" AfterTargets="Build" Condition=" '$(Configuration)' == 'Debug' ">
```
majora2007: You're going to want to commit this back in.

API/Controllers/KoreaderController.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+    }
+    */
+
+    [HttpGet("{apiKey}/users/auth")]
+    public async Task<IActionResult> Authenticate(string apiKey)
+    {
+        var userId = await GetUserId(apiKey);
+        var user = await _unitOfWork.UserRepository.GetUserByIdAsync(userId);
+        return Ok(new { username = user.UserName });
```
majora2007, line 59: Please have newlines above any return statements. 

For the sake of a full review, please use Dto Objects under a DTOs/Koreader folder, so we can easily track and understand the interfaces. 

Also add documentation on the APIs

API/Controllers/KoreaderController.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    [HttpPut("{apiKey}/syncs/progress")]
+    public async Task<IActionResult> UpdateProgress(string apiKey, KoreaderBookDto request)
+    {
+        var userId = await GetUserId(apiKey);
+        await _koreaderService.SaveProgress(request, userId);
+        var response = new
+        {
+            document = request.Document,
+            timestamp = DateTime.Now
```
majora2007: I feel like this should be UtcNow. Otherwise depending on where the server is, Koreader will have different account for timestamp.

API/Controllers/KoreaderController.cs
```diff
@@ -0,0 +1,94 @@
...
+        return Ok(new { username = user.UserName });
+    }
+
+
+    [HttpPut("{apiKey}/syncs/progress")]
+    public async Task<IActionResult> UpdateProgress(string apiKey, KoreaderBookDto request)
+    {
+        var userId = await GetUserId(apiKey);
+        await _koreaderService.SaveProgress(request, userId);
+        var response = new
```
majora2007: You can inline this response

API/Controllers/KoreaderController.cs
```diff
@@ -0,0 +1,94 @@
...
+        };
+        return Ok(response);
+    }
+
+    [HttpGet("{apiKey}/syncs/progress/{ebookHash}")]
+    public async Task<IActionResult> GetProgress(string apiKey, string ebookHash)
+    {
+        var userId = await GetUserId(apiKey);
+        var response = await _koreaderService.GetProgress(ebookHash, userId);
+        Console.WriteLine(response.Progress);
```
majora2007: Use a logger. Might want to put this on Trace so if we need to debug, we can.

API/DTOs/Koreader/EpubPositionDto.cs
```diff
@@ -0,0 +1,8 @@
+
+namespace API.DTOs.Koreader;
+
+public class EpubPositionDto
+{
+    public string KavitaPosition { get; set; }
```
majora2007: I'd love some extra documentation to explain what these fields are. Because once it's merged, I will own it.

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    /// <summary>
+    /// Hashes the document according to a custom Koreader algorithm
+    /// </summary>
+    /// <remarks>The hashing algorithm is relatively quick as it only hashes ~10,000 bytes for the biggest of files.</remarks>
+    /// <param name="filePath">The path to the file to hash</param>
+    public static string HashContents(string filePath)
+    {
+
+        if (string.IsNullOrEmpty(filePath))
```
majora2007, line 24: Move return on same line as if or put {} around

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+            return null;
+
+        using (var file = File.OpenRead(filePath))
```
majora2007: We don't need scoped using aka using (). You can just put using on the same line and let .net do the heavy work for us.

majora2007: We need to use _directoryService.FileSystem.File. This allows us to mock the filesystem.

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                else
+                {
+                    break;
+                }
+            }
+
+            file.Close();
+            md5.TransformFinalBlock(new byte[0], 0, 0);
+            byte[] hashBytes = md5.Hash;
+            return BitConverter.ToString(hashBytes).Replace("-", "").ToUpper();
```
majora2007: Use string.Empty

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+                }
+                else
+                {
+                    break;
+                }
+            }
+
+            file.Close();
+            md5.TransformFinalBlock(new byte[0], 0, 0);
+            byte[] hashBytes = md5.Hash;
```
majora2007: Remove the hashBytes variable, it's not needed.

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        progress.PageNum = Int32.Parse(docNumber) - 1;
+        string lastTag = path[5].ToUpper();
+        if (lastTag == "A")
+        {
+            Console.WriteLine("It's an A tag!");
+            progress.BookScrollId = null;
+        }
+        else
+        {
+            progress.BookScrollId = $"//html[1]/BODY/APP-ROOT[1]/DIV[1]/DIV[1]/DIV[1]/APP-BOOK-READER[1]/DIV[1]/DIV[2]/DIV[1]/DIV[1]/DIV[1]/{lastTag}";
```
majora2007, line 87: Help me understand what this is

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        var fileName = Path.GetFileName(filePath);
+        var fileNameBytes = Encoding.ASCII.GetBytes(fileName);
+        byte[] bytes = md5.ComputeHash(fileNameBytes);
+        return BitConverter.ToString(bytes).Replace("-", "");
+    }
+
+    public static void UpdateProgressDto(string koreaderPosition, ProgressDto progress)
+    {
+        var path = koreaderPosition.Split('/');
+        var docNumber = path[2].Replace("DocFragment[", "").Replace("]", "");
```
majora2007: Please add bounds checking so we don't throw when something isn't as expected.

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+        progress.PageNum = Int32.Parse(docNumber) - 1;
+        string lastTag = path[5].ToUpper();
```
majora2007: Don't use explicit types, use var whenever possible.

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,94 @@
...
+            progress.BookScrollId = $"//html[1]/BODY/APP-ROOT[1]/DIV[1]/DIV[1]/DIV[1]/APP-BOOK-READER[1]/DIV[1]/DIV[2]/DIV[1]/DIV[1]/DIV[1]/{lastTag}";
+        }
+    }
+
+
+    public static string GetKoreaderPosition(ProgressDto progressDto)
+    {
+        string lastTag;
+        int koreaderPageNumber = progressDto.PageNum + 1;
+        if (string.IsNullOrEmpty(progressDto.BookScrollId))
```
majora2007, line 96: if statements always need {} unless a simple jump (return/break/continue) that can be on one line.

API/Services/KoreaderService.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        _readerService = readerService;
+        _unitOfWork = unitOfWork;
+    }
+
+    public async Task SaveProgress(KoreaderBookDto koreaderBookDto, int userId)
+    {
+        var file = await _unitOfWork.MangaFileRepository.GetByHash(koreaderBookDto.Document);
+        var userProgressDto = await _unitOfWork.AppUserProgressRepository.GetUserProgressDtoAsync(file.ChapterId, userId);
+        KoreaderHelper.UpdateProgressDto(koreaderBookDto.Progress, userProgressDto);
```
majora2007, line 37: You need to break up the code into sections so it's easier to understand. The first 2 lines are variable setup. Next 2 are the actual logic. Last is the commit, so break into 3 sections with a newline inbetween.

API/Services/KoreaderService.cs
```diff
@@ -0,0 +1,52 @@
...
+    public async Task<KoreaderBookDto> GetProgress(string bookHash, int userId)
+    {
+        var file = await _unitOfWork.MangaFileRepository.GetByHash(bookHash);
+        var progressDto = await _unitOfWork.AppUserProgressRepository.GetUserProgressDtoAsync(file.ChapterId, userId);
+        var koreaderProgress = KoreaderHelper.GetKoreaderPosition(progressDto);
+        return new KoreaderBookDto
+        {
+            Document = bookHash,
+            Device_id = "kavita",
+            Device = "kavita",
```
majora2007: Kavita should be capital. Is device_id supposed to be a unique identifier? Could we use the ServerId? That is unique unless someone changes their token in appsettings.json (which usually only happens on reinstalling incorrectly).

API/Services/KoreaderService.cs
```diff
@@ -0,0 +1,52 @@
...
+            Progress = koreaderProgress,
+            Percentage = 0.5f
```
majora2007: Curious why we are sending a percentage if it's not correct...

API/Services/Tasks/Scanner/ProcessSeries.cs
```diff
@@ -818,6 +818,7 @@ private void AddOrUpdateFileForChapter(Chapter chapter, ParserInfo info, bool fo
             var file = new MangaFileBuilder(info.FullFilePath, info.Format, _readingItemService.GetNumberOfPages(info.FullFilePath, info.Format))
                 .WithExtension(fileInfo.Extension)
                 .WithBytes(fileInfo.Length)
+                .WithHash()
```
majora2007, line 819: I'd love to know the overhead of hashing files, perhaps a benchmark can help us understand. (See the Benchmark project)

API/Services/Tasks/Scanner/ProcessSeries.cs
```diff
@@ -590,8 +590,8 @@ private void DeterminePublicationStatus(Series series, List<Chapter> chapters)
 
             var nonSpecialVolumes = series.Volumes.Where(v => v.MaxNumber.IsNot(Parser.Parser.SpecialVolumeNumber));
 
-            var maxVolume = (int) (nonSpecialVolumes.Any() ? nonSpecialVolumes.Max(v => v.MaxNumber) : 0);
-            var maxChapter = (int) chapters.Max(c => c.MaxNumber);
+            var maxVolume = (int)(nonSpecialVolumes.Any() ? nonSpecialVolumes.Max(v => v.MaxNumber) : 0);
```
majora2007: Please add the space between the cast back

majora2007 (changes requested):
Lots of style comments and the need for documentation to get the full gist. 

I def would recommend adding some links to places where you learned to reverse engineer, which will not only help me, but any other project that wants to add this sort of support.

API/Data/Migrations/20240704154945_BookHash.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace API.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class BookHash : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "Hash",
```
majora2007: Second thought, let's make this KoreaderHash given they use a custom MD5 hashing algorithm.

API/API.csproj
MFDeAngelo: Oops! Didn't mean to commit this! Adding it back now.

API/Controllers/KoreaderController.cs
MFDeAngelo: I have a DTO in the DTOs/Koreader folder. I'm not sure what other DTO's you'd like me to add there. Are you suggesting that I change the responses from anonymous types to concrete DTOs?

majora2007: Exactly that.

API/DTOs/Koreader/EpubPositionDto.cs
MFDeAngelo: I ended up removing this. It felt weird to pass this object around, as it was often immediately remapped to another.

API/Helpers/KoreaderHelper.cs
MFDeAngelo: This presents an interesting problem. 
KoreaderHelper is currently a static class. In order to add DirectoryService as a member of KoreaderHelper, I have to make it a concrete type. If KoreaderHelper is a concrete type, then it needs to be injected into MangaFileBuilder using dependency injection. MangaFileBuilder cannot be instantiated in the dependency tree (or composition root. Whatever you call it) since it is supposed to be instantiated ad hoc.
I can add a static member to KoreaderHelper and insert a DirectoryService object in, but that feels wrong. Do you have an alternative idea for how to achieve the mocking capability?

majora2007: Hmm, it all comes down to the unit tests. I needed to use a mocked filesystem for a lot of the unit tests to emulate last write times and whatnot, but technically, if we have some non-copywrited files in TestData, we can avoid this whole mess. 

I just want to make sure we have good test coverage on KoreaderHelper.

MFDeAngelo: Understandably. I think the underlying issue is that we are trying to mock out static functionality (reading a file) using a concrete object.

I haven't thought this through all the way, but what if we added the mocking capability to a static object? It obviously has its drawbacks...
```
public static class KavitaFile
{
    private Dictionary<string, Stream> mockedFiles;

    public static Stream OpenRead(string path)
    {
        if (mockedFiles.Any())
        {
            return File.OpenRead(path);
        }
        else
        {
            return mockedFiles[path];
        }
    }

    public static void AddMockedFile(string path, Stream stream)
    {
        mockedFiles.Add(path, stream);
    }
}
```
This is incomplete but gets the idea across. Before any tests run, you need to set up the predefined mock files. It makes sense to me that you'd only want one mocking strategy, so I'm assuming you'd need to convert all places that you're using the directory service.
The most obvious drawback is abusing the mock functionality in application code. I don't think I'd include this in the current PR. Is this a route worth pursuing?

Otherwise, I can grab a few books of Project Gutenberg for test data.

API/Services/KoreaderService.cs
MFDeAngelo: ServerId would make more sense if it is a unique identifier. Could you tell me where to get that from? I can't find anything by searching `ServerId` or its synonyms.

API/Services/KoreaderService.cs
MFDeAngelo: By what I've seen, Koreader doesn't appear to use this value. I think it could be used by the Koreader progress sync server to display statistics or something. I'll see if I can track it down a little better.

API/Services/KoreaderService.cs
majora2007: You should find this on BuildInfo I believe, but maybe getting from the db is better. You can get from setting item. I believe it also goes as InstallId.

API/Helpers/KoreaderHelper.cs
majora2007: I think let's just grab some files. Moving to Manga files adds a lot of mess and really not needed. As long as we can test, I'm happy.

API/Services/KoreaderService.cs
MFDeAngelo: So I've been looking at this even more, and [it appears to be only for message boxes.](
https://github.com/koreader/koreader/blob/master/plugins/kosync.koplugin/main.lua#L798)
A [little earlier in the method](https://github.com/koreader/koreader/blob/master/plugins/kosync.koplugin/main.lua#L764) it compares against a previous percentage and a new percentage to see if if is in the same position. If so, it isn't supposed to update the progress in Koreader. However, in testing it doesn't behave that way. Maybe it has to do with direct comparison of a floating point.

API/Services/Tasks/Scanner/ProcessSeries.cs
MFDeAngelo: I set up the benchmark and will be committing it soon. I set up the benchmark to build a MangaFile using the MangaFileBuilder. I went ahead and ran the benchmark on hashing a 300 KB epub. Here are the results:
```
| Method                  | Mean      | Error     | StdDev    | Median    | Ratio | RatioSD | Rank | Allocated | Alloc Ratio |
|------------------------ |----------:|----------:|----------:|----------:|------:|--------:|-----:|----------:|------------:|
| TestBuildManga_baseline |  40.18 us |  1.631 us |  4.731 us |  38.49 us |  1.00 |    0.00 |    1 |     922 B |        1.00 |
| TestBuildManga_withHash | 230.09 us | 10.654 us | 31.415 us | 229.86 us |  5.81 |    0.99 |    2 |    7151 B |        7.76 |
```

MFDeAngelo: One thing to keep in mind about the hashing algorithm is that a 300KB file will only have 4KB of its contents hashed. At most, the algorithm will hash 10 KB of the file. That means that even though there are files much larger than 300KB, they shouldn't take that much longer to hash.

majora2007: For Koreader, is this only for epub files? Because if so, I think we could swallow this type of overhead. If it's for archives as well, then we might want to look into offloading this into a separate task that performs the hashing. This will offload the extra spend on the scan loop and shortly make it available after scan for syncing.

MFDeAngelo: I switched it to only hash EPUBs. I don't think there is any sense in hashing the other files as I don't believe Koreader supports Archives. It supports PDFs though. We can add support for that later.

MFDeAngelo:
@majora2007, I've merged the most recent changes into my branch and updated the PR. I've made a few small changes since the last time I ran the project. however, I noticed that I can't run the project anymore. I launch the API project and it directs me to localhost, but it returns an HTTP 500.30. I even pulled straight from the Kavita develop branch, and am getting the same result. I'm not sure if it is my local setup, or if there is something wrong in develop. 

Regardless, I'd like to run a few quick tests, then I'm ready for an in depth review. If we can resolve the HTTP 500 error, then I'll run those tests and get it to you to review.

majora2007:
I just pulled down your code and the webapp is working just fine, I'm not able to reproduce any 500 errors. Are you also launching the webui server as well with angular? You should be going to localhost:4200/

API/Entities/MangaFile.cs
```diff
@@ -21,6 +21,10 @@ public class MangaFile : IEntityDate
     /// </summary>
     public required string FilePath { get; set; }
     /// <summary>
+    /// A hash of the document using Koreader's unique hashing algorithm
+    /// </summary>
+    public string KoreaderHash { get; set; }
```
majora2007: This should be `string?` as we aren't using for all file types. I would also update `<remarks>` to include information about the supported file types.

API/Helpers/Builders/MangaFileBuilder.cs
```diff
@@ -60,4 +60,13 @@ public MangaFileBuilder WithId(int id)
...
         return this;
     }
+
+    public MangaFileBuilder WithHash()
+    {
+        if (_mangaFile.Format == MangaFormat.Epub)
+        {
+            _mangaFile.KoreaderHash = KoreaderHelper.HashContents(_mangaFile.FilePath);
+        }
+        return this;
```
majora2007, line 71: Newlines above returns

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <summary>
+    /// Hashes the document according to a custom Koreader hashing algorithm.
+    /// Look at the util.partialMD5 method in the attached link.
+    /// </summary>
+    /// <remarks>The hashing algorithm is relatively quick as it only hashes ~10,000 bytes for the biggest of files.</remarks>
+    /// <see href="https://github.com/koreader/koreader/blob/master/frontend/util.lua#L1040"/>
+    /// <param name="filePath">The path to the file to hash</param>
+    public static string HashContents(string filePath)
+    {
+
```
majora2007: Remove empty line

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+        if (string.IsNullOrEmpty(filePath))
+        {
+            return null;
+        }
+
+        using var file = File.OpenRead(filePath);
+
+        int step = 1024;
```
majora2007: Use var whenever possible

API/Helpers/KoreaderHelper.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// <summary>
+    /// Koreader can identitfy documents based on contents or title.
+    /// For now we only support by contents.
+    /// </summary>
+    public static string HashTitle(string filePath)
+    {
+        var fileName = Path.GetFileName(filePath);
+        var fileNameBytes = Encoding.ASCII.GetBytes(fileName);
+        var bytes = MD5.HashData(fileNameBytes);
+        return BitConverter.ToString(bytes).Replace("-", string.Empty);
```
majora2007, line 67: Newline above returns

API/Services/KoreaderService.cs
majora2007: It looks like it's a user friendly percentage. Can we not use the Progress mapped to a 0-1 float?

majora2007 (commented):
I did a code review, I will pull this down soon and start testing the heck out of it. 

@DieselTech if you want to also start testing. 

We missed our opportunity for the v0.8.3 release, but v0.8.4 development is right around the corner and I know many people will want this feature.

MFDeAngelo:
@majora2007 I fixed the stylistic concerns you had. I'm not sure how exactly to map a progress to a float though. It seems like it could be difficult, because of the way Kavita tracks a user's progress. I could use currentPage/totalPages, but that isn't the same way that Koreader tracks progress, and would lead to discrepancies. Do you have any other ideas?

majora2007:
> @majora2007 I fixed the stylistic concerns you had. I'm not sure how exactly to map a progress to a float though. It seems like it could be difficult, because of the way Kavita tracks a user's progress. I could use currentPage/totalPages, but that isn't the same way that Koreader tracks progress, and would lead to discrepancies. Do you have any other ideas?

Well Kavita does technically use currentPage/totalPage. It just stores an additional xpath to resume scroll position, but the underlying progress would be currentPage/totalPage. This would ensure that the modal on Koreader side shows something correct. 

I'll try and pull this down this weekend and give it a round of review. I don't have Koreader, but we can deploy this to a test instance and get some testers before we merge it into develop.

majora2007:
@DieselTech Can we get this deployed to the dev instance so we can ask users to test it? I don't want to merge into nightly yet without more validation testing from people with actual devices. We can lock the users down to dedicated testers.

majora2007:
Just to update you, I'm planning to put this in a canary and asking for some testing before I merge it into develop. Canary is currently being used for a large refactor of the scan loop, to ensure it's not breaking anything before moving the code into nightly/develop. Once that wraps, this is up next.

calilex:
This feature is great, I’m looking forward to it very much.

API/Services/KoreaderService.cs
majora2007: Someone from Stump did some research on this:
https://github.com/stumpapp/stump/issues/239#issuecomment-2428256328

majora2007:
Just an update, canary should be coming to an end today/tomorrow. It's been nearly a week of just working on rewriting the scanner but we are at a good merge point. Once that occurs, I will be creating a new canary with this for testing (which should be quick, I already have a user to perform the extended testing). 

Thank you for your patience. With bigger PRs, it's critical to ensure things work well before merging in and taking ownership.

majora2007:
If you could resolve the conflicts then I'll get you merged into canary and start the testing. 

Discussion for testing: 
https://discord.com/channels/821879810934439936/1298773623825367170

MFDeAngelo:
Hi @majora2007, apologies for the delay. I've never worked with synchronizing a fork. I finally understand what having multiple remotes is for now, so I think I'll be a little better at merging these things.

Anyways, I've brought the branch up to date with the feature/koreader-sync branch.

API/Entities/MangaFile.cs
```diff
@@ -21,6 +21,11 @@ public class MangaFile : IEntityDate
     /// </summary>
     public required string FilePath { get; set; }
     /// <summary>
+    /// A hash of the document using Koreader's unique hashing algorithm
+    /// <remark> KoreaderHash is only available for epub types </remark>
```
majora2007, line 25: Reminder to self, I need to fix this (remarks should be outside of summary tag)

API/Helpers/Builders/MangaFileBuilder.cs
```diff
@@ -60,4 +60,14 @@ public MangaFileBuilder WithId(int id)
         _mangaFile.Id = Math.Max(id, 0);
         return this;
     }
+
+    public MangaFileBuilder WithHash()
+    {
+        if (_mangaFile.Format == MangaFormat.Epub)
```
majora2007, line 66: Reminder to self, inversion of control this (aka check if not epub and return this)

majora2007 (approved):

majora2007:
I'm going to do some cleanup for you and once done, I encourage you to read over the upcoming PR into canary. 

Some items I need from you:
* Descriptions for the KoreaderController methods
* Acknowledgement the epub included is not under copywrite

majora2007:
@MFDeAngelo A few more things that I'll need implementation help on. In the KoreaderService, there are a lot of cases of data coming back null that aren't being handled. I'm going to augment the code with TODOs and push it up. Please branch off my branch and get everything in order. 

You wont have to deal with any merges anymore since the code is in my own branch. I just need help since I don't have Koreader setup. 

You can see the TODOs on the PR:
https://github.com/Kareadita/Kavita/pull/3311

john-okeefe:
I'm very happy to see this feature being worked on. I do hope it can be implemented soon.

PR #3046, opened 2024-07-08
Login/Registration page updates

therobbiedavis:
# Added
- Added: Added Login Background css vars.

# Changed
- Changed: Changed max characters on password field to 256.

# Fixed
- Fixed: Fixed an issue where long passwords would truncate on autofill. This was due to the max characters.


majora2007:
The backend doesn't support more than 32 right? I think it's in Identity service.

PR #3073, opened 2024-07-23
Fixes highlight issues in side nav

Fesaa:
# Fixed
- Fixed: All custom filters being highlighted when a specific one is selected
- Fixed: "All Series" being highlighted when opening the web app on a custom filter


Fesaa:
Dev notes; 
~~I've just removed the comparisonMethod as I felt like it was making the code a lot more complicated than need be. 
I did my best to look at all usages and check they're working correctly. But I might have missed something(?), although just looking at the code. It should be fine, I believe. Happy to adjust if I looked over something dumb~~ (I was infact dumb)

The changes now just fix the behavior of the equals comparisonMethod. If you look at the previous code, equals never checked queryParams, and the queryParams check for startsWith does nothing.

I also looked for an issue mentioning this, but oddly enough, I couldn't find anything. Although I suppose I also noticed it only last weekend :woman_shrugging: 

Before:

https://github.com/user-attachments/assets/335db368-7cc8-422a-92d0-dc62c468ad77

After:

https://github.com/user-attachments/assets/5012d457-3910-4665-a93e-8e26762adb02

majora2007:
I'll take a look when I'm back from holiday but I made some changes already in feature/ux-overhaul (it's Robbie's PR) already. Maybe you could look at that (it has a db migration, might want to backup db or patch changes) and see if there is a modification on that you can do.

Fesaa:
Looking at the code in that branch, the same issues would still be present there. 

I can merge that branch into mine and update the changes to work with the ones already present there. And then change the target branch for the pr; If that's what you meant (if it's present after double-checking)

majora2007:
I see that you're saying this still exists. Can you provide me some reproduction steps to generate the issue?

Fesaa:
Not sure what happened to my WebStorm, but it decided to redo some syntax. Sorry about that...

Yes, this is still an issue. Easiest is to just add two smart filters to your nav bar. And they'll both light up if you select one

majora2007:
I manually ported your code over and it doesn't work

image

I looked at the bug and it seems that I'm not looking at query params for comparison, which is how those smart filters differ. I'll take over driving this home. Thanks for the initial PR.

Fesaa:
Rather confused why the code doesn't work 😔, as it does for me, but happy to see this being fixed with the UX reworks 🎉

majora2007:
This was the way I fixed it:
```
const queryParamsEqual = this.queryParams === queryParams;
    

    if (this.comparisonMethod === 'equals' && page === this.link && fragmentEqual && queryParamsEqual) {
      this.highlighted = true;
      this.cdRef.markForCheck();
      return;
    }
```

PR #3094, opened 2024-08-06
 Added Button in "Statistics" Tab to Toggle Compact Number Display On and Off

RPLgrime:
# Added
- Added: New button on 'Statistics' tab below the server stats display that allows user to toggle CompactNumbers on and off.


RPLgrime:
I have a large library and Compact Numbers truncates quite a bit for me, so I thought it would be nice to give users a quick toggle to see their data in both ways. I preserved the Kavita intent by defaulting to Compact Numbers = true.
Below is an example. If I was using my actual library with more files, the numbers would also update on 'Series', 'Volumes', 'Genres', 'Tags' as well. 

image
image

majora2007:
If you hover over the compact number, it will tell you the exact amount. I don't see a need to have a button to disable. Plus this button looks out of place and doesn't have good UX.

PR #3108, opened 2024-08-12
PDF Metadata Support

microtherion:
# Added
- Added: Added the ability to read embedded metadata from PDFs that conform to Calibre's embedding. (Closes FR #3103)


API/Services/BookService.cs
```diff
@@ -425,10 +426,189 @@ private async Task InlineStyles(HtmlDocument doc, EpubBookRef book, string apiBa
...
+    private ComicInfo? GetPdfComicInfo(string filePath)
     {
-        if (!IsValidFile(filePath) || Parser.IsPdf(filePath)) return null;
+        try
+        {
+            const int chunkSize = 4096;
+            const int overlap = 16;
+            var stream = File.OpenRead(filePath);
+            byte[] buffer = new byte[chunkSize + overlap];
+            int bytesAvailable = 0;
```
majora2007: Try to use var whenever possible.

API/Services/BookService.cs
```diff
@@ -425,10 +426,189 @@ private async Task InlineStyles(HtmlDocument doc, EpubBookRef book, string apiBa
...
+
+        return info;
+    }
+
+    private ComicInfo? GetPdfComicInfo(string filePath)
     {
-        if (!IsValidFile(filePath) || Parser.IsPdf(filePath)) return null;
+        try
+        {
+            const int chunkSize = 4096;
```
majora2007: This is a very technical implementation. Is it not possible to get the comment field from docnet then parse out the xml from that?

API/Services/BookService.cs
```diff
@@ -425,10 +426,189 @@ private async Task InlineStyles(HtmlDocument doc, EpubBookRef book, string apiBa
...
+            }
+            if (found)
+            {
+                return pos;
+            }
+        }
+        return -1;
+    }
+
+    private string? GetTextFromXmlNode(XmlDocument doc, XmlNamespaceManager ns, string path)
```
majora2007: I would move all this PDF stuff into a PDF helper class. Because there are so many methods that it's polluting the book service.

API/Services/BookService.cs
```diff
@@ -425,10 +426,189 @@ private async Task InlineStyles(HtmlDocument doc, EpubBookRef book, string apiBa
...
+        var info = new ComicInfo();
+        var publicationDate = GetDateTimeFromXmlNode(metaDoc, ns, "//dc:date")
+            ?? GetDateTimeFromXmlNode(metaDoc, ns, "//xmp:createdate");
+        if (publicationDate != null)
+        {
+            info.Year  = publicationDate.Value.Year;
+            info.Month = publicationDate.Value.Month;
+            info.Day   = publicationDate.Value.Day;
+        }
+        info.Summary   = GetTextFromXmlNode(metaDoc, ns, "//dc:description") ?? "";
```
majora2007: Use string.Empty please

majora2007: Newline after closing bracket

API/Services/BookService.cs
```diff
@@ -425,10 +426,189 @@ private async Task InlineStyles(HtmlDocument doc, EpubBookRef book, string apiBa
...
+        info.Publisher = GetTextFromXmlNode(metaDoc, ns, "//dc:publisher") ?? "";
+        info.Writer    = GetListFromXmlNode(metaDoc, ns, "//dc:creator") ?? "";
+        info.Title     = GetTextFromXmlNode(metaDoc, ns, "//dc:title") ?? "";
+        info.Genre     = GetListFromXmlNode(metaDoc, ns, "//dc:subject") ?? "";
+        info.LanguageISO = ValidateLanguage(GetTextFromXmlNode(metaDoc, ns, "//dc:language"));
+        ComicInfo.CleanComicInfo(info);
```
majora2007: Clean should be last

majora2007 (commented):
Preliminary look. I'm mainly concerned with the implementation to fetch the data from the pdf. I would like to investigate grabbing the string via docnet, which we already use for other things in Kavita.

API/Services/BookService.cs
microtherion: I was following the structure of the corresponding epub implementation. Should clean in that one also be moved?

API/Services/BookService.cs
microtherion: As far as I could tell from the sparse documentation, Docnet is rather page oriented, and the metadata does not seem to be attached to any particular page. And I don't see APIs to access non-visible parts of the PDF file anyway.

It's not clear to me whether PDFium has support for this metadata (documentation is downright nonexistent). But I see neither support for returning the entire XML blob as a whole, nor specific support for calibre specific info.

So I don't see how to make this approach work.

majora2007: You're right, I just checked. I had done some experiment in the past, so I had assumed one of those could expose the comment field. If we are going to move forward then with a custom implementation, we need to wrap it up into a PdfService (and move some of the pdf code from bookservice to proxy into it) so it's more contained.

microtherion:
Implemented review feedback. Was not sure about the placement of `PdfMetadataExtractor`. Is `API.Helpers` OK?

API/Services/BookService.cs
majora2007: That's a great question. I would advice to leave it for now to avoid having to do regression testing from moving it.

API.Tests/Services/BookServiceTests.cs
```diff
@@ -81,4 +81,14 @@ public void ShouldParseAsVolumeGroup_WithSeriesIndex()
         Assert.Equal("Accel World", comicInfo.Series);
     }
 
+    [Fact]
```
majora2007, line 84: Can you add a unit test for when there is no metadata or a failure, to ensure we fail correctly.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+    public PdfMetadataExtractor(ILogger<BookService> logger, IMediaErrorService mediaErrorService)
+    {
+        _logger = logger;
+        _mediaErrorService = mediaErrorService;
+    }
+
+    private int FindInBuffer(byte[] buffer, int bufLen, byte[] match)
+    {
+        var maxPos = bufLen - match.Length;
+        for (var pos = 0; pos<=maxPos; ++pos)
```
majora2007: pos <= maxPos

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            var found = true;
+            for (var ch = 0; ch<match.Length; ++ch)
```
majora2007: ch < match.Length

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        return float.Parse(text);
+    }
+
+    private string? GetListFromXmlNode(XmlDocument doc, XmlNamespaceManager ns, string path)
+    {
+        var nodes = doc.DocumentElement?.SelectNodes(path+"//rdf:li", ns);
+        if (nodes == null) return null;
+        var list = new StringBuilder();
+        foreach (XmlNode n in nodes)
```
majora2007: Newlines above for loops please.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+        if (string.IsNullOrEmpty(text)) return null;
+
+        return float.Parse(text);
+    }
+
+    private string? GetListFromXmlNode(XmlDocument doc, XmlNamespaceManager ns, string path)
+    {
+        var nodes = doc.DocumentElement?.SelectNodes(path+"//rdf:li", ns);
+        if (nodes == null) return null;
+        var list = new StringBuilder();
```
majora2007: Newline after an if statement, especially if it's performing a jump

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+        ns.AddNamespace("calibreSI", "http://calibre-ebook.com/xmp-namespace-series-index");
+        ns.AddNamespace("calibre", "http://calibre-ebook.com/xmp-namespace");
+        ns.AddNamespace("pdfx", "http://ns.adobe.com/pdfx/1.3/");
+        ns.AddNamespace("prism", "http://prismstandard.org/namespaces/basic/2.0/");
+        ns.AddNamespace("xmp", "http://ns.adobe.com/xap/1.0/");
+
+        var info = new ComicInfo();
+        var publicationDate = GetDateTimeFromXmlNode(metaDoc, ns, "//dc:date")
+            ?? GetDateTimeFromXmlNode(metaDoc, ns, "//xmp:createdate");
+        if (publicationDate != null)
```
majora2007: Newline above if statement please

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+        info.Publisher = GetTextFromXmlNode(metaDoc, ns, "//dc:publisher") ?? String.Empty;
+        info.Writer    = GetListFromXmlNode(metaDoc, ns, "//dc:creator") ?? String.Empty;
+        info.Title     = GetTextFromXmlNode(metaDoc, ns, "//dc:title") ?? String.Empty;
+        info.Genre     = GetListFromXmlNode(metaDoc, ns, "//dc:subject") ?? String.Empty;
+        info.LanguageISO = BookService.ValidateLanguage(GetTextFromXmlNode(metaDoc, ns, "//dc:language"));
+
+        info.Isbn = GetTextFromXmlNode(metaDoc, ns, "//pdfx:isbn") ?? GetTextFromXmlNode(metaDoc, ns, "//prism:isbn") ?? String.Empty;
+        if (!ArticleNumberHelper.IsValidIsbn10(info.Isbn) && !ArticleNumberHelper.IsValidIsbn13(info.Isbn))
+        {
+            _logger.LogDebug("[BookService] {File} has invalid ISBN number", filePath);
```
majora2007: [BookService] {File} has **an** invalid ISBN number

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+        try
+        {
+            const int chunkSize = 4096;
+            const int overlap = 16;
+            var stream = File.OpenRead(filePath);
+            var buffer = new byte[chunkSize + overlap];
+            var bytesAvailable = 0;
+            var hasMetaData = false;
+            var meta = new byte[0];
+            while (!hasMetaData)
```
majora2007: Newline before while

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                var bytesRead = stream.Read(buffer, bytesAvailable, chunkSize);
+                if (bytesRead == 0) break;
```
majora2007: Newline after if statements that perform any control flow (break/continue/return)

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+                bytesAvailable += bytesRead;
+                var found = FindInBuffer(buffer, bytesAvailable, Encoding.UTF8.GetBytes("<x:xmpmeta"));
+                if (found >= 0)
+                {
+                    meta = buffer[found..bytesAvailable];
+                    hasMetaData = true;
+                    break;
+                }
+                else
```
majora2007: Remove else statement, it's not needed as the if statement performs a jump

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+            while (!hasMetaData)
+            {
+                var bytesRead = stream.Read(buffer, bytesAvailable, chunkSize);
+                if (bytesRead == 0) break;
+                bytesAvailable += bytesRead;
+                var found = FindInBuffer(buffer, bytesAvailable, Encoding.UTF8.GetBytes("<x:xmpmeta"));
+                if (found >= 0)
+                {
+                    meta = buffer[found..bytesAvailable];
+                    hasMetaData = true;
```
majora2007: hasMetadata should be the casing of the variable.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+            while ((bytesAvailable = stream.Read(buffer, 0, chunkSize)) > 0 || hasMetaData)
+            {
+                hasMetaData = false;
+                if (bytesAvailable > 0) {
+                    byte[] newMeta = new byte[meta.Length + bytesAvailable];
+                    Buffer.BlockCopy(meta, 0, newMeta, 0, meta.Length);
+                    Buffer.BlockCopy(buffer, 0, newMeta, meta.Length, bytesAvailable);
+                    meta = newMeta;
+                }
+                var found = FindInBuffer(meta, meta.Length, Encoding.UTF8.GetBytes("</x:xmpmeta>"));
```
majora2007: Newline above found

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+                if (found >= 0)
+                {
+                    var metadata = meta[0..(found + "</x:xmpmeta>".Length)];
```
majora2007: I would bring the endLength into a variable to cleanup the reading of the 0..N sytnax.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,213 @@
...
+                    return GetComicInfoFromMetadata(Encoding.UTF8.GetString(metadata), filePath);
+                }
+            }
+            return null; // Unterminated metadata
```
majora2007: Undetermined

majora2007 (changes requested):
Overall code looks good, minus a few style points. Once the code is in shape, I can spare some time to pull it down and ensure it doesn't break anything.

majora2007:
Another thing that would be stellar is if you could update the wiki along with this to write the mapping of fields like we do with the epub reader and perhaps any settings needed on our calibre guide for ensuring metadata is written in the pdf.

API/Helpers/PdfMetadataExtractor.cs
microtherion: That was copied from the Epub code. Should I change that one as well?

API/Helpers/PdfMetadataExtractor.cs
microtherion: No, I did mean "unterminated", as in missing the closing `</x:xmpmeta>` tag: https://www.thefreedictionary.com/unterminated

To me, "undetermined" means something different (unsettled, undecided).

microtherion: But I could simply spell out that the closing tag was missing, so we don't have to settle for either word.

microtherion:
Addressed second round of comments, and added documentation in https://github.com/Kareadita/Wiki-Nextra/pull/13

API/Helpers/PdfMetadataExtractor.cs
majora2007: Yes please.

DieselTech:
Did some initial testing of the PR and came across 2 aspects that need to be looked into:

1 - File locking - Once the scanner hits the files it doesn't seem to let go of the file lock. I couldn't move files around or modify them on the host file system after kavita scanned them. 

https://cdn.discordapp.com/attachments/1273694551739072532/1273694566226333736/image.png?ex=66bf8c00&is=66be3a80&hm=3ef14bddff709d1a7e9f748db3066d070e509ee8419c48240b2ffe25cc9e9a4a&

2 - Long scan times per file. Even on my fairly powerful desktop, scan times were taking 5-6 seconds per file scanned in. It took almost an hour to bring in the 749 files I have in a test library. This is only going to get worse on lower powered hardware. Especially if it has to load the entire file to search for metadata. People TTRPG collections are going to have 300MB+ files and hundreds of them, if not thousands. That can easily start to add up to weeks of scanning time.

microtherion:
@DieselTech do you see the same behavior re: (1) for epub and pdf?

For (2) were those 749 files pdf? If so, do you have comparison numbers for epub, so I know what I ought to aim for? It might inherently be a difficult problem, as there may not be a way around reading the whole file, but there may be an exploitable file structure in PDF that could be used.

DieselTech:
2 was all PDF files. I didn't test any epub with this branch but I can just to get some comparison

microtherion:
It turns out that PDF files indeed have an index that allows much faster access to the metadata than the implementation here, though accessing it is a bit trickier.

I'm still working on the new approach.

microtherion:
All right, here's an almost completely rewritten version. Instead of doing a brute force scan, it properly parses the PDF metadata. In my testing, even an 1.3G PDF was handled in under 600ms. It also handles more sources of truth, i.e. the Info directory, if no XML metadata is provided.

I tested with numerous PDFs I found lying around my hard drive. Two cases that the code does not handle are (1) encrypted files and (2) files using nonstandard conditioning for their Deflate compression. Case (1) is not all that uncommon in the wild, but probably not for the Kavita use case. (2) seems rare and weird, I've only seen it in one file in the Ghostscript distribution. In both cases, the PDF extraction engine will throw an exception, the higher level code will return `null`, and Kavita will use the previous, file name based, approach.

Added two more test files (both generated myself, so there should be no copyright issues), one to demonstrate that PDFs generated by MacOS (which use lots of indirect objects) work, the other to demonstrate that encrypted PDFs fail gracefully.

One remaining concern is that Visual Studio Code told me that throwing system exceptions is not proper to do. Is there some other exception class I should be using? None of the exceptions escape the subsystem, they are all caught and translated into a `null` return.

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+        return float.Parse(text);
+    }
+
+    private DateTime? GetDateTimeFromText(string? text)
+    {
+        if (string.IsNullOrEmpty(text)) return null;
+
+        DateTime date;
+
+        if (DateTime.TryParse(text, out date))
```
majora2007: We can refactor this out to be

if (DateTime.TryParse(text, out var date)) return date;

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        // Normalize possible PDF date
+        if (text[0] != 'D') {
+            text = "D:" + text;
+        }
+        text = text.Replace("'", ":");
+        text = text.Replace("Z", "+");
+
+        string[] pdfDateFormats = [
```
majora2007: Given these are fixed, let's put them in a const at the top of PdfComicInfoExtractor

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+            "D:yyyyMMddHHmmsszzz:", "D:yyyyMMddHHmmss+", "D:yyyyMMddHHmmss",
+            "D:yyyyMMddHHmmzzz:",  "D:yyyyMMddHHmm+",   "D:yyyyMMddHHmm",
+            "D:yyyyMMddHHzzz:", "D:yyyyMMddHH+", "D:yyyyMMddHH",
+            "D:yyyyMMdd", "D:yyyyMM", "D:yyyy"
+        ];
+
+        foreach(var format in pdfDateFormats)
+        {
+            if (DateTime.TryParseExact(text, format, null, System.Globalization.DateTimeStyles.None, out date))
```
majora2007: Since you will now have a conflict given my top comment, rename this to out var parsedDate

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                return date;
+            }
+        }
+
+        return null;
+    }
+
+    private string? MaybeGetMetadata(Dictionary<String, String> metadata, string key)
```
majora2007: Please use lowercase string throughout the file.

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+        info.Isbn      = MaybeGetMetadata(metadata, "ISBN") ?? String.Empty;
+
+        if (info.Isbn != String.Empty && !ArticleNumberHelper.IsValidIsbn10(info.Isbn) && !ArticleNumberHelper.IsValidIsbn13(info.Isbn))
+        {
+            _logger.LogDebug("[BookService] {File} has an invalid ISBN number", filePath);
+            info.Isbn = String.Empty;
+        }
+
+        info.UserRating = GetFloatFromText(MaybeGetMetadata(metadata, "UserRating")) ?? 0.0f;
+        info.TitleSort  = MaybeGetMetadata(metadata, "TitleSort") ?? String.Empty;
```
majora2007: string.Empty please

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+        info.Publisher = MaybeGetMetadata(metadata, "Publisher") ?? String.Empty;
+        info.Writer    = MaybeGetMetadata(metadata, "Author") ?? String.Empty;
+        info.Title     = MaybeGetMetadata(metadata, "Title") ?? String.Empty;
+        info.Genre     = MaybeGetMetadata(metadata, "Subject") ?? String.Empty;
+        info.LanguageISO = BookService.ValidateLanguage(MaybeGetMetadata(metadata, "Language"));
+        info.Isbn      = MaybeGetMetadata(metadata, "ISBN") ?? String.Empty;
+
+        if (info.Isbn != String.Empty && !ArticleNumberHelper.IsValidIsbn10(info.Isbn) && !ArticleNumberHelper.IsValidIsbn13(info.Isbn))
+        {
+            _logger.LogDebug("[BookService] {File} has an invalid ISBN number", filePath);
```
majora2007, line 109: Let's put this on Warning.

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+            info.Isbn = String.Empty;
+        }
+
+        info.UserRating = GetFloatFromText(MaybeGetMetadata(metadata, "UserRating")) ?? 0.0f;
+        info.TitleSort  = MaybeGetMetadata(metadata, "TitleSort") ?? String.Empty;
+        info.Series     = MaybeGetMetadata(metadata, "Series") ?? String.Empty;
+        info.SeriesSort = info.Series;
+        info.Volume     = Convert.ToInt32(GetFloatFromText(MaybeGetMetadata(metadata, "Volume")) ?? 0.0f).ToString();
```
majora2007: Could we not use int.TryParse() ?? 0.0f

API/Helpers/PdfComicInfoExtractor.cs
```diff
@@ -0,0 +1,158 @@
...
+        // If this is a single book and not a collection, set publication status to Completed
+        if (string.IsNullOrEmpty(info.Volume) && Parser.ParseVolume(filePath, LibraryType.Manga).Equals(Parser.LooseLeafVolume))
+        {
+            info.Count = 1;
+        }
+
+        var hasVolumeInSeries = !Parser.ParseVolume(info.Title, LibraryType.Manga)
+            .Equals(Parser.LooseLeafVolume);
+
+        if (string.IsNullOrEmpty(info.Volume) && hasVolumeInSeries && (!info.Series.Equals(info.Title) || string.IsNullOrEmpty(info.Series)))
```
majora2007: I'm wondering if we need this in here. Pdf aren't too common for Light Novels, so potentially we should comment this out with a // Is this needed?

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+namespace API.Helpers;
+#nullable enable
+
+//
+// Parse PDF file and try to extract as much metadata as possible.
+// Supports both text based XRef tables and compressed XRef streams (Deflate only).
+// Supports both UTF-16 and PDFDocEncoding for strings.
+// Lacks support for many PDF configurations that are theoretically possible, but should handle most common cases.
+//
+public interface IPdfMetadataExtractor
```
majora2007, line 45: Please use /// for the summary tag.

majora2007, line 45: Please add a comment that you are the original creator please. However you like, simplest would be to link to your Github.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+        _builder.Append(c);
+    }
+
+    public void AppendByte(byte b)
+    {
+        if (_builder.Length == 0 && !_isUnicode)
+        {
+            if (_secondByte)
+            {
+                if (b == 0xFF)
```
majora2007, line 100: What is 0xFF/0xFE? Is there any sort of documentation you can link this code with? Or can we rename these magic numbers to variables to add some context?

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+
+        _logger.LogInformation("PDF {File}, object offsets {ObjOffsetTime} ms, metadata {MetadataTime} ms", filename, objOffsetTime, metadataTime);
+    }
+
+    public Dictionary<string, string> GetMetadata()
+    {
+        return _metadata;
+    }
+
+    private readonly ILogger<BookService> _logger;
```
majora2007, line 801: Please move these type members to the top of the class above the constructor.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+        ReadObjectOffsets();
+        long objOffsetTime = timing.ElapsedMilliseconds;
+
+        ReadMetadata(filename);
+        long metadataTime = timing.ElapsedMilliseconds - objOffsetTime;
+
+        LogMetadata(filename);
+        timing.Stop();
+
+        _logger.LogInformation("PDF {File}, object offsets {ObjOffsetTime} ms, metadata {MetadataTime} ms", filename, objOffsetTime, metadataTime);
```
majora2007: Let's remove this and the timer.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+            this.root = root;
+            this.info = info;
+        }
+    }
+
+    private readonly Stack<MetadataRef> metadataRef = new();
+
+    private void LogMetadata(string filename)
+    {
+        _logger.LogDebug("Metadata for {Path}:", filename);
```
majora2007: Lets probably comment these out or move to Trace logging

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+            ReadXRefStream();
+
+            return;
+        }
+
+        var token = _lexer.NextToken();
+
+        if (token.type != PdfLexer.TokenType.Keyword || (string)token.value != "xref")
+        {
+            throw new Exception("Expected xref keyword");
```
majora2007: Since you asked about this, we could use a PdfMetadataExtractorException or something to capture everything coming out of here.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+            else
+            {
+                throw new Exception("Unexpected token in xref");
+            }
+        }
+
+        ReadTrailerDictionary();
+    }
+
+    private struct XRefSection
```
majora2007: Move to the top or bottom, don't have members mixed in the middle of the class.

API/Helpers/PdfMetadataExtractor.cs
```diff
@@ -0,0 +1,1605 @@
...
+
+    private void ReadMetadata(string filename)
+    {
+        // We read potential metadata sources in backwards historical order, so
+        // we can overwrite to our hearts content
+        while (metadataRef.Count > 0)
+        {
+            var meta = metadataRef.Pop();
+
+            _logger.LogDebug("DocumentCatalog for {Path}: {Root}, Info: {Info}", filename, meta.root, meta.info);
```
majora2007: Make all your Debugs into Trace. Debug is normal for users to be on with Kavita.

majora2007 (changes requested):
Overall looks pretty good. A few comments for you. I'm doing the regression and validation testing as well.

API/Helpers/PdfMetadataExtractor.cs
microtherion: This is a big endian UTF-16 BOM marker. I'll add a comment to that effect.

majora2007:
From testing I found 2 code changes that needs to be handled. This will make it actually viable and tweak it based on testing. 

In `PdfParser.cs`, remove this block of code
```cs
// Pdfs may have .pdf in the series name, remove that
        if (Parser.IsPdf(filePath) && ret.Series.ToLower().EndsWith(".pdf"))
        {
            ret.Series = ret.Series.Substring(0, ret.Series.Length - ".pdf".Length);
        }
```
        
In `PdfComicInfoExtractor.cs`, make this change:
```cs
info.Series     = MaybeGetMetadata(metadata, "Series") ?? info.TitleSort; // Calibre title is more likely set thant Series
```

API/Helpers/PdfComicInfoExtractor.cs
microtherion: I picked Debug in analogy to the same message in `BookService.cs`. Should the other message also be changed?

API/Helpers/PdfComicInfoExtractor.cs
microtherion: The outer conversion is entirely unnecessary, so I removed that, and I switched `GetFloatFromText` to `TryParse`.

microtherion:
I made the `PdfParser.cs` change, though I don't understand under what circumstances the now removed code could have been harmful.

I made the `info.Series` change, but the fallback I chose was `info.Title` not `info.TitleSort`. I assume this is what you wanted?

majora2007:
> I made the `PdfParser.cs` change, though I don't understand under what circumstances the now removed code could have been harmful.

The code will override the Series name which is what Kavita uses to group. This means even though you are reading it from the metadata, it would never be used. 
> 
> I made the `info.Series` change, but the fallback I chose was `info.Title` not `info.TitleSort`. I assume this is what you wanted?

Correct. We should reorder the TitleSort to be with the Title. 


I did some further testing last night, but while the metadata is being read in successfully, I'm not seeing Genres set or it used (in Kavita), at least with a file I manually tagged with Calibre. 

This still needs a bit more testing around ensuring the metadata flows through into Kavita as users would expect.

API/Helpers/PdfComicInfoExtractor.cs
majora2007: No, I suppose this should stay the same.

microtherion:
> The code will override the Series name

But only if the name previously ended with `.pdf`, right?

majora2007:
> > The code will override the Series name
> 
> But only if the name previously ended with `.pdf`, right?

Yes but considering this code only runs for PDF files, this would always be true.

majora2007:
Code looks fine, going to run this through my library once more and ensure all the cases are met.

majora2007:
Please make this change and then this is good to merge.

Remove the pdf check so Series-level metadata can be set. 
image

majora2007:
I'm going to merge this into my branch and do a bit of further cleanup to get it released. I want to extend a major thanks for driving this feature through. I know our users will benefit greatly from this.

microtherion:
Thanks for your patience in shepherding me through this!

PR #3141, opened 2024-08-29
Added Dynamic Conversion of JPEG-XL, JPEG 2000, HEIF and AVIF to JPG when the browser does not support them.

maxpiva:
# Added
- Support for JPEG-XL, JPEG 2000, HEIF and AVIF.
- Dynamic transcoding to JPEG in the cache, when the browser does not have support for the image types.
- Direct serving when the browser supports the image type.
- Transcoding of the thumbnails.
- Added ImageConverterServices, with ImageConverterProviders to easy expansion.
- Edited Dockefile, adding required library support for reading the new image formats in Linux versions. [Bumped ubuntu to oracular, because of the JPEG-XL image support]
- Added Magick.NET nuget for reencoding support [Toyed with the idea, of using vips, but netvips binding does not support the 'ALL version', so netvips binding need to be recompiled, mantained, etc, dropped in favor of Magick.NET, until netvips support the above image types]
- [Not Related] Update nugets to latest, Flurl Http changed the way its configured, edited accordingly.

# Missing
- Additional Testing methods.


API/Controllers/ReaderController.cs
```diff
@@ -23,6 +23,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Tokens;
 using MimeTypes;
+using Org.BouncyCastle.Ocsp;
```
majora2007: Remove this

API/Extensions/FileTypeGroupExtensions.cs
```diff
@@ -22,4 +28,18 @@ public static string GetRegex(this FileTypeGroup fileTypeGroup)
...
+    /// <summary>
+    /// Gets the MIME type for the specified file format. Extends original MimeTypeMap adding non supported extensions by the nuget.
+    /// </summary>
+    /// <param name="format">The file format.</param>
+    /// <returns>The MIME type.</returns>
+    public static string GetMimeType(this string format)
+    {
+        // Add jxl format
+        format = format.ToLowerInvariant();
+        if (format == ".jxl")
```
majora2007: return on the same line please

API/Extensions/FileTypeGroupExtensions.cs
```diff
@@ -22,4 +28,18 @@ public static string GetRegex(this FileTypeGroup fileTypeGroup)
...
+            return "image/jxl";
+        return MimeTypeMap.GetMimeType(format);
```
majora2007, line 44: Newline before return statement

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +56,56 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
             response.Headers.CacheControl =  $"max-age={maxAge}";
         }
     }
+
+    private static void AddExtension(List<string> extensions, string extension)
+    {
+        if (string.IsNullOrEmpty(extension))
+            return;
+        if (!extensions.Contains(extension))
+            extensions.Add(extension);
```
majora2007, line 65: Please always use {} whenever using any if statement that isn't a direct control flow (continue/break/return).

For any control flow jump logic, they need to be on the same line as the if else use {}

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +56,56 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
             response.Headers.CacheControl =  $"max-age={maxAge}";
         }
     }
+
+    private static void AddExtension(List<string> extensions, string extension)
+    {
+        if (string.IsNullOrEmpty(extension))
+            return;
+        if (!extensions.Contains(extension))
```
majora2007, line 64: Newline before the if statement to help keep the code clear

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +56,56 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
...
+
+    /// <summary>
+    /// Retrieves the supported image types extensions from the Accept header of the HTTP request.
+    /// </summary>
+    /// <param name="request">The HTTP request.</param>
+    /// <returns>A list of supported image types extensions.</returns>
+    public static List<string> SupportedImageTypesFromRequest(this HttpRequest request)
+    {
+        var acceptHeader = request.Headers["Accept"].ToString();
+        string[] spl1 = acceptHeader.Split(';');
```
majora2007: Use var whenever possible

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +56,56 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
...
+        supportedExtensions.Add("png");
+        supportedExtensions.Add("gif");
+        supportedExtensions.Add("webp");
+        //Browser add specific image mime types, when the image type is not a global standard.
+        //Let's reuse that to identify the additional image types supported by the browser.
+        foreach (string v in split)
+        {
+            if (v.StartsWith("image/", StringComparison.InvariantCultureIgnoreCase))
+            {
+                string n = v.Substring(6).ToLowerInvariant();
```
majora2007: what is n? A different variable name would probably help.

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +56,56 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
...
+        string[] split = acceptHeader.Split(',');
+        List<string> supportedExtensions = new List<string>();
+
+        //Add default extensions supported by all browsers.
+        supportedExtensions.Add("jpeg");
+        supportedExtensions.Add("jpg");
+        supportedExtensions.Add("png");
+        supportedExtensions.Add("gif");
+        supportedExtensions.Add("webp");
+        //Browser add specific image mime types, when the image type is not a global standard.
```
majora2007, line 91: Newline above the comment to help separate the code please.

API/Services/CacheService.cs
```diff
@@ -98,13 +102,19 @@ public IEnumerable<FileDimensionDto> GetCachedFileDimensions(string cachePath)
             for (var i = 0; i < files.Length; i++)
             {
                 var file = files[i];
-                using var image = Image.NewFromFile(file, memory: false, access: Enums.Access.SequentialUnbuffered);
+                var dimension = _converterService.GetDimensions(file);
+                if (dimension == null)
```
majora2007: Why would dimension be null?

API/Extensions/ApplicationServiceExtensions.cs
```diff
@@ -33,6 +34,13 @@ public static void AddApplicationServices(this IServiceCollection services, ICon
         services.AddScoped<ITaskScheduler, TaskScheduler>();
         services.AddScoped<ICacheService, CacheService>();
         services.AddScoped<IArchiveService, ArchiveService>();
+
+        services.AddScoped<IImageConverterProvider, JpegXLImageConverterProvider>();
```
majora2007, line 38: Just for my own learning, how does this work?

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +56,56 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
...
+    public static List<string> SupportedImageTypesFromRequest(this HttpRequest request)
+    {
+        var acceptHeader = request.Headers["Accept"].ToString();
+        string[] spl1 = acceptHeader.Split(';');
+        acceptHeader = spl1[0];
+        string[] split = acceptHeader.Split(',');
+        List<string> supportedExtensions = new List<string>();
+
+        //Add default extensions supported by all browsers.
+        supportedExtensions.Add("jpeg");
```
majora2007, line 86: Can we use the extensions in Parser so this is all colocated in case we ever add more formats in the future?

API/Services/ImageConverterService.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <param name="converters">The collection of image converter providers.</param>
+        public ImageConverterService(IEnumerable<IImageConverterProvider> converters)
+        {
+            _converters = converters;
+        }
+
+        /// <inheritdoc/>
+        public Stream ConvertStream(string filename, Stream source)
+        {
+            IImageConverterProvider provider = _converters.FirstOrDefault(a => a.IsSupported(filename));
```
majora2007: How do you control the order of these?

API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -25,7 +25,7 @@ public static class Parser
 
     public static readonly TimeSpan RegexTimeout = TimeSpan.FromMilliseconds(500);
 
-    public const string ImageFileExtensions = @"^(\.png|\.jpeg|\.jpg|\.webp|\.gif|\.avif)"; // Don't forget to update CoverChooser
+    public const string ImageFileExtensions = @"^(\.png|\.jpeg|\.jpg|\.webp|\.gif|\.avif|\.jxl|\.heif|\.heic|\.j2k|\.jp2)"; // Don't forget to update CoverChooser
```
majora2007, line 28: Looks like you forgot to update CoverChooser component on the UI

Dockerfile
```diff
@@ -14,15 +14,15 @@ RUN /copy_runtime.sh
 RUN chmod +x /Kavita/Kavita
 
 #Production image
-FROM ubuntu:focal
+FROM ubuntu:oracular
```
majora2007, line 17: @DieselTech

majora2007 (commented):
I looked over the code and provided comments. 

I'm not ecstatic about ImageMagick. I'm not sure if it supports ARM deployments as well, so I would need confirmation. Likewise, changing the docker file base image. 

I would also need to test this locally to see if it's worth it to add support for transcoding on the fly and if the performance difference is worth it or not. 

I personally haven't added JXL because it's DOA with support from the browsers. Jpeg2000 is also quite rare. Owning the maintenance of this code for flavors of formats that aren't common doesn't make much sense to me. 

Since you're opening the PR out of nowhere, please provide justification or any supporting Feature Requests.

API/Extensions/ApplicationServiceExtensions.cs
maxpiva: Since all providers are IImageConverterProviders. You can DI them as IEnumerable<IImageConverterProviders> in the ImageConveterServices, and you will get the whole collection of Dependency Injected Converters added in here.

API/Extensions/HttpExtensions.cs
maxpiva: For sure, will do, also all the coding styles referenced.

API/Controllers/ReaderController.cs
maxpiva: I can only blame Resharper on this ;)
Wil do

API/Services/ImageConverterService.cs
maxpiva: Since, two Converters cannot support the same extension, there is no ordering needed.

API/Services/Tasks/Scanner/Parser/Parser.cs
maxpiva: I wasn't sure on this, Cover is converted to jpg in case of the new formats.

Dockerfile
maxpiva: move to oracular because the include JPEG-XL in there, focal, requires custom build the support. Docker worked really fine. But didn't test ARM.

DieselTech:
```
[Kavita] [2024-08-29 17:32:53.271 -04:00  60] [Warning] API.Services.ArchiveService [GetCoverImage] There was an exception when reading archive stream: Y:/Projects/Image Format Tests/JXL/Max Compression Test.cbz. Defaulting to no cover image
ImageMagick.MagickBlobErrorException: unable to open image 'Terminator - One Shot - 00 - FC.jpg': Permission denied @ error/blob.c/OpenBlob/3596
   at ImageMagick.MagickExceptionHelper.Check(IntPtr exception) in /_/src/Magick.NET/Exceptions/MagickExceptionHelper.cs:line 18
   at ImageMagick.NativeHelper.CheckException(IntPtr exception) in /_/src/Magick.NET/Native/NativeHelper.cs:line 20
   at ImageMagick.MagickImage.NativeMagickImage.WriteFile(IMagickSettings`1 settings) in /_/src/Magick.NET/Generated/Magick.NET.SourceGenerator/ImageMagick.SourceGenerator.NativeInteropGenerator/MagickImage.g.cs:line 8851
   at ImageMagick.MagickImage.Write(String fileName) in /_/src/Magick.NET/MagickImage.cs:line 7047
   at API.Services.ImageConversion.ImageMagickConverterProvider.Convert(String filename) in Z:\Projects\Kavita\API\Services\ImageConversion\JpegXLImageConverterProvider.cs:line 57
   at API.Services.ImageConverterService.ConvertStream(String filename, Stream source) in Z:\Projects\Kavita\API\Services\ImageConverterService.cs:line 73
   at API.Services.ImageService.WriteCoverThumbnail(String sourceFile, Stream stream, String fileName, String outputDirectory, EncodeFormat encodeFormat, CoverImageSize size) in Z:\Projects\Kavita\API\Services\ImageService.cs:line 252
   at API.Services.ArchiveService.GetCoverImage(String archivePath, String fileName, String outputDirectory, EncodeFormat format, CoverImageSize size) in Z:\Projects\Kavita\API\Services\ArchiveService.cs:line 236
```

The filename it is trying to open doesn't exist in the archive. The file inside `Max Compression Test.cbz` is `Terminator - One Shot - 00 - FC.jxl`

Dockerfile
maxpiva: Yes, adding MagicK was the last thing I wanted to do, problem is, while vips supports all the image formats, netvips binding and libraries are boundled with the standard version, not the ALL version.

Will test on ARM, tested on X64 docker, and windows without an issue. git said, every flavor is supported, but, yes,  testing will be needed for ARM.

I'm not going to enter the browser war, or format war, just adding the capability of supporting multiple formats and be dynamic about it, independent of the browser, seems the way to go.  And since, from my own testing JPEG-XL seems to be a techincal marvel.

maxpiva:
@DieselTech  WIll fix. In that case, the file need to be created in a temp directory. 

Viewer works fine in your machine?

API/Services/CacheService.cs
maxpiva: _converterService will only work with their supported formats, will refactor, so, when the format is not supported, will fallback to netvips.

maxpiva:
@DieselTech Please retry, thanks in advance. Seems fixed, but the devil never rests.

For completion's sake I'll add two TODO to my list.

1) using the list of images in the Parser, in the Accept -> SupportedImageTypes, as suggested by @majora2007 

2) Currently when generating covers, the converter uses a temporary file to create a jpg, that can be future converted into a thumbnail with netvips. This is subpar, it will be better to directly use MagicK in that case to generate the thumb instead netvips, to avoid the extra conversion.

DieselTech:
Preface: These comments are before your latest commit. I am seeing the images in the browser. However it seems like it's trying to call the image endpoint twice, once with the proper image number and a second time that is using page # 1000000 

`api/reader/image?chapterId=1&apiKey=6e2885d7-e627-466b-b365-56863ad98995&page=1000000`

image

One thing I would like to see is a log entry for the conversion. Not everyone runs these servers on powerful hardware, and it will become necessary when troubleshooting with them why their page loads are taking 5 seconds. That way we can point directly to the conversion task.

maxpiva:
Totally agree on slow servers, 

on/off flag?

The delay exists, mostly, because there is frontend caching, client will try to get multiple images, and Kavita will trigger the conversion of them all.
What I see from my end, is, initial delay, 1-2 secs, then normal behavior. since the frontend caching mechanism, hides the additional delay after the initial load, docker with 1 core assigned.

Also, the transcoding only happens once. 
Normal Kavita behavior seems to be ARCHIVE -> CACHE -> BROWSER
The conversion happens in the cache, so after the initial conversion, always the converted image will be used going forward. until you clear the read cache.




IDK about the page 1000000. 

I can add a stopwatch.

DieselTech:
> Totally agree on slow servers,
> 
> on/off flag?

Joe will have to answer this one. There are some things that we do give the user a choice in, but otherwise it's opinionated in order to make the compromise of not layering complexity over and over. I would likely say that IF the image does get converted, to output its duration. If it doesn't get converted, then do not log it. 


> The delay exists, mostly, because there is frontend caching, client will try to get multiple images, and Kavita will trigger the conversion of them all. What I see from my end, is, initial delay, 1-2 secs, then normal behavior. since the frontend caching mechanism, hides the additional delay after the initial load, docker with 1 core assigned.

Honestly I couldn't see a delay at all. The image API responded within 10ms and the page loads felt instant to me. I had to double check to see if it was even working correctly.  But using my workstation as a testbed is an unfair comparison with how much power it has. That is why I thought about the log entry so we can know the time taken. 
 
> Also, the transcoding only happens once. Normal Kavita behavior seems to be ARCHIVE -> CACHE -> BROWSER The conversion happens in the cache, so after the initial conversion, always the converted image will be used going forward. until you clear the read cache.

Okay so they should stay around until the users cleanup tasks kick in, which defaults to daily. 

That said, you sent this PR right as the lead dev is going away for a little bit. This will need to sit until he gets back and looks more in-depth into it. If you have discord you are welcome to join our server and we can communicate in real time there.

maxpiva:
No problemo, will resolve the remaining issues and add some timings to the logging. Will join you guys in discord, tomorrow.

majora2007:
Is this ready for a proper review? If so, I can get it scheduled. I'm still not 100% sure I want to merge this in, but i do want to give it a fair test and also see if there is potential to remove NetVips and ImageSharp altogether, because that could allow for Kavita to run on CPUs that don't have SSE4.2.

maxpiva:
Define proper ;)
I did some research on netvips usage, and one of the main issues, is the way that covers are made, it supports that center on 'subject' on cover creation, and I think is triggered when the aspect ratio does not match.  If you can live without it. I think is feasible to remove both. The other is backporting that functionality from netvips code base.

majora2007:
Haha proper means, me pulling everything down, testing, polishing the code for actual inclusion. 

Yeah, I recently added that attention type of scaling, which helps when users need covers with webtoons, which are long strip format. That is a feature we might see if we can build with ImageMagick.NET.

maxpiva:
maxpiva/kavita:nightly

 if you want to test the docker. x64/armv7/arm64 (It has your lastest changes). (Release version).

Build and debug from windows works as expected.

On my machines :)

Let'me sit a little with the code, I'll see if we can remove both dependencies.... After that you can take the decision or not to merge. 
IOS 18 did a big push on JPEG-XL, so the codec is in a better place today.

maxpiva:
@majora2007 I have a new branch on my fork.

https://github.com/maxpiva/Kavita/tree/RemoveSixLaborsAndNetVips

This one, removes all dependencies on NetVips and ImageSharp and replaces it with ImageMagick.

Only differences so far:

1) Colorspaces colors are kinda more vibrant, probably related to the resizing interpolation algorithm.
2) The Attention (Focus on webtoons) is not ported over, it uses a lots of matrix transformation, so, it might take some time.

Of course, the above fork includes the (JP2, JPX, HEIF, AVIF) transcoding depending on what the browser supports.

On iOS per example, all is served directly.

The docker image is updated to this fork.

majora2007:
Awesome, I'm wrapping up some work on the scanner and while I collect testing from the community, I'm planning to give this a run, especially the new fork. 

If needed, we could leave ImageSharp to help with the matrix math. NetVips is what brings the CPU dependency on the project. 

I'm hoping for favorable results to bring more support to the project and to more people.

maxpiva:
Updated, instead of using netvips attention, using a direct replacement ported to C#.

from https://github.com/muesli/smartcrop

So far so good.

Docker is also updated.

majora2007:
I have officially started testing the full NetVips removal branch.

maxpiva:
Nice, probably a refactor will be needed afterwards. We have salted image conversion, thumbnails, and image processing, maybe we can join everything into a service/interface, and use ImageMagick as the first implementation, in that way, if you want to swap motors, another implementation can be created.

majora2007:
Yeah that's a good idea. I was also looking that jpg was selected as default image format. We might want to think about making it a server setting or even choosing png/webp. Given that Kavita supports major 2 browsers and webp is now supported across the browsers nowadays. 

I def think we can likely remove the conversion service and move it within the Image service and refactor quite a bit to streamline the code.

majora2007:
Since we don't have a PR for the other branch and in reality if I'm merging this, I want to use that other branch, here are some results from testing. So far, it's pretty close and looks fine from my perspective. I included the unit test results as well. 

From the other branch, attention port:
Old
image

New
image

Old:
image

New:
image



[Existing Code.zip](https://github.com/user-attachments/files/17247157/Existing.Code.zip)

[New Code.zip](https://github.com/user-attachments/files/17247287/New.Code.zip)

majora2007:
Okay I'm good to move forward with your NetVips removal branch. Let's clean up the code then I'll do another pass on it myself and we can move this into the nightly. We will need more documentation on the methods as well to cover our bases in case something breaks down the line.

maxpiva:
You want me to refactor the Image Service or, and create documentation about the new methods before merging, I think Image Service is a little bloated, maybe a secondary service, that handles image operations.

Yes currently, if the format is not supported, it'll be converted into jpg with 99 quality. Same for covers. 
We could change that from the config, but we can actually do that in a next iteration from fixed to be driven from setting?

I could do a PR from the other branch.

majora2007:
Yeah, let's close this PR and recreate a new one from the other branch. Let's add documentation and refactor the code to be more streamlined. 

Let's make the format a const in ImageService (jpg or png) so we don't have magic strings. I think png/webp could be good targets or even having some sort of tier like webp then png depending accept response headers could be good. 

Once that's done, I'll merge into a branch here, then do a round of cleanup to make sure the style is just as I like it and we can finally merge into develop to get into the hands of nightly testers.

kleisauke:
> [...] but netvips binding does not support the 'ALL version', so netvips binding need to be recompiled, mantained, etc, dropped in favor of Magick.NET, until netvips support the above image types

This is not correct. The NetVips bindings _does_ support JPEG XL, JPEG 2000, HEIC and AVIF. However, except for AVIF, the pre-built binaries (i.e. `NetVips.Native`) doesn't support these image formats.

With the update of `ubuntu:oracular`, you can just remove `NetVips.Native` and install libvips with all these formats enabled via:
```console
$ sudo apt-get install libvips --no-install-recommends
```

maxpiva:
Good call @kleisauke. I was unable to make it work on Windows. Since we need to create a custom native and maintain it, correct? 
But docker is another beast.

@majora2007 has some secondary concerns related to Netvips, and some CPU limitation that brings to the table on docker.

Anyway, since the current image processing implementation is decoupled into an interface and a ImageMagick is an implementation, we could create another implementation supporting NetVips. The question is, we want to?

This PR, got overseeded by this one:

https://github.com/Kareadita/Kavita/pull/3250

kleisauke:
> I was unable to make it work on Windows. Since we need to create a custom native and maintain it, correct?

For Windows, you could use the [<code>vips-dev-w64-<b>all</b>-x.y.z.zip</code> variant](https://github.com/libvips/build-win64-mxe#libvips-all-dependencies), but that contains ALL the file format loaders that libvips supports, including some very minor ones, and a complete copy of ImageMagick. It can process many different file types, but it is also rather vulnerable to hacking.

> [...] and some CPU limitation that brings to the table [...]

The CPU concerns only applies to the pre-built Linux binaries provided by NetVips, see: https://github.com/Kareadita/Kavita/issues/1423#issuecomment-1213071271.

majora2007:
My 2 cents is if we can keep NetVips without any CPU dependency, I would like that. NetVips is MUCH faster than ImageMagik.

maxpiva:
@kleisauke  I was unable to make it work on Windows, do you have a way to use the net bindings with that vips-dev, i was unable to make it work without the netvips native nugets. Including copying the libraries manually.

I agree that netvips is faster than ImageMagick, but mantaning our custom netvips.native will be probably a lots of pain.

I could also make some defines what is enabled with #define at build time. if we solve the issue with netvips.

for netvips to work. We need to:

- Have a NonDocker version working with all the image types. (netvips.native like but working).
- Have a docker image created that supports all the image types, but do not have specific CPU limitations. IE: Only working with AVX2. etc.,

Addtional, I'l add the capability of having netvips implementation ready inside kavita. In that way, we can actually by putting a conditional define in the build, we can create with one or another.... 

This will open the possibility to create a docker image/layer using netvips, and other using ImageMagick. If the CPU limitations are still an issue.

kleisauke:
Hmm, that should work without issues; I regularly test the `-all` variant on the NetVips test suite. Did you add the directory containing the DLLs to your `PATH` environment variable?

I won't distribute the `-all` variant on NuGet, as it contains GPL-licensed libraries. Regarding the other image formats not supported by the `-web` variant and mentioned here:
- JPEG XL: support within `-web` is being tracked in https://github.com/libvips/build-win64-mxe/issues/40.
- JPEG 2000: excluded from `-web` as it offers (in my opinion) minimal advantages. It primarily serves as a compatibility solution for legacy software that still uses it. Also note that Safari was [the only browser](https://caniuse.com/jpeg2000) to provide support for this format, and has removed support for this in version 18, see:
  https://webkit.org/blog/15443/news-from-wwdc24-webkit-in-safari-18-beta/#images
- HEIC: excluded due to its reliance on HEVC, which is patent-encumbered.

maxpiva:
@kleisauke  Thanks. Will start testing again. @majora2007. Could you ask @kleisauke about the docker CPU issues, how we can bypass that?

@majora2007  Leme try this out, initially since the native nuget couldn't be used in this case, will try to add a download/unzip or include a custom tool in the normal build pipeline (no docker), and for the docker pipeline do, what @kleisauke recommended.

majora2007:
I know this is closed, but I wanted to check with you @kleisauke if I'm understanding you correctly. 

You're saying that NetVips can support Formats JPEG XL/2000, HEIC, and AVIF via a native install of `libvips` and keeping the NetVips C# package (but removing `NetVips.Native`). 

For non-docker installs, `NetVips.Native` will provide all the needed file formats, like it is today? 

My biggest concern is how to get out of the box support of all the formats without user intervention.

kleisauke:
> You're saying that NetVips can support Formats JPEG XL/2000, HEIC, and AVIF via a native install of `libvips` and keeping the NetVips C# package (but removing `NetVips.Native`).

Correct. Though, AVIF should already be supported by `NetVips.Native`. Also, I would be careful with HEIC, as its underlying HEVC codec is _not_ royalty-free.

> For non-docker installs, `NetVips.Native` will provide all the needed file formats, like it is today?

The prebuilt binaries provided by NetVips rely on web-focused dependencies, which are generally safe for untrusted input and are continuously fuzz-tested via OSS-Fuzz - see https://github.com/google/oss-fuzz/tree/master/projects/libvips. It will only support patent-encumbered codecs that are made available on a royalty-free basis.

maxpiva:
@majora2007 
Reading between the lines....

We can use netvips on docker releases and support all HEIC, AVIF, JPEG2000, JPEGXL.... since native libraries will be installed separately as dependencies in the docker. (We don't use NetVips.Native). and we can forget about SixLabors and ImageMagick dependencies, since they're not going to be used anymore.

But, on normal windows build, NetVips.Native does not support JPEG2000, JPEGXL, and HEIC. And there is no native nuget build maintained with those features, so, we need to figure out, if we able to download a binary release for windows from github supporting those features, dynamically or at build time, or build it ourselves and mantain, to be used instead NetVips.Native, that, if we want to get rid of SixLabors and ImageMagick.

majora2007:
One question I have @maxpiva is where is JPEG2000 and HEIC coming from? I have only started seeing JPEGXL in the comic/book scene, but I've not seen JPEG2000 or HEIC. If they aren't common or up-and-coming formats, we can omit them and focus on XL and like you said, ditch SixLabors and ImageMagick.

maxpiva:
> One question I have @maxpiva is where is JPEG2000 and HEIC coming from? I have only started seeing JPEGXL in the comic/book scene, but I've not seen JPEG2000 or HEIC. If they aren't common or up-and-coming formats, we can omit them and focus on XL and like you said, ditch SixLabors and ImageMagick.

Sorry meant to be JPEGXL, the fat builds contain also JPEG2000 and HEIC support. Mix and match will be difficult. I think there is dinamic build, that reads the plugins from a directory. But it doesn't come as package. I think there was some jpeg2000 in the early days of the format, I don't think is used right now, it just come free with the fat builds, so, why not add it. HEIC IDK, if create concerns, we just remove our support.

majora2007:
I think we can realistically drop JPEG2000/HEIC and just focus on JPEGXL. It sounds like this is a doable feature and we can keep NetVips/NetVips.Native (which offers better cpu/memory performance) for Native builds and update our docker to have the full package as provided by Kleis. 

I'll try do a bit of work this weekend to test out the NetVips and JPEGXL and see if we can then rework transcoding support. Many people expressed interest in this and it would make Kavita more resilient.

Thanks @kleisauke for the help and quick reply.

maxpiva:
Well, the problem is not HEIC or JPEG2000, the problem is the support of the Windows builds. If we get JPEGXL running, we get HEIC and JPEG2000 for free. netvips.native does not support jpegxl. if we add the fat binaries, all become available.

So, the critical path, is getting the Windows version decode JPEGXL through netvips, instead imagemagick. That support cannot come from netvips.native since it does not include support for that format...

In the case, of docker, there is no issue, native dependencies, will take care.

majora2007:
But looking at the csproject, I see we have:
```
<PackageReference Include="NetVips" Version="3.0.0" />
<PackageReference Include="NetVips.Native" Version="8.16.1" />
```

Are you saying just removing NetVips.Native will include the fat build?

maxpiva:
No, NetVips is the C# wrapper around libvips.  
NetVips.Native is libvips compiled for every distribution, IE: Windows X64, X86, Linux X64, X86, ARM, etc.

NetVips.Native solves the issue, of installed libvips, on the machine.

But, the compiled libvips version that NetVips.Native includes, does not support a bunch of other formats, one of them being JPEGXL.

So, if you use netvips.native, you don't get jpegxl support....


Now, if we remove netvips native, and we install libvips fats ones (The ones including all the extra image supports), as dependencies in docker, we get JPEGXL support in docker, since the one installing libvips into the docker, are those dependencies and not NetVips.Native nuget.

The problem happens on the Windows version, the windows version does not have a way to install libvips other than NetVips.Native, which does not include JPEGXL.


The way to solve this, is to manually install a fat libvips on the windows build, without relying on the native nuget.

majora2007:
Okay then it's a no-go. We can't ask users to install stuff on their machine to use features out of the box. Kavita.exe should just work out of the box.

maxpiva:
Not necessary, I need to figure a way to install libvips fats at build time, in the Windows versions without relying on netvips.native. I failed last time I checked. But will go again trough that rabbit hole. If that is solved, we can go all the way with netvips.

PR #3164, opened 2024-09-14
Use KMeans++ For ColorScape Generation

F0x1:
# Changed
- Changed: Switched to K-Means++ clustering algorithm for improved initial centroid selection and parallelized some of the code. 

Implemented K-means++ for improved initial centroid selection
Parallelized point assignment to nearest centroids
Added early termination when centroids converge
Optimized distance calculations using squared distances

Performance Improvements

Execution time: 15.86x faster (3.58 ms, down from 56.85 ms)
Memory allocation: 99% reduction (532.07 KB, down from 97,486.85 KB)
GC Gen0 collections: 99% reduction (175.7, down from 31777.7)

This Optimizing has passed test cases and returns the same values as the original function ensuring maintained functionality alongside the performance gains.
This enhancements significantly improve the efficiency of parsing through different series, reducing resource usage and improving overall system performance.


API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -1106,9 +1107,21 @@ public static float MaxNumberFromRange(string range)
...
     }
 
+    private static readonly ConcurrentDictionary<string, string> NormalizedCache = 
+        new ConcurrentDictionary<string, string>();
+
     public static string Normalize(string name)
     {
-        return NormalizeRegex.Replace(name, string.Empty).Trim().ToLower();
+        // Check cache first
+        if (NormalizedCache.TryGetValue(name, out string cachedResult))
```
majora2007: Why would we want to cache every string that goes through normalization? That will balloon memory usage.

F0x1: Implementing caching seemed beneficial, given that the Normalize function may be invoked hundreds of times per series, depending on the volume of metadata and chapters. I observed that each series contains a significant amount of recurring metadata. However, I accidently overlooked the necessity of clearing the cache after processing each series.

majora2007: The Normalize is used throughout the codebase, not just the scan loop as well. So this sort of change would not work.

majora2007:
I'll take a look at your enhancements with K means++ clustering, but unfortunately it's too late to include with the initial release of the colorscape system as we have just wrapped up all the testing for the next stable.

F0x1:
That's fine, I appreciate you taking a look at least

majora2007:
I tested this and the colors it comes out with is a bit questionable. I've made a lot of tweaks recently wrt Colorscapes so I may need to revisit this and play with the grouping.

majora2007:
Decided to just close this PR. As mentioned the colors didn't come out well and I feel like the colorscape system is good enough for now. Thanks for making the PR.

PR #3179, opened 2024-09-18
Fix/base url 3146

### LINKED ISSUE #3146: Base URL not respected in Library URL and smart filter hrefs
### What happened?

The baseurl set up, is not respected/used in the library pages. Nor in the links to smart filtered in the dashboard customize page.

They're simply not used, which works fine while on the web app. But reloading the page, or opening it at a later time may cause issues.

I've made a video, with a side by side setup of the stable, and nightly branch.

https://github.com/user-attachments/assets/b318d7ef-e6d6-47ad-b499-ee53fddb1458
If you look closely, the /kavita/ part is included when going to it. But removed when the page loads, maybe this helps a bit.

The issue also occurs with the "load filter" references on this page. I've not included this in the video, as this is also an issue on the stable branch. (The code just doesn't add it)
image

### What did you expect?

The baseURL to be appended at all times.

Nightly Testing Branch

Docker (Dockerhub Container)

Linux

Firefox, Chrome, Safari

iOS

_No response_

### Relevant log output

_No response_

### Additional Notes

Relevant discord messages;

https://discord.com/channels/821879810934439936/851515194185744464/1278814185093857280
https://discord.com/channels/821879810934439936/851515194185744464/1277175419463794690
https://discord.com/channels/821879810934439936/851515194185744464/1275565898996125737

Fesaa:
# Fixed
- Fixed: Fixed base URL being ignored in some places (Fixes #3146)
- Fixed: Fixed base URL being removed after opening a series (Fixes #3146)

### DIFF
--- a/UI/Web/src/environments/environment.ts
+++ b/UI/Web/src/environments/environment.ts
@@ -6,6 +6,7 @@ const IP = 'localhost';
 
 export const environment = {
   production: false,
+  baseUrl: 'http://' + IP + ':5000/',
   apiUrl: 'http://' + IP + ':5000/api/',
   hubUrl: 'http://'+ IP + ':5000/hubs/',
   buyLink: 'https://buy.stripe.com/test_9AQ5mi058h1PcIo3cf?prefilled_promo_code=FREETRIAL',
--- a/UI/Web/src/environments/environment.ts
+++ b/UI/Web/src/environments/environment.ts
@@ -6,7 +6,7 @@ const IP = 'localhost';
 
 export const environment = {
   production: false,
-  baseUrl: 'http://' + IP + ':5000/',
+  baseUrl: '/',
   apiUrl: 'http://' + IP + ':5000/api/',
   hubUrl: 'http://'+ IP + ':5000/hubs/',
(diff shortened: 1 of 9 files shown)


Fesaa:
I tried to get it to work without adding a new field to the environment, but I couldn't get it to work nicely. It also doesn't look terrible at all like this. Would've been nice if we could just pass the query stuff, but Angular always encodes it again. So didn't work out, at least what I could find.

UI/Web/src/environments/environment.ts
```diff
@@ -6,6 +6,7 @@ const IP = 'localhost';
 
 export const environment = {
   production: false,
+  baseUrl: 'http://' + IP + ':5000/',
```
majora2007: Wouldn't this just be /?

Fesaa: Ah, yes, it probably should be. I had to test with a docker image, or the base url stuff didn't seem to work at all. So didn't notice

majora2007:
I took a stab at it and am almost done, I have it working without an additional environment variable and much less intrusive. The only problem in my code is smart filter doesn't recognize the baseurl and redirects to home when loading the filter.

PR #3242, opened 2024-09-30
draft: Feat add pgsql

Rob117:
Welcome to speed and concurrency. No more database locks because of SQLite, and insanely fast performance.

# Changed
- Changed: Removed SQLite, swapped to Postgres

# Todo
- Dynamic Database Connection String
- How to smartly apply schema to DB on bootup migration instead of manually doing so (not used to ASP.NET)


majora2007:
Rob, I don't want to discourage any PRs, but you should check in with me when trying to make massive shifts to the software. Postgres is not something I'm interested in. Kavita should be able to run without any extra DB setup or requirement on docker. To my knowledge that is not something anything but litedb (which is crap) or sqlite can support.

In addition, I like the manual migration system I have right now. It is very explicit and helps me recall information during support requests. 

I'm extremely open, but I have a vision and while I would love the performance of Postgres, it would not align with the project's goal of being simple to run.

Rob117:
I did not realize that I made this change against the main repo! I meant to make this against my own develop branch. I have no idea how I did this.

I hear you loud and clear. I made the fork _because_ you clarified somewhere else that you had no interest in other databases. I was running into a db lock issue while running Kavita on k8s and using a nfs to store files (including the sqlite db instance). It would lock up like clockwork until the container running Kavita failed a healthcheck for long enough and k8s killed it and redeployed. With this, it runs like lightning.

It does require a lot more configuration and is not a drag-and-drop style of solution. I also think your decision to keep Kavita on sqlite and make the code easy to follow and deploy is the correct choice. If I had intended to make a such a sweeping change and commit it to the upstream I definitely would have talked with you first.

I'm sorry for taking your time with this - of course, I appreciate the work that you do, and it was a breeze to modify the code because your work is so clean.

Thank you again for making Kavita!

PR #3250, opened 2024-10-04
[Canary] Transcoding Support + More Image Format Support (JPEGXL)

maxpiva:
Canary testers, I need your help! This is a massive PR that swaps out the Image system in Kavita and brings Transcoding support. This is built from the nightly. Please give it a serious stress test and report back on the [nightly thread](https://discord.com/channels/821879810934439936/1316389775426588732).

# Added
 - Added: Added support for JPEG-XL, JPEG 200, and HEIF image formats
 - Added: Kavita now has transcoding support. This means Kavita will respect Accept Headers and if the target browser/device doesn't support the image, Kavita will convert it on the fly. From testing this is not noticeable and without quality drop.
 
 # Changed
- Changed: Kavita should no longer require SSE4.2 CPU Instructions and should be able to run on older hardware. 
- Changed: Due to the massive overhaul to the image manipulation library, smart cropping (mainly visible on oddly-shaped cover images) was changed from NetVips to SmartCrop.js C# library. Minor differences from testing. 

# Developer
- Swapped out NetVips with ImageMagick (Magick.NET)
- Updated Flurl.Http with a new configuration 
- Removed NetVips and SixLabors.ImageSharp


adamradocz:
HEIF and AVIF are superior formats compared to JpegXL and Jpeg2000. Why the hassle to support them?

maxpiva:
> HEIF and AVIF are superior formats compared to JpegXL and Jpeg2000. Why the hassle to support them?

Apple might want to talk to you, while I don't want to start a format war, I want to know how much the google ~~chrome~~ AVIF team is paying you, I want the same deal! :)

Support came from changing netvips to ImageMagick we don't need to add specific support/mantain for JPEG-XL. There is a secondary why, @majora2007 probably can explain it better, but netvips limits the possible CPU targets.

What is added, is, we identify if your browser supports a new format A, B, C or D, and if not with transcode the image to a format that your browser/app supports. 

While is true that the original intent was adding JPEG-XL, let's say because my own bias, and current is in vogue. 

In the end, we end supporting AVIF, HEIF, JPEG-XL and JPEG2000 comics/manga, regardless your browser/app supports it or not.

API.Benchmark/ArchiveServiceBenchmark.cs
```diff
@@ -24,16 +24,16 @@ public class ArchiveServiceBenchmark
     private readonly ArchiveService _archiveService;
     private readonly IDirectoryService _directoryService;
     private readonly IImageService _imageService;
-    private readonly PngEncoder _pngEncoder = new PngEncoder();
-    private readonly WebpEncoder _webPEncoder = new WebpEncoder();
+    private readonly IImageFactory _imageFactory;
     private const string SourceImage = "C:/Users/josep/Pictures/obey_by_grrsa-d6llkaa_colored_by_me.png";
```
majora2007: We should probably remove this hardcoded path and point to an arbitrary image in TestData

maxpiva: Done

API/Entities/Enums/EncodeFormat.cs
```diff
@@ -2,12 +2,16 @@
...
+public enum EncodeFormat 
 {
     [Description("PNG")]
     PNG = 0,
     [Description("WebP")]
     WEBP = 1,
     [Description("AVIF")]
-    AVIF = 2
+    AVIF = 2,
+    //Internal Use
```
majora2007: Let's /// this so it shows in the IDE that this is internal only.

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +57,51 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
...
+                if (mimeimagepart.StartsWith("*")) continue;
+                if (Parser.NonUniversalSupportedMimeMappings.ContainsKey(mimeimagepart))
+                {
+                    Parser.NonUniversalSupportedMimeMappings[mimeimagepart].ForEach(x => AddExtension(supportedExtensions, x));
+                }
+                else if (mimeimagepart == "svg+xml")
+                {
+                    AddExtension(supportedExtensions, "svg");
+                }
+                else
```
majora2007, line 102: Please fix the formatting in this file. 

Space after start of a comment
Newline after 84 to help break up the code
Use var whenever possible
mimeimagepart should be mimeImagePart
Newline before return on line 105

API/Extensions/HttpExtensions.cs
```diff
@@ -53,4 +57,51 @@ public static void AddCacheHeader(this HttpResponse response, string filename, i
...
+        {
+            if (v.StartsWith("image/", StringComparison.InvariantCultureIgnoreCase))
+            {
+                string mimeimagepart = v.Substring(6).ToLowerInvariant();
+                if (mimeimagepart.StartsWith("*")) continue;
+                if (Parser.NonUniversalSupportedMimeMappings.ContainsKey(mimeimagepart))
+                {
+                    Parser.NonUniversalSupportedMimeMappings[mimeimagepart].ForEach(x => AddExtension(supportedExtensions, x));
+                }
+                else if (mimeimagepart == "svg+xml")
```
majora2007: Can we not use MimeType library?

API/Services/BookService.cs
```diff
@@ -1273,16 +1272,15 @@ private string GetPdfCoverImage(string fileFilePath, string fileName, string out
...
         using var pageReader = docReader.GetPageReader(pageNumber);
         var rawBytes = pageReader.GetImage(new NaiveTransparencyRemover());
         var width = pageReader.GetPageWidth();
         var height = pageReader.GetPageHeight();
-        var image = Image.LoadPixelData<Bgra32>(rawBytes, width, height);
-
+        IImage image = _imageService.ImageFactory.CreateFromBGRAByteArray(rawBytes, width, height);
         stream.Seek(0, SeekOrigin.Begin);
-        image.SaveAsPng(stream);
+        image.Save(stream, EncodeFormat.PNG, 100);
```
majora2007: You can leave off the 100 since it's default quality.

API/Services/CacheService.cs
```diff
@@ -91,20 +94,23 @@ public IEnumerable<FileDimensionDto> GetCachedFileDimensions(string cachePath)
         }
 
         var dimensions = new List<FileDimensionDto>();
-        var originalCacheSize = Cache.MaxFiles;
```
majora2007, line 94: I can't recall why I had this here, I believe it's around buffering for NetVips right?

API/Services/ImageService.cs
```diff
@@ -1,82 +1,164 @@
...
+
+    /// <summary>
+    /// Replaces the file format of an image with the specified supported image formats by the browser, if needed, otherwise original file will be served.
+    /// </summary>
+    /// <param name="filename">The name of the image file.</param>
+    /// <param name="supportedImageFormats">The list of supported image formats by the browser.</param>
+    /// <param name="format">The encoding format to convert the image. Default is JPG</param>
+    /// <param name="quality">The quality of the image. Default is 99.</param>
+    /// <returns>The file name with extension of the replaced image file.</returns>
+    string ReplaceImageFileFormat(string filename, List<string> supportedImageFormats = null, EncodeFormat format = EncodeFormat.JPEG, int quality = 99);
```
majora2007: Can we not switch to PNG by default? PNG should be better than JPEG because we already have it as a default for the whole app. I would really like to investigate this.

API/Services/ImageService.cs
```diff
@@ -1,82 +1,164 @@
...
 }
 
 public class ImageService : IImageService
 {
-    public const string Name = "BookmarkService";
+    public const string Name = "ImageService";
     private readonly ILogger<ImageService> _logger;
     private readonly IDirectoryService _directoryService;
     private readonly IEasyCachingProviderFactory _cacheFactory;
+    private readonly IImageFactory _imageFactory;
```
majora2007, line 134: Let's move all the consts together and services below.

API/Services/ImageService.cs
```diff
@@ -209,20 +292,18 @@ private static bool IsLikelyWideImage(int width, int height)
         return aspectRatio > 1.25;
     }
 
-    public string GetCoverImage(string path, string fileName, string outputDirectory, EncodeFormat encodeFormat, CoverImageSize size)
+
+    /// <inheritdoc/>
+    public string GetCoverImage(string path, string fileName, string outputDirectory, EncodeFormat encodeFormat, CoverImageSize size, int quality = 100)
```
majora2007: Is there ever a case when we wouldn't use quality of 100? Because I can't forsee one and think we should just hardcode the quality as a const within the ImageService.

API/Services/ImageService.cs
```diff
@@ -265,77 +331,57 @@ public string WriteCoverThumbnail(Stream stream, string fileName, string outputD
...
-            // Saving pdf covers from a stream can fail, so revert to old code
-
-            if (stream.CanSeek) stream.Position = 0;
-            using var thumbnail2 = Image.ThumbnailStream(stream, targetWidth, height: targetHeight,
-                size: scalingSize,
-                crop: scalingCrop);
-            thumbnail2.WriteToFile(_directoryService.FileSystem.Path.Join(outputDirectory, filename));
-
-            return filename;
+            return string.Empty; //IDK?
```
majora2007: This should still return the filename

API/Services/ImageService.cs
```diff
@@ -265,77 +331,57 @@ public string WriteCoverThumbnail(Stream stream, string fileName, string outputD
...
         var file = _directoryService.FileSystem.FileInfo.New(filePath);
         var fileName = file.Name.Replace(file.Extension, string.Empty);
         var outputFile = Path.Join(outputPath, fileName + encodeFormat.GetExtension());
-
-        using var sourceImage = Image.NewFromFile(filePath, false, Enums.Access.SequentialUnbuffered);
-        sourceImage.WriteToFile(outputFile);
-        return Task.FromResult(outputFile);
+        using var sourceImage = _imageFactory.Create(filePath);
+        await sourceImage.SaveAsync(outputFile, encodeFormat, quality).ConfigureAwait(false);
+        return outputFile;
```
majora2007, line 336: Newline before the return.

API/Services/ImageService.cs
```diff
@@ -265,77 +331,57 @@ public string WriteCoverThumbnail(Stream stream, string fileName, string outputD
...
+    public Task<bool> IsImage(string filePath)
     {
         try
         {
-            var info = await SixLabors.ImageSharp.Image.IdentifyAsync(filePath);
-            if (info == null) return false;
-
-            return true;
+            var result= _imageFactory.GetDimensions(filePath);
+            if (result!=null)
```
majora2007: You need to fix the spacing up around the code and the return needs to be on the same line or in {}.

API/Services/ImageService.cs
```diff
@@ -265,77 +331,57 @@ public string WriteCoverThumbnail(Stream stream, string fileName, string outputD
...
+    /// <inheritdoc/>
+    public Task<bool> IsImage(string filePath)
     {
         try
         {
-            var info = await SixLabors.ImageSharp.Image.IdentifyAsync(filePath);
-            if (info == null) return false;
-
-            return true;
+            var result= _imageFactory.GetDimensions(filePath);
```
majora2007: Actually just rewrite this as `return Task.FromResult(_imageFactory.GetDimensions(filePath) != null);` and keep the try/catch

API/Services/ImageService.cs
```diff
@@ -407,24 +453,9 @@ public async Task<string> DownloadFaviconAsync(string url, EncodeFormat encodeFo
                 .GetStreamAsync();
 
             // Create the destination file path
-            using var image = Image.PngloadStream(faviconStream);
+            using var image = _imageFactory.Create(faviconStream);
```
majora2007: This is for me to come back to, not sure where the encode format is being used in Create.

API/Services/ImageService.cs
```diff
@@ -480,24 +496,11 @@ public async Task<string> DownloadPublisherImageAsync(string publisherName, Enco
...
-        // Convert image to RGB array
-        var pixels = processedImage.WriteToMemory().ToArray();
-
-        // Convert to list of Vector3 (RGB)
-        var rgbPixels = new List<Vector3>();
-        for (var i = 0; i < pixels.Length - 2; i += 3)
-        {
-            rgbPixels.Add(new Vector3(pixels[i], pixels[i + 1], pixels[i + 2]));
-        }
+        var rgbPixels = _imageFactory.GetRgbPixelsPercentage(imagePath, 10);
```
majora2007, line 358: What does 10 mean? Try to avoid magic numbers.

API/Services/ImageService.cs
```diff
@@ -739,6 +688,30 @@ public static ColorScape CalculateColorScape(string sourceFile)
...
+    {
+        if (supportedImageFormats == null)
+            return false;
+        string ext = Path.GetExtension(filename).ToLowerInvariant().Substring(1);
+        return supportedImageFormats.Contains(ext);
+    }
+    /// <inheritdoc/>
+    public string ReplaceImageFileFormat(string filename, List<string> supportedImageFormats = null, EncodeFormat format = EncodeFormat.JPEG, int quality = 99)
+    {
+        if (CheckDirectSupport(filename, supportedImageFormats))
```
majora2007: Again please don't have things like:
```
if (foo)
  bar
```

API/Services/ImageService.cs
```diff
@@ -739,6 +688,30 @@ public static ColorScape CalculateColorScape(string sourceFile)
...
+            return filename;
+        Match m = Regex.Match(Path.GetExtension(filename), Parser.NonUniversalFileImageExtensions, RegexOptions.IgnoreCase, Parser.RegexTimeout);
+        if (!m.Success)
+            return filename;
+        var sw = Stopwatch.StartNew();
```
majora2007: We can remove the timing now.

API/Services/ImageService.cs
```diff
@@ -739,6 +688,30 @@ public static ColorScape CalculateColorScape(string sourceFile)
...
+        string destination = Path.ChangeExtension(filename, format.GetExtension().Substring(1));
+
+        using var sourceImage = _imageFactory.Create(filename);
+        sourceImage.Save(destination, format, quality);
+        File.Delete(filename);
```
majora2007: We probably want try/catch around Delete statements. I've seen some OSes throw permission errors.

API/Services/ImageService.cs
```diff
@@ -739,6 +688,30 @@ public static ColorScape CalculateColorScape(string sourceFile)
...
+        _logger.LogDebug("Image converted from '{Extension}' to '{format.GetExtension()}' in {ElapsedMilliseconds} milliseconds", Path.GetExtension(filename), sw.ElapsedMilliseconds);
```
majora2007: We can remove this logger too

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+
+    /// <summary>
+    /// Creates an instance of <see cref="ImageMagickImage"/> from a base64 string.
+    /// </summary>
+    /// <param name="base64">The base64 string representing the image.</param>
+    /// <returns>An instance of <see cref="ImageMagickImage"/>.</returns>
+    public static IImage CreateFromBase64(string base64)
+    {
+        ImageMagickImage m = new ImageMagickImage();
+        m._image = (MagickImage)MagickImage.FromBase64(base64);
```
majora2007: Space after cast

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+    public int Height => _image?.Height ?? 0;
+
+    /// <summary>
+    /// Creates an instance of <see cref="ImageMagickImage"/> from a base64 string.
+    /// </summary>
+    /// <param name="base64">The base64 string representing the image.</param>
+    /// <returns>An instance of <see cref="ImageMagickImage"/>.</returns>
+    public static IImage CreateFromBase64(string base64)
+    {
+        ImageMagickImage m = new ImageMagickImage();
```
majora2007, line 34: We should Assert or something if we base64 is invalid or null.

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+        //Convert to RGBA float array (Image Magick 16 uses float array with values from 0-65535)
+        var floats = new float[bgraByteArray.Length];
+        for (var i = 0; i < bgraByteArray.Length; i += 4)
+        {
+            floats[i] = bgraByteArray[i + 2] << 8;
+            floats[i + 1] = bgraByteArray[i + 1] << 8;
+            floats[i + 2] = bgraByteArray[i] << 8;
+            floats[i + 3] = bgraByteArray[i + 3] << 8;
+        }
+        ImageMagickImage m = new ImageMagickImage();
```
majora2007, line 79: Newline to separate code from for loop.

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+        m._image = new MagickImage(MagickColor.FromRgba(0, 0, 0, 0), width, height);
+        using var pixels = m._image.GetPixels();
```
majora2007, line 81: Newline so the using is made clear.

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+        pixels.SetArea(0, 0, width, height, floats);
+        return m;
```
majora2007, line 83: Newline to make clear the return as well.

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+        _image = new MagickImage(stream);
+    }
+
+    /// <summary>
+    /// Creates an instance of <see cref="ImageMagickImage"/> from an existing <see cref="MagickImage"/>.
+    /// </summary>
+    /// <param name="image">The existing <see cref="MagickImage"/>.</param>
+    public ImageMagickImage(MagickImage image)
+    {
+        _image = (MagickImage)image.Clone();
```
majora2007: Space after cast

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+    {
+        _image.Thumbnail(new MagickGeometry(width, height) { IgnoreAspectRatio = true });
+    }
+
+    /// <inheritdoc/>
+    public void Composite(IImage overlay, int x, int y)
+    {
+        ImageMagickImage tile = overlay as ImageMagickImage;
+        if (tile == null) return;
+        _image.Composite(tile._image, x, y, CompositeOperator.Over);
```
majora2007, line 159: newline after if

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
```diff
@@ -0,0 +1,227 @@
...
+    public Task SaveAsync(Stream stream, EncodeFormat format, int quality, CancellationToken token = default)
+    {
+        _image.Quality = quality;
+        return _image.WriteAsync(stream, MagickFormatFromEncodeFormat(format), token);
+    }
+
+    /// <inheritdoc/>
+    public float[] GetRGBAImageData()
+    {
+        float[] data = null;
```
majora2007: This either needs to be an empty array or needs nullable annotation.

API/Services/Tasks/Scanner/Parser/Parser.cs
```diff
@@ -25,15 +26,29 @@ public static class Parser
 
     public static readonly TimeSpan RegexTimeout = TimeSpan.FromMilliseconds(500);
 
-    public const string ImageFileExtensions = @"^(\.png|\.jpeg|\.jpg|\.webp|\.gif|\.avif)"; // Don't forget to update CoverChooser
+    public static Dictionary<string, List<string>> NonUniversalSupportedMimeMappings = new Dictionary<string, List<string>>()
```
majora2007, line 36: You need to leave some documentation about when to add certain formats to each section and how it works. For example UniversalFileImageExtension means all major browsers must support the format.

API/Services/ImageService.cs
majora2007: I looked through the PR and don't see any place where the EncodeFormat is being used to appropriately ensure the file is outputted in the correct format. Can you help me find these cases?

majora2007 (changes requested):
I went through the whole PR. Please address comments and style changes. 

I'm worried about EncodeFormat not being taken into consideration and I want to research a way to remove JPEG as the default format as we can use PNG instead.

API/Services/CacheService.cs
maxpiva: I have no idea, but yes, it's related to netVIPs Constants.

API/Services/ImageService.cs
maxpiva: Hmm. The choice in here is JPEG, is because the encoding speed. PNG tends to be slower than JPG. Since, time in this case is essential to have a good Experience, I choose JPEG as a compromise. Anyway, this is from my own experience, not specifically from ImageMagick timings. 

Let met time log, both cases and we can compare numbers.

API/Services/ImageService.cs
maxpiva: I just standardized the service calls, so is up to the consumers to change the default values, totally subjective. If you want it back, we can.

API/Services/ImageService.cs
maxpiva: Sure

API/Services/ImageService.cs
maxpiva: 10% of the size. It was hardcoded on the original colorspace.

API/Services/Tasks/Scanner/Parser/Parser.cs
maxpiva: Yes, to all that I didn't comment before. 

btw, what IDE are you using, can I get your styles also? Visual Studio is Rewriting the UTF-8 BOM every time it can... Also, I notice, some .cs has BOM, but they're a few that do not have.

majora2007:
I saw a comment about what IDE do I use last night, but wasn't able to find it. I use Rider. We shouldn't have BOM. It's likely due to PRs submitted by the community where it might have been added.

therobbiedavis:
> I saw a comment about what IDE do I use last night, but wasn't able to find it. I use Rider. We shouldn't have BOM. It's likely due to PRs submitted by the community where it might have been added. 

By default VS will save with BOM. I had this issue as well and I believe you can change this default behavior to save without BOM.

API/Entities/Enums/EncodeFormat.cs
maxpiva: Done

API/Extensions/HttpExtensions.cs
maxpiva: Done

API/Extensions/HttpExtensions.cs
maxpiva: Not sure, this is a mime mapping against multiple extensions, added documentation to the parser stating how it works.

API/Services/BookService.cs
maxpiva: New IImage interface,  do not have default value, we can add it if you want.

API/Services/CacheService.cs
maxpiva: Yup, removed

API/Services/ImageService.cs
maxpiva: Some data, backing the JPEG 99 choice:

Big JXL 48 Bit RGB Image (JXL Loading time 1.7 seconds)

Whole Conversion time (Loading & Saving)
```
WEBP - 100 -  7.3 seconds
AVIF - 100 -  4.2 seconds
PNG  - 100 - 11.1 seconds (Probably deflate max)
PNG  -   0 -  2.1 seconds (Most likely, uncompressed, big file)
JPEG -  99 -  1.9 seconds (Best bang for the buck)
JPEG - 100 -  3.4 seconds

```

So, JPEG 99, takes (1.9s - 1.7s) ~200 ms to save against the other formats.


Please NOTE:
I also added, Per Image Lock on conversion, to prevent multiple threads from processing the same image at the same time.

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: Tell me if you want, to change the defaults, or remove from the signature.

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: The encoding and quality, are two lines below

           await image.SaveAsync(Path.Combine(_directoryService.FaviconDirectory, filename), encodeFormat, quality);

API/Services/ImageService.cs
maxpiva: Done. I guess. Not Sure really ;)

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageService.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/ImageServices/ImageMagick/ImageMagickImage.cs
maxpiva: Done

API/Services/Tasks/Scanner/Parser/Parser.cs
maxpiva: Edited on rider. 🤞

maxpiva:
> > I saw a comment about what IDE do I use last night, but wasn't able to find it. I use Rider. We shouldn't have BOM. It's likely due to PRs submitted by the community where it might have been added.
> 
> By default VS will save with BOM. I had this issue as well and I believe you can change this default behavior to save without BOM.

I did just that, then some other files, got BOM when they original did NOT.

Swapped to Rider, for this, Rider  does not touch the BOM (Nor add, nor remove)

API/Services/BookService.cs
majora2007: I don't see ever a need to not generate at 100 quality, so we should probably just leave it off entirely. If there is ever a need, it's easy to refactor the signature.

maxpiva: Well, if I remove the 100, I cannot control JPEG image quality required below, https://github.com/Kareadita/Kavita/pull/3250#discussion_r1796091070

if you want, I can add a default value to the interface.

FYI (ImageMagick)

100 PNG = Maximum Compression. loseless
0 PNG = No Compression. loseless
100 JPG = Loseless JPG
99 - 1 = Lossy JPG
AVIF and WEBP I think they're similar to JPG.

majora2007:
@maxpiva if you're on discord, can you reach out to me? I want to discuss about the quality a bit more and get a better understanding. 

My biggest questions are about the quality (100, 0, 99, etc) and how to know when to use what and the time delta you provided for PNG vs JPEG for transcoding, what type of image did you use and what is the quality delta between (is it noticeable to the user)?

API/Services/ImageServices/IImage.cs
```diff
@@ -64,15 +64,15 @@ public interface IImage : IDisposable
     /// <param name="filename">The name of the file to save the image to.</param>
     /// <param name="format">The format to save the image in.</param>
     /// <param name="quality">The quality of the saved image.</param>
```
adamradocz: Forgot to remove the quality comments. ;)

majora2007:
Just remembering that this impacts the docker image and that might mean we need to check with unraid/lsio/qnap where they make customizations.

@DieselTech can you confirm?

If this is true, we will need to pair this with a release that isn't primarily bugfixes (which is what the upcoming one is).

maxpiva:
Working in unraid as is, using (original unraid template) but changed the image path to my docker repo.

API/Services/ImageServices/IImage.cs
maxpiva: Doh. :O Done

majora2007:
Note to self: https://github.com/Kareadita/Kavita/pull/3164 should get merged in with this Transcoding PR so any Colorscape effects can happen together.

majora2007:
Just an update, I'm wrapping the current canary testing. https://github.com/Kareadita/Kavita/pull/3025 gets first priority to the canary testing (as it's likely to be a shorter test phase) then this is up. If I can, I will get this into this v0.8.4 release (as otherwise it's a bugfix release). 

Thanks for your patience. I think this will be a nice enhancement to Kavita and allow for much better format support in the future.

maxpiva:
Nice :) 
Hope the Panels guys add image/jxl in the accept header, since IOS supports it. Safari on iOS was working on all the new image types without transcoding.

majora2007:
I reached out to Dani from Panels and they will add Accept headers to their requests so we don't need to always transcode.

majora2007:
@maxpiva So what was the consensus of the NetVips stuff? Wondering if I can slot this in the canary (since the Koreader one needs some more work).

maxpiva:
@majora2007  IDK. Asked the guy how to make it possible, but have no answer, is he/she in the discord channel?

Ref:
https://github.com/Kareadita/Kavita/pull/3141#issuecomment-2432815434

if we make it work at OS level. I could code also the netvips implementation, and we can choose at build time which one to use.

majora2007:
> @majora2007 IDK. Asked the guy how to make it possible, but have no answer, is he/she in the discord channel?
> 
> Ref: [#3141 (comment)](https://github.com/Kareadita/Kavita/pull/3141#issuecomment-2432815434)
> 
> if we make it work at OS level. I could code also the netvips implementation, and we can choose at build time which one to use.

That is the maintainer of NetVips actually. They aren't in the discord. I'm not sure if it can be possible as well have native Linux builds as well. Let's move forward with canary testing for now.

majora2007:
Okay the time is near. Once you get the merge conflicts out of the way, I'm good to deploy this to our canary branch and start extended testing with our dedicated testers.

maxpiva:
Sorry Reopen the PR by mistake

maxpiva:
Not the right or, is the other one. Also will commit some changes that
gives us 2x and 4x transcoding speed, testing right now, will or later today

majora2007:
Sounds good. Once you are ready, give me a ping and I can get this deployed onto canary for our testers before I log off for the holiday season.

maxpiva:
@majora2007 @DieselTech 

Merge from Develop.
Added ImageMagick Q8, Q16 and Q16HDRI support.

The original implementation was using Q16HDRI which is 4 times memory heavier and slower than the other implementations.
Now we are using Q8 which is the lighter one.

Q8 - 8 Bit per channel per pixel processing on bytes. Similar to NetVips chain.
Q16 - 16 Bit per channel per pixel processing on ushorts. 16 Bit color precision.
Q16HDRI - 16 Bit per channel per pixel processing on floats. 16 Bit color precision, more mantissa and precision.

Right now, the .NET Proxy Library, Magick.Net does not support Side by Side, you must choose one implementation or the other at build time, so, currently the user, cannot choose the ImageMagick Implementation to use, without us, using a custom Magick.Net that native load the correct library, depending on ours choice.

Nevertheless, there is 3 new Define Constants, one can use to tailor Kavita in the future.

you can add to dotnet build : 

/p:DefineConstants="Q8"
or
/p:DefineConstants="Q16"
or
/p:DefineConstants="Q16HDRI"

Is nothing is included Kavita will be built with Q8.
That will build Kavita with that implementation, code, interop code is already tested on all implementations.

Q16 will make sense, if the user has a screen capable of more than 8 Bits per pixel, and the source has more than 8 Bit per pixel also, IE: HEIF, AVIF, JPEGXL.  Some Dithering bastards, could chime in, that is also good for 8-bit sources on 10-bit screens. ;)

In Regard NetVips, Supporting the new formats requires new tooling, docker needs to install netvips native, and I was unable to make it work on normal windows, since the native nuget dosn't support the extended formats, we need to make it download it and installing manually without breaking in the Windows Build stage, or do a .bat/.powershell.

As you see I abstracted all the image processes into interfaces, so swapping engines (ImageMagick x NetVips) is possible, since Kavita is now totally abstracted of the Image Processing libraries we use, and NetVips implementation should be easy to do, since we need to refactor only the existing NetVips actions into the interfaces.

The problem is the tooling to make NetVips viable again.

So far so good on my server.

majora2007:
I'm sorry @maxpiva, I created canary off the latest develop this morning and I guess there are some conflicts. Once you get them done, just dm me so I can merge and start the activities for the dedicated testers.

maxpiva:
Let me merge from Canary into this branch. Previously I merged and resolved from develop.

maxpiva:
@majora2007 
Merged with Canary. Runned backend tests and benchmarks. Seems fine.

FYI: Canary has differences in the frontend against develop (I just taked canary), also some version differences on Canary Kavita.Commons.csproj (not important).
Currently with Canary Fronted wont build. DO you want me to rollback frontend changes to develop?

`
Start 'Building UI'
Removing old wwwroot
Installing web dependencies

up to date, audited 725 packages in 2s

114 packages are looking for funding
  run `npm fund` for details

5 vulnerabilities (1 low, 1 moderate, 3 high)

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
Building UI

> kavita-webui@0.7.12.1 prod
> npm run cache-locale-prime && ng build --configuration production && npm run minify-langs && npm run cache-locale


> kavita-webui@0.7.12.1 cache-locale-prime
> node hash-localization-prime.js

Application bundle generation failed. [1.663 seconds]

X [ERROR] Angular compilation initialization failed. [plugin angular-compiler]

  Error: The Angular Compiler requires TypeScript >=5.4.0 and <5.6.0 but 5.7.2 was found instead.
      at checkVersion (file:///C:/Sources/Repos/Kavita/UI/Web/node_modules/@angular/compiler-cli/bundles/chunk-Y3H6JOVH.js:1161:11)
      at verifySupportedTypeScriptVersion (file:///C:/Sources/Repos/Kavita/UI/Web/node_modules/@angular/compiler-cli/bundles/chunk-Y3H6JOVH.js:1165:3)
      at new NgtscProgram (file:///C:/Sources/Repos/Kavita/UI/Web/node_modules/@angular/compiler-cli/bundles/chunk-Y3H6JOVH.js:4347:7)
      at C:\Sources\Repos\Kavita\UI\Web\node_modules\@angular\build\src\tools\angular\compilation\aot-compilation.js:53:88
      at profileSync (C:\Sources\Repos\Kavita\UI\Web\node_modules\@angular\build\src\tools\esbuild\profiling.js:68:16)
      at AotCompilation.initialize (C:\Sources\Repos\Kavita\UI\Web\node_modules\@angular\build\src\tools\angular\compilation\aot-compilation.js:53:60)
      at async initialize (C:\Sources\Repos\Kavita\UI\Web\node_modules\@angular\build\src\tools\angular\compilation\parallel-worker.js:36:50)
      at async C:\Sources\Repos\Kavita\UI\Web\node_modules\@angular\build\node_modules\piscina\dist\worker.js:146:26
`

majora2007:
Okay bare with me. I have no idea how canary got out of sync with develop as I deleted it this morning then recreated it off develop. 

I recreated it off of develop. It looks like there might be a few changes that need to be fixed up from your PR (or cherrypick from before the "canary" merge and remerge). Sorry about all the trouble.

xrishox:
wow this sounds great, i can't wait to test this out! when transcoding from jxl to a supported format, what format does it choose. at first blush it would seem like there would be no reason to not just convert to png since it's lossless, but on second thought that could be really problematic if people are on mobile and bandwidth restricted. on the other hand though encoding a lossy format to another lossy format could start to introduce artifacts. i wonder how you guys chose to address this?

PR #3292, opened 2024-10-20
Feature/people overhaul ux improvements

MrRobotjs:
# Changed
- Changed: UX improvements on people browse page cards
- Changed: UX improvements on the persons detail page
chrome_GF2iolnmBO
image


majora2007:
Don't deal with the merge. I patched over your person card changes into my branch as I like them (minus the border highlight as it's not consistent with other UX). 

For the person detail, I do want to take some of your stuff, but we should hold as users are going to start providing features and feedback once they test and I'll need to code those things (like a user wants some high level stats on this page). 

For now, I'll close the PR out and once feedback starts coming in, we can revisit the css. Feel free to join the nightly channel to discuss it.

PR #3368, opened 2024-11-14
Only show last 6 version updates on admin settings screen

DieselTech:
# Changed
- Changed: Show only the last 6 updates in the `settings#admin-system` screen. This stops the page from trying to load 20+ versions going back well over 2 years.

Honestly even 6 feels like a lot. We could lower it since going back that far brings us to pretty old versions that aren't relevant anymore.


majora2007:
I'd like this implementation to change. Have the API take a int (6) so that the UI can request only the lastest releases. Then refactor all the way down to the VersionService code. 

This way we don't do extra work for that one flow, but we can keep it existing for others. Use a default value of 0 to ignore.

majora2007:
I took care of this in my `feature/random` branch.

PR #3396, opened 2024-11-23
Use docker metadata to generate combos of labels, tags, and images

halkeye:
# Changed
- Changed: Updated docker build scripts to include docker labels

### DIFF
--- a/.github/workflows/develop-workflow.yml
+++ b/.github/workflows/develop-workflow.yml
@@ -128,7 +128,7 @@ jobs:
       - name: Compile dotnet app
         uses: actions/setup-dotnet@v4
         with:
-            dotnet-version: 8.0.x
+          dotnet-version: 8.0.x
 
       - name: Install Swashbuckle CLI
         run: dotnet tool install -g Swashbuckle.AspNetCore.Cli
@@ -137,6 +137,7 @@ jobs:
 
       - name: Login to Docker Hub
         uses: docker/login-action@v3
+        if: ${{ github.repository_owner == 'Kareadita' }}
         with:
           username: ${{ secrets.DOCKER_HUB_USERNAME }}
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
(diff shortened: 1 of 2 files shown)


halkeye:
Tools like renovate and dependabot use docker labels to determine where to look for changelogs
Also a few tweaks so forks can use the github actions to test things (like this PR)

Before (Especially incorrect as its not ubuntu nor 20.04, its kavita):
```
regctl image config --format '{{ jsonPretty .Config.Labels }}' ghcr.io/kareadita/kavita:nightly
{
  "org.opencontainers.image.ref.name": "ubuntu",
  "org.opencontainers.image.version": "20.04"
}
```

After
```
regctl image config --format '{{ jsonPretty .Config.Labels }}' ghcr.io/halkeye/kavita:nightly-0.8.4.0
{
  "org.opencontainers.image.created": "2024-11-23T06:01:02.916Z",
  "org.opencontainers.image.description": "Kavita is a fast, feature rich, cross platform reading server. Built with the goal of being a full solution for all your reading needs. Setup your own server and share your reading collection with your friends and family.",
  "org.opencontainers.image.licenses": "GPL-3.0",
  "org.opencontainers.image.ref.name": "ubuntu",
  "org.opencontainers.image.revision": "7ccfc4b0577bb7a74785740549669d3005831b7a",
  "org.opencontainers.image.source": "https://github.com/halkeye/Kavita",
  "org.opencontainers.image.title": "Kavita",
  "org.opencontainers.image.url": "https://github.com/halkeye/Kavita",
  "org.opencontainers.image.version": "nightly"
}
```

halkeye:
example renovate output with and without the tags
image

majora2007:
I'm not able to grasp, outside the fork changes, what renovate is actually doing or why it's needed. @DieselTech please shine some light as well.

halkeye:
so https://docs.renovatebot.com/modules/datasource/docker/ is the docs for renovate. I'll admit i don't fully understand which labels they use, I do know the ones generated by the action work

Essentially they compare the labels on previous (or latest) and current tag to get the source location (this repo) and the ref (the tag), then it can use the github releases to match release notes and include them in the PR updating an image tag (with a fallback to just diff link if no release note).

I ->believe<- docker hub also will use these container labels to include links to the original repo in the sidebar, but i honestly havn't tested it.

Generally the opencontainer labels (which the docker action adds by default) just is a nice convention to follow and lots of tools can use them.

halkeye:
> if you maintain a Docker image and want Renovate to find your changelogs, add a org.opencontainers.image.source field to your Dockerfile. The link must point to your GitHub or GitLab repository. Here's an example from our renovate/renovate Dockerfile:
> `LABEL org.opencontainers.image.source="https://github.com/renovatebot/renovate"`

majora2007:
Okay, but I'm not sure why we need them on Kavita. I'm not understanding the value add. That's what I'm trying to grasp.

halkeye:
Its like adding og: meta tags to web sites. It allows machines easier time to get metadata. In my case, I use gitops so instead of using :latest, I use a specific tag, and renovate sends me a PR to update the tag. Its nice to have renovate also include the release notes directly in the PR so I don't have to goto this repo and find the release and the release notes.

I would suspect watchtower has something with docker labels too, i saw a dependancy, but i couldn't find where it might actually read it.

Just nice to have for consumers, it shouldn't affect kavita directly at all.

majora2007:
Where is the revision tag being set from? Is this the commit of the PR merge?

halkeye:
I actually don't know exactly what the defaults are. It should be the Sha of the event that triggered the action. So push to master or the releases sha. Etc.

The metadata step can have labels overridden but it's worked out properly for every project I've submitted it to

PR #3442, opened 2024-12-06
Enforce a readable line length for Read More

heyhippari:
# Changed
- Changed: Changed the maximum line length of text inside of expandable blocks of text to a readable value

---

Before:
image

After (Ignore the slight differences in metadata, my prod server has Kavita+ and Komf alimenting it):
image

Review cards:
image

Usually, for longer texts, readability suffers with lines over 75 characters on desktop (On mobile, that's around 50 characters).

To make sure descriptions stay readable, the contents of the read-more component are now capped to a line length of 75 characters on larger breakpoints and 50 characters on smaller breakpoints.

I'm keeping this PR as a draft for now, because I need to adjust the cutoff limits for the various uses of the component. I'm generally trying to keep the number of lines similar to the current one, to not break design. It sacrifices the amount of content that's readable without expanding, but I feel like preserving alignment is much better than having 5+ lines of text shown on load.

Edit: Forgot a source for the length thing: https://baymard.com/blog/line-length-readability

Edit 2: This is now ready for review.

One thing I noticed while adjusting the cutoff, there's often an overflow on the review cards when reviews have line returns. I lowered the cutoff a bit (The max-width doesn't do anything for these, so I can omit this if you prefer), which fixes the overflow in some cases, but there should probably be some server-size cleanup of empty lines and such (Especially since we'll likely want to let clients decide how to display them, including the distance between paragraphs, so returning straight paragraphs separated by one \n seems like the best solution). In any case, that's for another PR some day, but I figured I'd raise it while I'm seeing it.

### DIFF
--- a/UI/Web/src/app/_single-module/review-card/review-card.component.html
+++ b/UI/Web/src/app/_single-module/review-card/review-card.component.html
@@ -16,7 +16,7 @@
             {{review.isExternal ? t('external-review') : t('local-review')}}
           </h6>-->
           <p class="card-text no-images">
-            <app-read-more [text]="(review.isExternal ? review.bodyJustText : review.body) || ''" [maxLength]="150" [showToggle]="false"></app-read-more>
+            <app-read-more [text]="(review.isExternal ? review.bodyJustText : review.body) || ''" [maxLength]="140" [showToggle]="false"></app-read-more>
           </p>
         </div>
       </div>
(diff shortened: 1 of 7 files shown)


majora2007:
I don't disagree with you. I had increased this as people with 4K monitors were complaining about not eating enough space.

majora2007:
This broke the read-more in changelog component. can you please raise a fix for it @heyhippari ?

image

majora2007:
I ended up reworking the UX here, but overall we need an ability to conditionally apply your new css so it can work as it was for different situations.

PR #3477, opened 2024-12-30
Fixed: [OPDS] Recreate a pdf chapter's cache if images are missing (This happens when the pdf chapter is accessed first using the web reader)

m-amazirh:
# Fixed
- Fixed: [OPDS] Recreate a pdf chapter's cache if images are missing

## More details

When a user acccesses a PDF Chapter for the first time using the web reader, a corresponding cache directory is created containing only the pdf file. The pages are not extracted as png images into the cache directory.

If the user then requests a page from the same PDF Chapter using the OPDS controller/api, Kavita returns the error `No such image for page {0}. Try refreshing to allow re-cache.` .

This happens because the OPDS controller expects the chapter's pages (png images) to already exist in the cache directory. It does not attempt to extract them from the PDF file if the directory exists.

This code change fixes the issues by recreating the chapter's cache directory if chapter's cache directory exists and the opds contoller doesn't find in it the requested page by the user.

(note : I'm not sure if this is the best way to handle this issue and I didn't look into how the fix behaves if the user tries to requests multiple pages in parallel)

## Steps to reproduce

1. Add a new PDF to the library
2. Access at least once the pdf/chapter using the web reader.
    - The web readers calls `ReaderController#GetPdf` which calls `CacheService#Ensure` with `extractPdf=false`. A chapter directory is created containing the pdf files without extracting the pages as images.
3. Next, access the same pdf/chapter using the opds api (ex : ` http://localhost:5000/api/opds/81235fbc-e77c-4ad5-b789-a0967beb3b6a/image?libraryId=1&seriesId=1&volumeId=1&chapterId=1&pageNumber=50`)
    - The OPDS api calls `OPDSController#GetPageStreamedImage` which calls `CacheService#Ensure` with `extractPdf=true`
    - But the call to `CacheService#Ensure` has no effect [since the cache directory already exists](https://github.com/Kareadita/Kavita/blob/6ae9cbf4aa9be398d096547e1e024198c1df6bec/API/Services/CacheService.cs#L176)
    - Result : the pages are not extracted in the cache directory.
    - Then the code tries to load the page from the cache directory but does not find it and returns an error `No such image for page {0}. Try refreshing to allow re-cache.`

### DIFF
--- a/API/Controllers/OPDSController.cs
+++ b/API/Controllers/OPDSController.cs
@@ -1237,7 +1237,7 @@ public class OpdsController : BaseApiController
 
         try
         {
-            var path = _cacheService.GetCachedPagePath(chapter.Id, pageNumber);
+            var path = await GetPage(chapterId, pageNumber);
             if (string.IsNullOrEmpty(path) || !System.IO.File.Exists(path))
                 return BadRequest(await _localizationService.Translate(userId, "no-image-for-page", pageNumber));
 
@@ -1359,5 +1359,17 @@ public class OpdsController : BaseApiController
         _xmlSerializer.Serialize(sm, feed);
 
         return sm.ToString().Replace("utf-16", "utf-8"); // Chunky cannot accept UTF-16 feeds
+    }
+
+    private async Task<String> GetPage(int chapterId, int pageNumber)
+    {
+        var path = _cacheService.GetCachedPagePath(chapterId, pageNumber);
+        if (string.IsNullOrEmpty(path) || !System.IO.File.Exists(path))
+        {
+            _cacheService.CleanupChapters([chapterId]);
+            await _cacheService.Ensure(chapterId, true);
+        }
+
+        return _cacheService.GetCachedPagePath(chapterId, pageNumber);
     }
 }


majora2007:
Just got back from Holiday and have some inflight work. Once I wrap that, I'll take a look at this. 

You are correct, this is 100% an oversight between the main reader and the way OPDS implementation works. I think there may be a better way to handle this, I'll get back to you once I look over the PR and the code.

m-amazirh:
Thank you

majora2007:
I ended up implementing this a different way. I updated the Ensure call so that if extractPdfToImages was true (which is basically only for the OPDS flow), then we explicitly check if there are images in the cache directory.

```cs
if (_directoryService.Exists(extractPath))
            {
                if (extractPdfToImages)
                {
                    var pdfImages = _directoryService.GetFiles(extractPath,
                        Tasks.Scanner.Parser.Parser.ImageFileExtensions);
                    if (pdfImages.Any())
                    {
                        return chapter;
                    }
                }
                else
                {
                    return chapter;
                }
            }
```

Thanks for bringing this issue up and proposing a fix. It makes it much easier for me to drive it home than just an issue.

PR #3496, opened 2025-01-08
fix(i18n): equals is not equitable in the filter context

tobidope:
# Fixed
- Fixed: Fixed german translation of equal in the filter context


majora2007:
Localization changes belong in Weblate and not in Github. 

https://hosted.weblate.org/projects/kavita/-/de/

PR #3576, opened 2025-03-01
Fix/3557 scanner subfolder always scanning

### LINKED ISSUE #3557: Scanner not being smart when a series has subfolders
### What happened?

The scanner is pretty smarts, and doesn't scan series if the OS says nothing has changed. However, I've noticed that this doesn't seem to be working for a good subsection of my series. Upon looking at my logs, these are all series that have directories inside of them. 

Take the following layout
```
Root
    - Series A
        - Series A Vol. 1.cbz
        - Series A Vol. 1.cbz
    - Series B
        - Series B Volume 1
            - Series B Ch. 1.cbz
            - Series B Ch. 2.cbz
        - Series B Volume 2
            - Series B Ch. 3.cbz
            - Series B Ch. 4.cbz 
        - Series B Ch. 5.cbz # Not yet assigned to a volume by a publisher 
```

The scanner will correctly skip Series A, but Series B will be scanned each scan regardless of any changes were made to it.

### What did you expect?

I expect the "last change" check to be done each directory inside a series, if there are any. 
Looking at the logs, it does know of the inner directories as it says it will scan/check those, and not the top one. It should just check them as well for last changed.

Nightly Testing Branch

Docker (Dockerhub Container)

None

_No response_

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

https://discord.com/channels/821879810934439936/851515194185744464/1341167146712961146

Fesaa:
# Fixed
- Fixed: Scanner scanning subfolders of a series when no changes have been made (Fixes #3557)

Still work todo, PR is easier to track

Copied from discord:
If one subfolder changes, all are scanned. This is important otherwise the other subfolders content are removed. Should be a fine trade-off. 

This should be safe to do here, as if it matches a wrong series, we know this series is its parent and thus will always have changes. 

I have tested this on Windows & MacOS. Haven't had the chance to build for my server (Linux) yet. That said, I have no idea how the Comic stuff works etc. So... I've only tested with Manga as that's what I had laying around in the right layout to just copy over. 

TODO:

- [x] Unit Tests


API/Services/Tasks/Scanner/ParseScannedFiles.cs
```diff
@@ -222,6 +227,21 @@ private bool HasSeriesFolderNotChangedSinceLastScan(IDictionary<string, IList<Se
...
+        {
+            return null;
+        }
+
+        if (seriesPaths.TryGetValue(directory, out var seriesList))
+        {
+            return seriesList;
+        }
+
+        return TryGetSeriesList(seriesPaths, _directoryService.GetParentDirectoryName(directory));
```
majora2007: You need another short circuit if directory is the library root so we don't go out of bounds in some weird case.

Fesaa: This case will be hit for every new series I'd image 😅

API.Tests/Services/ParseScannedFilesTests.cs
```diff
@@ -349,4 +357,101 @@ public async Task ProcessFiles_ShouldCallFolderActionOnce()
...
+    public async Task HasSeriesFolderNotChangedSinceLastScan_AllSeriesFoldersHaveChanges()
+    {
+        const string testcase = "Subfolders always scanning all series changes - Manga.json";
+        var infos = new Dictionary<string, ComicInfo>();
+        var library = await _scannerHelper.GenerateScannerData(testcase, infos);
+
+        var testDirectoryPath = Path.Join(
+            Path.Join(Directory.GetCurrentDirectory(), "../../../Services/Test Data/ScannerService/ScanTests"),
+            testcase.Replace(".json", string.Empty));
+        testDirectoryPath = Path.GetFullPath(testDirectoryPath);
```
majora2007: Instead of applying this here, instead apply it on the return of ScannerHelper.GenerateTestDirectory() so that you can remove this library.Folders hack and all other tests will work for you.

API.Tests/Services/ParseScannedFilesTests.cs
```diff
@@ -349,4 +357,101 @@ public async Task ProcessFiles_ShouldCallFolderActionOnce()
...
+
+        var fs = new FileSystem();
+        var ds = new DirectoryService(Substitute.For<ILogger<DirectoryService>>(), fs);
+        var scanner = _scannerHelper.CreateServices(ds, fs);
+        await scanner.ScanLibrary(library.Id);
+        var postLib = await _unitOfWork.LibraryRepository.GetLibraryForIdAsync(library.Id, LibraryIncludes.Series);
+
+        Assert.NotNull(postLib);
+        Assert.Equal(4, postLib.Series.Count);
+        var spiceAndWolf = postLib.Series.First(x => x.Name == "Spice and Wolf");
```
majora2007, line 384: This block of code is hard to parse through. Break up each fetch/assert by a new line.

API.Tests/Services/ParseScannedFilesTests.cs
```diff
@@ -349,4 +357,101 @@ public async Task ProcessFiles_ShouldCallFolderActionOnce()
...
+        Assert.Equal(2, spiceAndWolf.Volumes.Count);
+        var frieren = postLib.Series.First(x => x.Name == "Frieren - Beyond Journey's End");
+        Assert.Single(frieren.Volumes);
+        var executionerAndHerWayOfLife = postLib.Series.First(x => x.Name == "The Executioner and Her Way of Life");
+        Assert.Equal(2, executionerAndHerWayOfLife.Volumes.Count);
+
+        Thread.Sleep(1100); // Ensure at least one second has passed since library scan
+
+        var executionerCopyDir =
```
majora2007: I would add a comment to explain what this is doing and validating against. Something like:

// Add a new file to X directory and scan, validate that only Y and Z are showing as having had changes.

API.Tests/Services/ParseScannedFilesTests.cs
```diff
@@ -349,4 +357,101 @@ public async Task ProcessFiles_ShouldCallFolderActionOnce()
...
+            new MockReadingItemService(ds, Substitute.For<IBookService>()), Substitute.For<IEventHub>());
+
+        var folderMap = await _unitOfWork.SeriesRepository.GetFolderPathMap(postLib.Id);
+        Assert.Equal(6, folderMap.Count);
+
+        var res = await psf.ScanFiles(testDirectoryPath, true, folderMap, postLib);
+        var changes = res.Where(sc => sc.HasChanged).ToList();
+        Assert.Equal(2, changes.Count);
+        // Only volumes of The Executioner and Her Way of Life should be marked as HasChanged (Spice and Wolf also has 2 volumes dirs)
+        Assert.Equal(2, changes.Count(sc => sc.Folder.Contains("The Executioner and Her Way of Life")));
```
majora2007, line 410: You need to run the proper scanner as well to validate that with your changes, the nothing is removed and that the changes are properly ingested. This can be in ScannerServiceTests with the same json file.

API/Services/Tasks/Scanner/ParseScannedFiles.cs
```diff
@@ -721,7 +752,8 @@ private async Task ParseFiles(ScanResult result, IDictionary<string, IList<Serie
         // If folder hasn't changed, generate fake ParserInfos
         if (!result.HasChanged)
         {
-            result.ParserInfos = seriesPaths[normalizedFolder]
+            // We are certain TryGetSeriesList will return a valid result here, if the series wasn't present yet. It will have been changed.
```
majora2007, line 755: I don't really understand this comment, can you clarify it a bit?

API.Tests/Services/ParseScannedFilesTests.cs
Fesaa: I've done this for the test with sub folders. Does the publisher layout also need this test or not?

API/Services/Tasks/Scanner/ParseScannedFiles.cs
Fesaa: Since I'm forcing the result of the new method to be not null, I've added the comment to explain this is safe to do. And will not throw an exception.

As all series that would return null, will have HasChanged set to true

API.Tests/Services/ParseScannedFilesTests.cs
majora2007: No. Subfolders is the publisher layout.

PR #3629, opened 2025-03-14
Fix Subfolders Scanner Issue

Fesaa:
# Fixed
- Fixed: Fixed a bug where a series with files at its root, and in subfolders would have the chapters in the subfolders removed after a scan (develop)

### DIFF
--- a/API.Tests/Services/ScannerServiceTests.cs
+++ b/API.Tests/Services/ScannerServiceTests.cs
@@ -55,6 +55,28 @@ public class ScannerServiceTests : AbstractDbTest
         await _context.SaveChangesAsync();
     }
 
+
+    protected async Task SetAllSeriesLastScannedInThePast(Library library, TimeSpan? duration = null)
+    {
+        foreach (var series in library.Series)
+        {
+            await SetLastScannedInThePast(series, duration, false);
+        }
+        await _context.SaveChangesAsync();
+    }
+
+    protected async Task SetLastScannedInThePast(Series series, TimeSpan? duration = null, bool save = true)
+    {
+        duration ??= TimeSpan.FromMinutes(2);
+        series.LastFolderScanned = DateTime.Now.Subtract(duration.Value);
+        _context.Series.Update(series);
+
+        if (save)
+        {
+            await _context.SaveChangesAsync();
+        }
+    }
+
     [Fact]
     public async Task ScanLibrary_ComicVine_PublisherFolder()
     {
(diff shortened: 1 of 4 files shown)


API.Tests/Services/ScannerServiceTests.cs
```diff
@@ -822,4 +822,66 @@ public async Task SubFolders_NoRemovals_ChangesFound()
...
+
+        postLib = await _unitOfWork.LibraryRepository.GetLibraryForIdAsync(library.Id, LibraryIncludes.Series);
+        Assert.NotNull(postLib);
+        Assert.Single(postLib.Series);
+
+        spiceAndWolf = postLib.Series.First(x => x.Name == "Spice and Wolf");
+        Assert.Equal(4, spiceAndWolf.Volumes.Count);
+        Assert.Equal(5, spiceAndWolf.Volumes.Sum(v => v.Chapters.Count));
+
+        spiceAndWolf.LastFolderScanned = DateTime.Now.Subtract(TimeSpan.FromMinutes(2));
```
majora2007: We should probably refactor this into a helper method on this test class as we will be using it a lot.

API.Tests/Services/ScannerServiceTests.cs
```diff
@@ -822,4 +822,66 @@ public async Task SubFolders_NoRemovals_ChangesFound()
...
+
+        var spiceAndWolf = postLib.Series.First(x => x.Name == "Spice and Wolf");
+        Assert.Equal(3, spiceAndWolf.Volumes.Count);
+        Assert.Equal(4, spiceAndWolf.Volumes.Sum(v => v.Chapters.Count));
+
+        spiceAndWolf.LastFolderScanned = DateTime.Now.Subtract(TimeSpan.FromMinutes(2));
+        _context.Series.Update(spiceAndWolf);
+        await _context.SaveChangesAsync();
+
+        // Add file at series root
```
majora2007: This comment doesn't make sense, it's not adding a file at root, but adding a file in the Spice and Wolf Subfolder

majora2007 (commented):
Overall looks good at first glance. I'll pull it down today as well.

API.Tests/Services/ScannerServiceTests.cs
Fesaa: Right, root is probably the wrong name for it, it's supposed to explain adding a file in the series. But not in a subfolder of the series. So the "root" of the series, but not the root of the library.

majora2007: I'd write this as "Add a new Volume to Spice and Wolf series folder". This indicates what it's doing and implies that the volume count should change.

API.Tests/Services/ScannerServiceTests.cs
Fesaa: I'll get this done, and replace the other sleeps I had laying around from before

majora2007:
After merging this into my branch, multiple unit tests are failing.

PR #3644, opened 2025-03-16
Fix not being able to add new series a library in publisher layout

Fesaa:
# Fixed
- Fixed: Fixed an issue where you were unable to add new series if your library was in publisher layout

Right, so LowestFolderPath and FolderPath are generally the same, except in the following cases
- You're using the publisher layout
- Your series is using subfolders, but only one and has no files on the same level as subfolders (see commented out test)

If I forgot one, please let me know so I can test...

### First case
The FolderPath has no real meaning to the series, and can be safely ignored.  Including it caused the bug, as new series would match with an old one. Which has its LowestFolderPath not changed.

### Second case
The unit test I've commented out is broken with this fix and without, just a difference in why
**Before**
The new chapter picks up the series, but doesn't scan because LowestFolderPath doesn't have changes
**Now**
The chapter is picked up, and later assigned to the series. But the older chapter got skipped because it's LowestFolderPath didn't change. 

Before the subfolder changes, this would work as subfolders simply always got scanned. I'm really unsure how to fix this.
You can force scan the series to have it pick up all; after which LowestFolderPath is raised, and new scans will work as normal.

### Proper fix?
What we would really need is the HighestUniquePath containing only this series. And this should be used to construct the seriesPath dictionary
In a way that;
- Spice and Wolf/Spice and Wolf Vol. 1 => Spice and Wolf
- Publisher/Series => Series
- Publisher/Series/Series Vol. 1 => Series

Then I think the only "issue" you'd have is that when adding a new series to a Publisher layout with only one series both get scanned, and have their HighestUniquePath updated.

### DIFF
--- a/API.Tests/Services/ParseScannedFilesTests.cs
+++ b/API.Tests/Services/ParseScannedFilesTests.cs
@@ -399,9 +399,9 @@ public class ParseScannedFilesTests : AbstractDbTest
         File.Copy(Path.Join(executionerCopyDir, "The Executioner and Her Way of Life Vol. 1 Ch. 0001.cbz"),
             Path.Join(executionerCopyDir, "The Executioner and Her Way of Life Vol. 1 Ch. 0002.cbz"));
 
-        // 4 series, of which 2 have volumes as directories
+        // 4 series
         var folderMap = await _unitOfWork.SeriesRepository.GetFolderPathMap(postLib.Id);
-        Assert.Equal(6, folderMap.Count);
+        Assert.Equal(4, folderMap.Count);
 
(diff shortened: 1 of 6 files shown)


majora2007:
HighestUniquePath  is what series FolderPath is essentially. I originally added FolderPath but with the rewrite, moved over to LowestFolderPath. 

Let me take a closer look, but I think I'd like to undo the changes made around the subfolder optimization, release the hotfix, then you and I can work together and plan out some larger refactors. I have some light plans around changes and proof of concepts to see if I can get more performance with the same flexibility of the current scanner.

PR #3647, opened 2025-03-17
[skip ci] Fix unit tests on macOS (And probably Unix)

Fesaa:
This PR fixes all but 7 tests on macOS. The remaining tests are in DirectoryServiceTest

image

Top one fails because `\` is valid in a filename. And the othesr fails because `C:/` is hardcoded, I could change them to construct it from Root +, but the test already has them mixed; so I think that would  change the test too much. 
Unsure what to do with them.

# Fixed
- Fixed several tests not running on macOS/Unix because of `C:/` as root
- Fixed a test involving time not passing on timezones > UTC because of how the setter works in the MockFileData

# Changed
- Unified several test classes to use the new AbstractFsTest to simplify setup
- Skip the IsDriveMounted tests on macOS/Unix as the root `/` is always mounted, and `C:/` will throw
- Change some tests to test against the CurrentLocale instead as the result is user facing

### Notes
I've had a friend run the tests on their Windows PC, and everything passes for them. However, they're on the same timezone as me. So would probably need to have someone tests whose on a timezone "earlier" than UTC.

### DIFF
--- a/API.Tests/Services/DirectoryServiceTests.cs
+++ b/API.Tests/Services/DirectoryServiceTests.cs
@@ -4,6 +4,7 @@ using System.Globalization;
 using System.IO;
 using System.IO.Abstractions.TestingHelpers;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading.Tasks;
 using API.Services;
@@ -14,7 +15,7 @@ using Xunit;
 
 namespace API.Tests.Services;
 
-public class DirectoryServiceTests
+public class DirectoryServiceTests: AbstractFsTest
 {
     private readonly ILogger<DirectoryService> _logger = Substitute.For<ILogger<DirectoryService>>();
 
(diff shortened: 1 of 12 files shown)


majora2007 (approved):
Looks good. I'll merge this after i do the hotfix.

majora2007:
Actually builds on GA are failing now, we need to address that.

Fesaa:
~~This should fix it, but I'm unsure why~~

I'll be home again tomorrow, and will fix them then as I don't have a Windows PC here.

API.Tests/Services/DirectoryServiceTests.cs
```diff
@@ -373,9 +375,15 @@ public void CopyDirectoryToDirectory_ShouldCopyAllFileAndNestedDirectoriesOver()
     #endregion
 
     #region IsDriveMounted
+    // The root directory (/) is always mounted on non windows
     [Fact]
     public void IsDriveMounted_DriveIsNotMounted()
     {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return;
```
majora2007, line 392: Let's put a logger to indicate this test is not running.

majora2007 (approved):
I'm fine to merge then add the logging statement afterwards, or you can do it (on all applicable tests) and I will merge.

PR #3664, opened 2025-03-22
Version Update Modal Rework + A few bugfixes

### LINKED ISSUE #3418: Weird parsing bug
### What happened?

I have an issue with specials naming after updating kavita (i don't know which version reveal this issue)
I used to name my specials in my folders
```
SP01 1. nameOne
SP02 2. nameTwo
```

that way my specials had name
```
1. nameOne
2. nameTwo
```

now i just have
```
1
2
```

etc, i do not have any name in my specials.
it seems that the dot . brakes naming then it can't read the name after the dot
If i remove it, or replace with"-" or anything else, then i can see the full name.

### What did you expect?

naming specials that way
```
SP01 1. nameOne
SP02 2. nameTwo
```

should have this result
```
1. nameOne
2. nameTwo
```

0.8.4 - Stable

Linux

None

_No response_

None

_No response_

### Relevant log output

_No response_

### Additional Notes

_No response_

### LINKED ISSUE #3613: Image named as "back cover" being chosen as cover image
### What happened?

Ran into a couple of archives with "back cover" in a filename causing it to be selected as the cover image.  If possible, could it be added to the exception list?  I see back-cover, back_cover, and backcover are ... covered.

### What did you expect?

Not to have the back cover image picked as the cover, though understand there are only so many special cases you might want to hardcode in.

Nightly Testing Branch

Docker (Dockerhub Container)

Windows

Chrome

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

### LINKED ISSUE #3657: Version Updater (UI) Overhaul
Overhaul the UI Version checking code. 

There are 3 different types of notifications for updates:
- Client Refresh - The Server has updated, the client (javascript) needs to be refreshed
- New Update - There is an update for the server owner (nightly/stable) that can be updated to
- Out of Band - You are super out of date, please update

For each of these message types, we need to be respectful to the user experience, so the following popup notification schedule will be used:
- Client Refresh - Pop once (refresh is required)
- New Update - Pop once a week
- Out of Band - Pop once a month

Caveats:
- Need to handle how to show this to users when they never got the version marker initially (perhaps put in login flow)
- Don't pop any of these modals during reading sessions

Design:
- To facilitate tracking of data, we will use localstorage.

majora2007:
# Changed
- Changed: Reworked the version update logic so that we don't pop the new version update so often. It will now pop once and if no action is taken, it will wait a week to pop again. The same logic applies to the you are X releases behind, etc (but longer windows).  (Closes #3657)
- Changed: Caught a few more Invariant parse issues (develop)
- Changed: Updated dependencies and Security issues from downstream libraries
- Changed: Enhanced parsing code for (back cover) falsely being recognized as a cover image. (Closes #3613)
- Changed: Cleaned up how specials are parsed to allow supporting periods in the Title name (Closes #3418 )

# Fixed
- Fixed: Fixed a bug where Top Readers on Stats page wasn't rendering the numbers correctly


API.Tests/Parsers/BasicParserTests.cs
```diff
@@ -162,7 +162,25 @@ public void Parse_MangaLibrary_SpecialInFilename2()
...
+        Assert.True(actual.IsSpecial);
+    }
+
+    /// <summary>
+    /// Tests that when the filename parses as a special, it appropriately parses
+    /// </summary>
+    [Fact]
+    public void Parse_MangaLibrary_SpecialInFilename_StrangeNaming()
+    {
+        var actual = _parser.Parse("M:/My Dress-Up Darling/SP01 1. Special Name.cbz",
```
Fesaa: I'm not sure if it'll fail, but safer to extend AbstractFsTest and use the Root field here instead of M:/

API.Tests/Services/SettingsServiceTests.cs
```diff
@@ -0,0 +1,293 @@
...
+            EnableLocalizedName = true,
+            EnablePublicationStatus = true,
+            EnableRelationships = true,
+            EnablePeople = true,
+            EnableStartDate = true,
+            EnableGenres = true,
+            EnableTags = true,
+            FirstLastPeopleNaming = true,
+            EnableCoverImage = true,
+            AgeRatingMappings = new Dictionary<string, AgeRating> { { "Adult", AgeRating.R18Plus } },
```
Fesaa, line 79: I saw you were changing list constructions to [] in the other commits, would also be possible here, and generally throughout this file.

Fesaa (commented):
There are also a bunch of console logs left in the UI code for version popup. 

And a bunch of rather useless comments in the new tests, by the AI, I think haha.

PR #3686, opened 2025-03-29
Fix first locale request producing response

nevivurn:
# Fixed
- Fixed: Fix first locale request empty response


majora2007:
Nice catch, going to merge into my branch which has a boat load of bugfixes so we get one nice release for nightly users.

PR #3713, opened 2025-04-08
bugfix - remove hard 140px height limit for reading list items

ryanfiller:
There was a bit of an issue with text overflowing the bounds of reading list items, this should fix it.

before/after showing the 14px height limit was causing an overflow


majora2007:
I'll take a look, if I recall, the fixed height is for the infinite scroller when you hit over 100 items in the list.

majora2007:
I took a look and the experience worked better but not a full blown solution. I ended up increasing the amount of characters for the read more (since it would never show the full description regardless). Thanks for the PR.

PR #3762, opened 2025-04-21
[skip ci] Fix color scape test

goremykin:
# Changed
- Changed: Removed System.Drawing.Common dependency

# Fixed
- Fixed: Fixed color scape test

I noticed that we reference System.Drawing.Common (that is not cross platform) just for one test which does not work correctly. So this PR fixes the test and gets rid of the System.Drawing.Common.

The test generates a html file with posters and color. Before this change it was generating black rectangles for some reason:
Screenshot 2025-04-21 at 13-27-10 Color Scape Comparison

After the fix:
Screenshot 2025-04-21 at 13-35-43 Color Scape Comparison

To be honest, this test is not very useful. If we suddenly start generating wrong colors, it will not fail and we will not know about the problem. Should I rewrite it to validate results colors instead of generating html?


majora2007:
The test here is not very useful because it was just there for the initial implementation and in case I made major changes to it. 

How would you propose validating the results of the colors? The idea with this test is more-so to validate that the colors chosen will result in a good ColorScape (and since the rest of the logic is on the FE, not trivial).

majora2007 (approved):

goremykin:
@majora2007, If the purpose of this test is to manually verify that colors look good in cases of significant changes, then it is a useful test, sorry.

majora2007:
I'll take a look at this PR soon, a good catch on the System.Drawing.Common reference. For your other PRs, I will probably just roll them into one or manually remove the references.

goremykin:
@majora2007, No problem, do it the way that is convenient for you. Thanks!

majora2007 (approved):

majora2007:
FYI, I tested this after removing the System.Common.Drawing dependency and it broke the build. This is because MarkdownDeep requires it to render out markdown.

goremykin:
> FYI, I tested this after removing the System.Common.Drawing dependency and it broke the build. This is because MarkdownDeep requires it to render out markdown.

That is unfortunate :(
I'll check it out in my free time, thanks.

PR #3782, opened 2025-05-02
Converted majority PX values to REM

MrRobotjs:
# Changed
- Changed: Changed PX to REM in .SCSS and .HTML files except in `dark.scss` and values inside `calc()` this way everything is based off the 16px html font size

### DIFF
--- a/UI/Web/src/_card-item-common.scss
+++ b/UI/Web/src/_card-item-common.scss
@@ -1,27 +1,27 @@
-$image-height: 232.91px;
-$image-width: 160px;
+$image-height: 14.556875rem;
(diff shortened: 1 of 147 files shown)


majora2007:
What's the overall goal of this PR? While there is some marginal benefit for users that scroll, I'm not sure I get why you put in the work to change over to rem.

majora2007:
Just wanted to follow up on this. At first, I didn't understand the need to perform this, however I do now. I ended up implementing this in #4464. Thanks for bringing this as a suitable enhancement to my attention. For next time, please add context to your opening PR. Thanks.

Issue #3800, opened 2025-05-13
400 Error Could not find the appropriate html for that page when using manga stitched together by Calibre - SingleOrDefault()


pilotMike:
### What happened?

I took a look at what was happening, since Calibre can open this file just fine. I started working on adjusting the source code, but I'll need more conversation and context to fix it.

What's happening is that when I advance past the first page, I get the above error. I saw that it was in `BookService.GetKeyForImage`, so I started making a unit test to see if I could make a best attempt at a fix.

What I see is that in this file there are 227 images and they look like:
imageKey: cover.jpeg
keys:
* cover.jpg
* 1/cover.jpeg ... all the way to 8/cover.jpeg (not in order, there are other images in between)
* 1/images/00001.jpg
* 1/images/00002.jpg, etc
(Notice that the attribute has _jpeg_ and not _jpg_, otherwise the cover would have been found, but this is page 1 (technically 2), so I'm not sure why this particular file has 2 covers at the start. It seems like manga that are stitched together could be an edge-case.

The same thing happens with **stylesheets.css** as well.

This is causing this code to fail on `SingleOrDefault()`. I started making an attempt to try and choose the best one, but I realized that I don't have enough context. I'm also not sure if

### Solutions
I don't stitch the manga together and keep it as chapters 😆 

I can provide a link to the stitched file. This one is ~50MB.
I believe that I did the stitching conversion in the MOBI format, possibly.

## More Info
Below are all of the keys in this particular file. The initial numbers, "{n}/images/{...}", correspond to the chapter, as this has 8 chapters. What looks like is happening is on the second one in the list, we do a SingleOrDefault for anything that ends with `cover.jpeg`, so we find ones from other chapters. In this case, passing in `page` variable could be used to find it. It looks like the initial cover request would be page 0, which wouldn't correspond to any top level directory here, and then `1` would start matching.

I'm thinking that this search could be done as an approach with the "ignore the front part" as a fallback. If this sounds like a good idea, I could contribute this bit of code.

All of the keys in the document:
cover.jpg
1/cover.jpeg
1/images/00001.jpg
... (continues in the same pattern up to 33)
1/images/00033.jpg
2/cover.jpeg
2/images/00001.jpg
...
2/images/00025.jpg
3/cover.jpeg
3/images/00001.jpg
...
3/images/00027.jpg
4/cover.jpeg
4/images/00001.jpg
...
4/images/00027.jpg
5/cover.jpeg
5/images/00001.jpg
...
[...]

PR #3802, opened 2025-05-13
[Feat] Adds support for multiple chapters.

pilotMike:
Adds support for multiple chapters. The previous version ignored any leading chapter numbers in the uri of {chapterId}/images/{image}. It now accounts for that and is working in a locally run user interface for an epub that was broken. I didn't check in the unit test over the GetBookPage method since it requires a specific file. 

I can provide a link to the specific epub that wasn't working - it's 50MB though.

References: https://github.com/Kareadita/Kavita/issues/3800

# Added
- Multi-chapter support that can show up in merged epubs. Chapter URIs were previously ignoring everything except the last end of an image or stylesheet uri, which could result in multiple matches and fail. It now works.

# Changed
- BookService in getting image keys and stylesheet keys. The `page` variable is now getting passed down and included in the key uri logic.
- BookController to also use chapter number in generating a key

# Fixed
- Fixed: Fixed a bug (Fixes #issue 3800)

## Not included unit test
```csharp
// Test included as example. I can provide a link to the file for further testing.
    /// <summary>
    /// Scenario: a manga volume that was created by stitching together the chapters in Calibre has 2 cover files:
    /// * cover.jpg
    /// * 1/cover.jpeg
    /// * 1/images/00001.jpg
    ///
    /// With the image file determined as cover.jpeg (note the 'e').
    /// As a result, it was not found with single or default
    /// </summary>
    [Theory]
    [MemberData(nameof(ChapterLists))]
    public async Task ShouldParseSecondPageComicStitchedTogetherInCalibre(int page)
    {
        var bookPath = Path.Combine("Data", "The Color of the End - Haruo Iwamune.epub");

        var page2 = await _bookService.GetBookPage(page, 0,  bookPath, "");
    }

    public static IEnumerable<object[]> ChapterLists() => Enumerable.Range(0, 9).Select(i => new object[] { i });
```

Previous Result:
image

Current Result:

https://github.com/user-attachments/assets/3059bd43-8031-4c64-a51f-fdfaa9d8180b

### DIFF
--- a/API/Services/BookService.cs
+++ b/API/Services/BookService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -321,7 +322,7 @@ public class BookService : IBookService
         }
     }
 
-    private static void ScopeImages(HtmlDocument doc, EpubBookRef book, string apiBase)
+    private static void ScopeImages(HtmlDocument doc, EpubBookRef book, string apiBase, int page)
     {
         var images = doc.DocumentNode.SelectNodes("//img")
                      ?? doc.DocumentNode.SelectNodes("//image") ?? doc.DocumentNode.SelectNodes("//svg");
@@ -345,7 +346,7 @@ public class BookService : IBookService
 
             if (string.IsNullOrEmpty(key)) continue;
 
-            var imageFile = GetKeyForImage(book, image.Attributes[key].Value);
+            var imageFile = GetKeyForImage(book, image.Attributes[key].Value, page);
             image.Attributes.Remove(key);
 
             if (!imageFile.StartsWith("http"))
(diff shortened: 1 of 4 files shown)


API/Services/BookService.cs
```diff
@@ -395,6 +395,68 @@ private static string GetKeyForImage(EpubBookRef book, string imageFile)
...
+
+        if (fullMatch != null) return fullMatch;
+
+
+        return cssLocal.Select(x => x.Key).SingleOrDefault(k => k.GetLastSegmentSpan().SequenceEqual(imageFile));
+    }
+
+    private static string? GetCorrectedKeyOEBPS(IReadOnlyCollection<EpubLocalByteContentFileRef> images, int page, string imageFile)
+    {
+        imageFile = imageFile.Replace("..", string.Empty);
```
pilotMike, line 453: previously the string.Replace was happening repeatedly on every instance in the collection, now it happens once

pilotMike:
@majora2007 tagging in case this is something that you'd want to incorporate :)

majora2007:
User sent file: The Color of the End - Haruo Iwamune.epub for testing this issue

majora2007:
@pilotMike I was trying to find the file you sent me. Can you resend it to me so I can try to understand what this is trying to fix?

pilotMike:
@majora2007 Here is a burn after reading bin link to the file. Let me know if I need to resend another https://paste.michaeldileo.org/?6bc3d13ab21fd428#-Dy61zGXfnk4ocHYCEe5vMYoDRk6vpkfmaPTfX3gbUWC3

majora2007:
Thanks

majora2007:
I looked at the epub and it's not valid. It is essentially an epub with multiple folders within, each folder being a self contained epub. That's not part of the standard and seems to just be hacked together. 

How did you come across this book? Are you creating it to keep multiple chapters into a series (you would probably be better off taking each folder and just zipping into a chapter.epub)

pilotMike:
Its been a while since I did it, but iirc, each chapter was available individually of the manga and the "book" you'd buy had a certain number of chapters. I downloaded each chapter of the book and imported it into Calibre to combine/merge into a single epub. This file was how calibre did the output. For what it's worth, other e-reader software is able to open this file, such as the e-reader on OSX (and calibre of course).

PR #3835, opened 2025-06-01
Browse by Genre/Tag/Person with new metadata system for People

### LINKED ISSUE #3850: Import menu missing for admin user
### What happened?

I have one user, who is an admin, and when I go to settings I don't see the Import / CBL Reading List menu listed. I created a new user with the default login checked, and with that user I was able to see the Import menu as expected. I changed that user to admin, logged in again and I did not see the Import menu. I changed the user back to just login and I was able to see the import menu again.

### What did you expect?

The Import / CBL Reading List menu item should be available to my admin user.

Nightly Testing Branch

Docker (Dockerhub Container)

None

Chrome

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

I saw the issue before and after the v0.8.6.14 update.  Prior to the update I saw this error in the console:
setting-switch.component.ts:43 ERROR TypeError: crypto.randomUUID is not a function

After the update that went away, but the settings issue was still there.

Running on Unraid docker.

Discussed on discord and requested a bug be created.

majora2007:
This branch consists of a bunch of fixes and mainly me going off and doing whatever I was inspired to do after my holiday. This adds some nice new browse pages and expands the metadata system to People entities, paving the way to expand it in the future. 

I have also included feedback and bugs from the Reading Profile from Amelia. 

image

image

image

# Added
- Added: Added the ability to browse all Genres and Tags in the Server (Closes FR #3173)
- Added: Updated Matched Metadata to have a library type filter, given that a user might want to do just Comics or Manga at a time.
- Added: Added a page to browse all Genres. Currently accessible from user menu until I find a good linking location.
- Added: Added a page to browse all Tags.  Currently accessible from user menu until I find a good linking location.
- Added: Added the ability to perform metadata filtering on Browse People (prev Browse Authors) page. Can filter by Role, Name, Series Count, and Chapter Count. Can Sort on Role, Name, Series Count, and Chapter Count. 
- Added: Added a setting to disable the width override when reaching specific breakpoints (uses the new theme variables)

# Changed
- Changed: Tweaked some logic for Kavita+ Matching on Comics to result in better matches.  (develop)
- Changed: Disable Library settings to manage (create) Collections and Reading Lists from ComicInfo tags by default. 
- Changed: Made the cutoff before the More on the side nav to be slightly larger
- Changed: When opening a Genre, Tag, Person, etc, the title of all series page will show a nicer message, like 'Written by {{name}}' (Closes FR #3825)
- Changed: Browse Authors/Artists is now a Browse People page with filtering and sorting. 
- Changed: Made it more clear that the title of a reading profile is editable by changing the cursor on hover (develop)
- Changed: Changed the mobile layout of the reading profile settings so it's useable (develop)
- Changed: Changed the help text in the reading profile settings to be up-to-date with its behaviour (develop)

# Fixed
- Fixed: Fixed a bug where publisher images would fail to download 
- Fixed: Fixed a localization issue which prevented issue count showing on Series Match Result (develop)
- Fixed: Fixed some minor issues around the rendering of the sidenav items (Thanks @MrRobotjs)
- Fixed: Fixed a rare bug with crypto.UUID by backfilling it.
- Fixed: Fixed a bug where updating the name of a person wasn't updating their normalized name, potentially breaking the flow.
- Fixed: Fixed import menu (Import CBL) in pref side nav missing. (Fixes #3850)
- Fixed: Fixed not being able to create a new reading profile after the default profile is changed in the same session (develop)
- Fixed: Fixed ReadingDirection impacting what clicking does when using UpDown ReadingMode (manga reader)

# Theme
- Added '--setting-mobile-breakpoint', '--setting-tablet-breakpoint', '--setting-desktop-breakpoint' theme variables

# API
- Added reading list owner to the API for Mio's Mihon enhancements.


API/Extensions/QueryExtensions/RestrictByAgeExtensions.cs
Fesaa, line 1: Comment for the RestrictAgainstAgeRestriction for tags (and genres), the implementation requires all series the tag is a part of to be within age restriction. Is this what we want? 

This would hide tags that are used on series the user can see, if they're also used on series they can't see. 

The implementation for Person does use Any, which I think is more appropriate

UI/Web/src/app/_single-module/sort-button/sort-button.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  // input is replacement for @Input
+  disabled = input<boolean>(false);
+  isAscending = input<boolean>(true);
+
+  // signal is for internal state, whenever component needs to update the state internally. Not needed for disabled since component doesn't internally modify
+  private isAscendingSignal = signal(this.isAscending());
+
+  // Computed signals for template
+  protected currentAscending = computed(() => this.isAscendingSignal());
+  protected isDisabled = computed(() => this.disabled());
```
Fesaa: I don't think we need a computed signal for the template, the input is already a signal. So, isDisabled is just an extra step

We _should_ be able to drop the extra ones for ascending as well, but only if we assume the parent will always update the value in the input after it gets the emit as input signals are not writeable. I'm not sure if this has any delay to it. If it does, worth keeping around.

Fesaa: Just read this as well, https://angular.dev/guide/components/inputs#two-way-binding-with-plain-properties, which would allow us to simply the code a lot

```ts
// input is replacement for @Input
  disabled = input<boolean>(false);
  isAscending = model(true);

  updateSortOrder() {
    this.isAscending.set(!this.isAscending())
  }
```

and then
```ts
<app-sort-button [disabled]="filterSettings().sortDisabled" (isAscendingChange)="updateSortOrder($event)" [isAscending]="isAscendingSort" />

```

UI/Web/src/app/all-series/_components/all-series/all-series.component.ts
```diff
@@ -124,13 +132,46 @@ export class AllSeriesComponent implements OnInit {
...
       this.titleService.setTitle('Kavita - ' + this.title);
-      this.filterActiveCheck = this.filterUtilityService.createSeriesV2Filter();
-      this.filterActiveCheck!.statements.push(this.filterUtilityService.createSeriesV2DefaultStatement());
-      this.filterSettings.presetsV2 =  this.filter;
+
+      // To provide a richer experience, when we are browsing just a Genre/Tag/etc, we regenerate the title (if not explicitly passed) to "Browse {GenreName}"
+      if (this.shouldRewriteTitle()) {
+        const field = this.filter!.statements[0].field;
+
+        console.log('field', field);
```
Fesaa: Left over log

UI/Web/src/app/all-series/_components/all-series/all-series.component.ts
```diff
@@ -124,13 +132,46 @@ export class AllSeriesComponent implements OnInit {
...
+
+        // This api returns value as string and number, it will complain without the casting
+        (this.metadataService.getOptionsForFilterField<FilterField>(field, 'series') as Observable<Select2Option[]>).subscribe((opts: Select2Option[]) => {
+          console.log('opts:', opts);
```
Fesaa: Another left over log, we should write these down if they're wanted in nightly but not in stable

UI/Web/src/app/browse/browse-tags/browse-tags.component.scss
```diff
@@ -0,0 +1 @@
+@use '../../../tag-card-common';
```
Fesaa, line 1: There is something odd with the css for tags & genres. 

image

The backgrounds seem to have two layers, it's more clear if you hover over it and the bottom layer does not have an animation or rounded corners. Don't know if this is intended, but I noticed it immediately when I hovered

UI/Web/src/app/cards/card-detail-layout/card-detail-layout.component.ts
```diff
@@ -101,13 +125,21 @@ export class CardDetailLayoutComponent implements OnInit, OnChanges {
...
 
-  filter: SeriesFilterV2 = this.filterUtilityService.createSeriesV2Filter();
   libraries: Array<FilterItem<Library>> = [];
 
   updateApplied: number = 0;
   bufferAmount: number = 1;
 
 
+  filterSignal: WritableSignal<FilterV2<number, number> | undefined> = signal(undefined);
+  hasCustomSort = computed(() => {
```
Fesaa, line 135: I think we need to decide on some rules for both computed and effect. I'm a little paranoid after I had that bug with the effect not updating because the signal wasn't called in the first run. So wouldn't call when the signal updated. 

Is `filteringDisbaled` is false, and later becomes true. The `hasCustomSort` won't be updated if `filterSignal` updates, because it wasn't called before. Only after `customSort` does.

I _think_ in this specific scenario, it won't be an issue with how the variables work. But it's something to think and decide on. 

I think I'm personally going to always write all my signal assignments at the top, or at most under one guard statement. Or only guard statement that require the previous to be true to also be true (to prevent the order of them stopping the method from running)
Don't know if this is going to cause issues later, I can revisit then

UI/Web/src/app/cards/card-detail-layout/card-detail-layout.component.ts
```diff
@@ -125,13 +157,9 @@ export class CardDetailLayoutComponent implements OnInit, OnChanges {
       this.trackByIdentity = (_: number, item: any) => `${this.header}_${this.updateApplied}_${item?.libraryId}`;
```
Fesaa: Header needs to be called, as it's a signal now. 

There is this migration Angular has we could use to migration @Input's; https://angular.dev/reference/migrations/signal-inputs

Fesaa: Is this a unique value? Won't libraryId possibly be shared between items?

UI/Web/src/app/cards/card-detail-layout/card-detail-layout.component.ts
```diff
@@ -101,13 +125,21 @@ export class CardDetailLayoutComponent implements OnInit, OnChanges {
...
   @ViewChild(VirtualScrollerComponent) private virtualScroller!: VirtualScrollerComponent;
 
-  filter: SeriesFilterV2 = this.filterUtilityService.createSeriesV2Filter();
   libraries: Array<FilterItem<Library>> = [];
 
   updateApplied: number = 0;
   bufferAmount: number = 1;
 
 
+  filterSignal: WritableSignal<FilterV2<number, number> | undefined> = signal(undefined);
```
Fesaa, line 134: About naming signals, when do you want to append signal to the name, and when not? To make it consistent across all components.

UI/Web/src/app/metadata-filter/_components/metadata-filter-row/metadata-filter-row.component.ts
```diff
@@ -276,72 +306,29 @@ export class MetadataFilterRowComponent implements OnInit {
...
-  }
-
-
   handleFieldChange(val: string) {
-    const inputVal = parseInt(val, 10) as FilterField;
-
-
-    if (StringFields.includes(inputVal)) {
+    const inputVal = parseInt(val, 10) as TFilter;
+    console.log('input', inputVal);inputVal
```
Fesaa: Left over log, and double copied variable I think?

UI/Web/src/app/_models/library/library.ts
```diff
@@ -13,6 +13,8 @@ export enum LibraryType {
 }
 
 export const allLibraryTypes = [LibraryType.Manga, LibraryType.ComicVine, LibraryType.Comic, LibraryType.Book, LibraryType.LightNovel, LibraryType.Images];
+export const allKavitaPlusMetadataApplicableTypes = [LibraryType.Manga, LibraryType.LightNovel, LibraryType.ComicVine, LibraryType.Comic];
+export const allKavitaPlusEligibleTypes = [LibraryType.Manga, LibraryType.LightNovel];
```
Fesaa: Is this for scrobble? I'd add it in the name

UI/Web/src/app/cards/card-detail-layout/card-detail-layout.component.ts
majora2007: Great catch

UI/Web/src/app/cards/card-detail-layout/card-detail-layout.component.ts
majora2007: I actually don't know, i didn't see any recommendation on the site. I feel like since we don't have $, we should just always have Signal for now. As my editor isn't throwing when not accessing the value of the signal.

UI/Web/src/app/_single-module/sort-button/sort-button.component.ts
majora2007: Yeah, this is very cool. This was the first component I learned signals on, but switching to model is MUCH better.

API/Extensions/QueryExtensions/RestrictByAgeExtensions.cs
majora2007: Good catch, I do believe you're right.

UI/Web/src/app/browse/browse-tags/browse-tags.component.scss
majora2007: Fixed it and left a message for @therobbiedavis to hopefully fix it. Cool hover over effect.

PR #3881, opened 2025-06-29
fix: fix login logo request to respect basepath

Squirreljetpack:
# Fixed
- Fixed: login page makes a malformed request to /assets/images/logo.png when basepath is set - fixed it to respect basepath


majora2007:
We unfortunately are on a release boundary and the code is frozen to prepare for the next stable. I can get this in for v0.8.8.

majora2007:
The implementation isn't correct. Use of / is correct currently. Are you sure you don't have your application setup incorrectly with base url?

PR #3895, opened 2025-07-05
No More Sort Prefixes

majora2007:
# Added
- Added: Added a new library setting to remove common sort prefixes (The, A, An) from Sort titles. Disabled by default (Closes #3037)


API.Tests/Helpers/BookSortTitlePrefixHelperTests.cs
```diff
@@ -0,0 +1,179 @@
...
+    public void TestNonSpaceSeparators(string inputString, string expected)
+    {
+        Assert.Equal(expected, BookSortTitlePrefixHelper.GetSortTitle(inputString));
+    }
+
+    [Theory]
+    [InlineData("三国演义", "三国演义")] // Chinese - no processing due to CJK detection
+    [InlineData("こんにちは世界", "こんにちは世界")] // Japanese Hiragana
+    [InlineData("カタカナ", "カタカナ")] // Japanese Katakana
+    [InlineData("한국어", "한국어")] // Korean - not in CJK range, would be processed normally
```
Fesaa, line 154: Are these processed normally actually testing anything? They just return the same, right?

API.Tests/Services/ScannerServiceTests.cs
```diff
@@ -972,4 +972,28 @@ public async Task ScanLibrary_MetadataDisabled_NoOverrides()
...
     }
+
+    [Fact]
+    public async Task ScanLibrary_SortName_NoPrefix()
+    {
+        const string testcase = "Series with Prefix - Book.json";
+
+        var library = await _scannerHelper.GenerateScannerData(testcase);
+
+        // Disable metadata
```
Fesaa: Wrong comment

API.Tests/Helpers/BookSortTitlePrefixHelperTests.cs
```diff
@@ -0,0 +1,179 @@
...
+    [InlineData("שלום עולם", "שלום עולם")] // Hebrew - not CJK, processed normally
+    public void TestNonLatinNonCjkScripts(string inputString, string expected)
+    {
+        Assert.Equal(expected, BookSortTitlePrefixHelper.GetSortTitle(inputString));
+    }
+
+    [Theory]
+    [InlineData("в мире", "мире")] // Russian "в" (in) - should be removed
+    [InlineData("на столе", "столе")] // Russian "на" (on) - should be removed
+    [InlineData("с друзьями", "друзьями")] // Russian "с" (with) - should be removed
```
Fesaa, line 173: There were checks for an => and, The => Then etc, does this not exist in Russian? If it does, worth adding in the tests

Fesaa: Just saw the implementation, I don't think it's that needed

API.Tests/Helpers/BookSortTitlePrefixHelperTests.cs
```diff
@@ -0,0 +1,179 @@
+﻿using API.Helpers;
+using Xunit;
+
+namespace API.Tests.Helpers;
+
+public class BookSortTitlePrefixHelperTests
+{
+     [Theory]
```
Fesaa: Extra space

Fesaa (approved):
Just nit picks really

API.Tests/Helpers/BookSortTitlePrefixHelperTests.cs
majora2007: Yeah I left these in just in case users asked for the functionality, since from research it doesn't seem like it happens in non-romanji languages.

Issue #2518, opened 2024-01-04
New Feature: OIDC authentication support


jakkarth:
**If this is a feature request, request [here](https://feats.kavitareader.com/) instead. Feature requests will be deleted from Github.**

It's been discussed there as https://feats.kavitareader.com/posts/23/alternative-account-provider but I'd like to take a shot at an actual implementation, which is why I'm posting here.

## The pitch:

OIDC is an open security standard that allows applications to offload the task of identity management and validation. There are large companies that implement OIDC, such as Microsoft and Google and Facebook and Discord, there are companies that provide identity management as a service like Okta, and there are self-hosted options like Keycloak and Authentik. Rather than implement an identity management interface ourselves (eg LDAP), OIDC is a light-weight standard in terms of code changes for Kavita. It also allows users/administrators to create more robust authentication mechanisms like MFA that Kavita's built-in user authentication system does not support, or to integrate with their existing communities eg manga discussion discords. This would also give us feature parity for [Komga](https://komga.org/docs/installation/oauth2/).

## Implementation thoughts

This is my first time proposing a new feature for this project, and I'm doing my best to follow the guidelines outlined in CONTRIBUTING.md. I'm very open to constructive feedback. I have my local dev environment set up, but I have not started working on the changes because I want to check the approach with you all first. It sounds like opening an issue is the right way to do that rather than doing work and then opening a PR.

OIDC is an open standard with a single URL the client can use to request all of the necessary authentication endpoint URLs. The general flow would be that the user clicks "Sign in with SSO" option presented at login, the user is redirected to the identity provider (IdP) and completes the appropriate login process if necessary (may be skipped if user is already logged in, for example), the user is redirected back to Kavita with an authentication code. Kavita on the backend makes a request to the IdP to exchange the code for information about the user, such as preferred username and email address. Kavita would then use that information to see if an existing user matches that information and logs them in, or creates a new account if there's no match.

Minimum configuration for Kavita would include the metadata endpoint, the client id, and the client secret. Optional configuration would include a user-friendly name of the IdP they'll be redirected to, and an option to disable the built-in authentication system so that only OIDC authentication will be possible. We may also need options to map values in the OIDC response to Kavita fields, for example an OIDC provider might use preferredName instead of displayName for the user.

[...]

PR #3912, opened 2025-07-10
OpenID Connect support

Fesaa:
This brings the most requested Feature Request to Kavita (#2518)! Server admins can now set up OpenID Connect (OIDC) for easier login and user management. We've implemented it in such a way that you have lots of control over how OIDC interacts with Kavita; allowing user creations, permission synchronization, or just using it as a login screen. Read more about it on the [Wiki](https://wiki.kavitareader.com/guides/admin-settings/open-id-connect/)!

# Added
- Added: Added OIDC Authentication support. Users can now log in with OIDC providers (Auth0, Keycloak, etc). Admins can setup default user profiles for OIDC-based accounts. 
- Added: Added automatic account creation when logging in via OIDC.
- Added: Allow admins to delegate user management to OIDC (Sync libraries, age restrictions, etc with OIDC roles)
- Added: Added the recent 'Enable Metadata' Library setting to KavitaStats reporting. 

# Fixed
- Fixed: Fixed not being returned to your last used page after being forcefully logged out


API/Controllers/AccountController.cs
```diff
@@ -72,6 +74,32 @@ public AccountController(UserManager<AppUser> userManager,
...
+    /// <exception cref="UnauthorizedAccessException"></exception>
+    /// <remarks>Also throws UnauthorizedAccessException if the users is missing the Login role</remarks>
+    /// <remarks>Syncs Oidc settings if enabled, and user is Oidc owned</remarks>
+    [HttpGet]
+    public async Task<ActionResult<UserDto>> GetCurrentUserAsync()
+    {
+        var user = await _unitOfWork.UserRepository.GetUserByIdAsync(User.GetUserId(), AppUserIncludes.UserPreferences | AppUserIncludes.SideNavStreams);
+        if (user == null) throw new UnauthorizedAccessException();
+
+        if (user.Owner == AppUserOwner.OpenIdConnect)
```
majora2007: I'm not in love with the `Owner` notation. Can we use `Provider` to match the other entities in Kavita?

API/Controllers/AccountController.cs
```diff
@@ -72,6 +74,32 @@ public AccountController(UserManager<AppUser> userManager,
...
         _localizationService = localizationService;
+        _oidcService = oidcService;
+    }
+
+    /// <summary>
+    /// Returns the current user, as it would from login
+    /// </summary>
+    /// <returns></returns>
+    /// <exception cref="UnauthorizedAccessException"></exception>
+    /// <remarks>Also throws UnauthorizedAccessException if the users is missing the Login role</remarks>
```
majora2007: This comment is incorrect. We return an Unauthorized to redirect the user to Login screen.

API/Controllers/AccountController.cs
```diff
@@ -248,6 +281,11 @@ public async Task<ActionResult<UserDto>> Login(LoginDto loginDto)
             }
         }
 
+        return Ok(await ConstructUserDto(user));
+    }
+
+    private async Task<UserDto> ConstructUserDto(AppUser user)
+    {
         // Update LastActive on account
         user.UpdateLastActive();
```
majora2007: This is a side effect. Move it to the create user/login flows. This also doesn't do anything here now because it needs a commit to save.

API/Controllers/AccountController.cs
```diff
@@ -505,6 +542,7 @@ public async Task<ActionResult> UpdateAgeRestriction(UpdateAgeRestrictionDto dto
     /// </summary>
     /// <param name="dto"></param>
     /// <returns></returns>
+    /// <remarks>OIDC managed users cannot be edited if SyncUsers is enabled</remarks>
```
majora2007: Is syncUsers a Setting? If so, use a see tag to help drive that link.

API/Controllers/AccountController.cs
```diff
@@ -517,6 +555,28 @@ public async Task<ActionResult> UpdateAccount(UpdateUserDto dto)
         var user = await _unitOfWork.UserRepository.GetUserByIdAsync(dto.UserId, AppUserIncludes.SideNavStreams);
         if (user == null) return BadRequest(await _localizationService.Translate(User.GetUserId(), "no-user"));
 
+        // Disallowed editing users owned by OIDC
+        var oidcSettings = (await _unitOfWork.SettingsRepository.GetSettingsDtoAsync()).OidcConfig;
+        if (user.Owner == AppUserOwner.OpenIdConnect && dto.Owner != AppUserOwner.Native && oidcSettings.SyncUserSettings)
```
majora2007: Instead of AppUserOwner as the enum, let's use `IdentityProvider.Kavita` instead. It reads much more clear. Native has some connotation if we ever build an Electron-like app and use Windows login/etc.

majora2007: This line doesn't make sense, should be:
`if (user.Owner == AppUserOwner.OpenIdConnect && oidcSettings.SyncUserSettings)`

API/Controllers/AccountController.cs
```diff
@@ -517,6 +555,28 @@ public async Task<ActionResult> UpdateAccount(UpdateUserDto dto)
...
+
+        var defaultAdminUser = await _unitOfWork.UserRepository.GetDefaultAdminUser();
+        if (user.Id == defaultAdminUser.Id && dto.Owner != AppUserOwner.Native)
+        {
+            return BadRequest(await _localizationService.Translate(User.GetUserId(), "cannot-change-ownership-original-user"));
+        }
+
+        user.Owner = dto.Owner;
+
+        if (user.Owner == AppUserOwner.OpenIdConnect)
```
majora2007: I would like to see these OIDC checks in a private method to make it easier to understand. Do the checks, make the change, and commit. Return true if successful, otherwise false. On false, will not return and fall through to rest of code. We can then unit test it to ensure the validations work. 

```cs
try {
if (await _accountService.ChangeAuthProvider(...)) {
return Ok();
}
} catch (...)
```

API/Controllers/OidcControlller.cs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /// <summary>
+    /// Validate if the given authority is reachable from the server
+    /// </summary>
+    /// <param name="authority"></param>
+    /// <returns></returns>
+    [Authorize("RequireAdminRole")]
+    [HttpPost("is-valid-authority")]
+    public async Task<ActionResult<bool>> IsValidAuthority([FromBody] IsValidAuthorityBody authority)
```
majora2007: This is not a good name for the payload.

API/DTOs/Account/UpdateUserDto.cs
```diff
@@ -25,4 +26,5 @@ public sealed record UpdateUserDto
     public AgeRestrictionDto AgeRestriction { get; init; } = default!;
     /// <inheritdoc cref="API.Entities.AppUser.Email"/>
     public string? Email { get; set; } = default!;
+    public AppUserOwner Owner { get; init; }
```
majora2007: I'd like to see the default here (that matches what's in the DataContext)

API/DTOs/Settings/OidcConfigDto.cs
```diff
@@ -0,0 +1,38 @@
+﻿#nullable enable
+
+using System.Collections.Generic;
+using API.Entities.Enums;
+
+namespace API.DTOs.Settings;
+
+public record OidcConfigDto: OidcPublicConfigDto
```
majora2007: I would like a comment to help understand what the difference between this and Public is.

API/DTOs/Settings/OidcConfigDto.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// If true, auto creates a new account when someone logs in via OpenID Connect
+    /// </summary>
+    public bool ProvisionAccounts { get; set; }
+    /// <summary>
+    /// Require emails to be verified by the OpenID Connect provider when creating accounts on login
+    /// </summary>
+    public bool RequireVerifiedEmail { get; set; } = true;
+    /// <summary>
+    /// Overwrite Kavita roles, libraries and age rating with OpenIDConnect provides roles on log in.
```
majora2007: provided

API/DTOs/Settings/OidcConfigDto.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public bool SyncUserSettings { get; set; }
+
+    // Default values used when SyncUserSettings is false
+    #region Default user settings
+
+    public List<string> DefaultRoles { get; set; } = [];
+    public List<int> DefaultLibraries { get; set; } = [];
+    public AgeRating DefaultAgeRating { get; set; } = AgeRating.Unknown;
```
majora2007: I would like to see Restriction here. To make it more clear.

API/DTOs/Settings/OidcConfigDto.cs
```diff
@@ -0,0 +1,38 @@
...
+    public bool DefaultIncludeUnknowns { get; set; } = false;
+
+    #endregion
+
+
+    /// <summary>
+    /// Returns true if the <see cref="OidcPublicConfigDto.Authority"/> has been set
+    /// </summary>
+    public bool Enabled => Authority != "";
```
majora2007: Return `!string.IsNullOrEmpty(Authority)`

API/DTOs/Settings/OidcPublicConfigDto.cs
```diff
@@ -0,0 +1,25 @@
+#nullable enable
+using API.Entities.Enums;
+
+namespace API.DTOs.Settings;
+
+public record OidcPublicConfigDto
```
majora2007, line 9: Again, a lack of top level comments makes it hard to understand what this is for.

API/DTOs/Settings/OidcConfigDto.cs
majora2007: This should be sealed

API/DTOs/Stats/V3/UserStatV3.cs
```diff
@@ -76,6 +77,10 @@ public sealed record UserStatV3
     /// Roles for this user
     /// </summary>
     public ICollection<string> Roles { get; set; }
+    /// <summary>
+    /// Who manages the user (OIDC, Kavita)
+    /// </summary>
+    public AppUserOwner Owner { get; set; }
```
majora2007: Don't forget to refactor this to Provider as well. (and update the KStats too)

API/Data/Migrations/20250520073818_OpenIDConnect.cs
```diff
@@ -0,0 +1,28 @@
...
+
+#nullable disable
+
+namespace API.Data.Migrations
+{
+    /// <inheritdoc />
+    public partial class OpenIDConnect : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
```
majora2007: Can you redo migrations so they are combined into 1 single migration?

API/Data/Repositories/UserRepository.cs
```diff
@@ -557,6 +558,16 @@ public async Task<AppUser> GetUserByDeviceEmail(string deviceEmail)
             .FirstOrDefaultAsync();
     }
 
+    public async Task<AppUser?> GetByExternalId(string? externalId, AppUserIncludes includes = AppUserIncludes.None)
```
majora2007: Documentation needed.

API/Data/Migrations/20250520073818_OpenIDConnect.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "ExternalId",
```
majora2007: ExternalId is not easy to understand just with the variable name. Let's use something better.

API/Data/Migrations/DataContextModelSnapshot.cs
```diff
@@ -116,6 +119,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasMaxLength(256)
                         .HasColumnType("TEXT");
 
+                    b.Property<int>("Owner")
+                        .HasColumnType("INTEGER");
```
majora2007: Setup the default here

API/Entities/Enums/AppUserOwner.cs
```diff
@@ -0,0 +1,16 @@
+namespace API.Entities.Enums;
+
+/// <summary>
+/// Who own the user, can be updated in the UI if desired
+/// </summary>
+public enum AppUserOwner
+{
+    /**
```
majora2007: Use the correct comment tags. 

You need [Description("Native")] annotation.

API/Extensions/EnumerableExtensions.cs
```diff
@@ -68,4 +69,9 @@ public static IEnumerable<Chapter> RestrictAgainstAgeRestriction(this IEnumerabl
 
         return q;
     }
+
+    public static string AsJoinedString(this IEnumerable<IdentityError> errors)
+    {
+        return string.Join(",", errors.Select(e => e.Description));
```
majora2007: Can you use the ',' overload?

API/Extensions/IdentityServiceExtensions.cs
```diff
@@ -47,42 +63,146 @@ public static IServiceCollection AddIdentityServices(this IServiceCollection ser
...
             .AddEntityFrameworkStores<DataContext>();
 
+        var auth = services.AddAuthentication(DynamicJwt)
+            .AddPolicyScheme(DynamicJwt, JwtBearerDefaults.AuthenticationScheme, options =>
+            {
+                var iss = Configuration.OidcAuthority;
+                var enabled = Configuration.OidcEnabled;
+                options.ForwardDefaultSelector = context =>
+                {
+                    if (!enabled)
```
majora2007: return statement needs to be on the same line as if

API/Extensions/IdentityServiceExtensions.cs
```diff
@@ -47,42 +63,146 @@ public static IServiceCollection AddIdentityServices(this IServiceCollection ser
...
 
-        services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
-            .AddJwtBearer(options =>
+                    return LocalIdentity;
+                };
+            });
+
+        if (Configuration.OidcEnabled)
+        {
+            // TODO: Investigate on how to make this not hardcoded at startup
```
majora2007: Is this still relevant?

API/Extensions/IdentityServiceExtensions.cs
```diff
@@ -47,42 +63,146 @@ public static IServiceCollection AddIdentityServices(this IServiceCollection ser
...
+                        /* Swallow */
+                    }
 
-        services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
-            .AddJwtBearer(options =>
+                    return LocalIdentity;
+                };
+            });
+
+        if (Configuration.OidcEnabled)
```
majora2007, line 98: I dont' recall seeing OidEnabled in the settings reference that it's maintained in Configuration (appsetting)

API/Extensions/IdentityServiceExtensions.cs
```diff
@@ -47,42 +63,146 @@ public static IServiceCollection AddIdentityServices(this IServiceCollection ser
...
+        var user = await oidcService.LoginOrCreate(ctx.Principal);
+        if (user == null)
+        {
+            ctx.Principal = null;
+            await ctx.HttpContext.SignOutAsync(OpenIdConnect);
+            return;
+        }
+
+        // Add the following claims like Kavita expects them
+        var claims = new List<Claim>
```
majora2007, line 168: Since this code is now effectively duplicated, can we use one method?

API/Helpers/Converters/ServerSettingConverter.cs
```diff
@@ -129,6 +130,21 @@ public ServerSettingDto Convert(IEnumerable<ServerSetting> source, ServerSetting
...
+                    destination.OidcConfig ??= new OidcConfigDto();
+                    destination.OidcConfig.Authority = row.Value;
+                    break;
+                case ServerSettingKey.OidcClientId:
+                    destination.OidcConfig ??= new OidcConfigDto();
+                    destination.OidcConfig.ClientId = row.Value;
+                    break;
+                case ServerSettingKey.OidcConfiguration:
+                    destination.OidcConfig ??= new OidcConfigDto();
+                    var configuration = JsonSerializer.Deserialize<OidcConfigDto>(row.Value)!;
```
majora2007, line 143: You need to handle the try catch so it doesn't break everything

API/I18N/en.json
```diff
@@ -17,6 +18,8 @@
     "generate-token": "There was an issue generating a confirmation email token. See logs",
     "age-restriction-update": "There was an error updating the age restriction",
     "no-user": "User does not exist",
+    "oidc-managed": "This user is managed by OIDC, cannot edit",
+    "cannot-change-ownership-original-user": "Ownership of the original admin account cannot be changed",
```
majora2007: Ownership should be changed to Login Provider.

API/I18N/en.json
```diff
@@ -17,6 +18,8 @@
     "generate-token": "There was an issue generating a confirmation email token. See logs",
     "age-restriction-update": "There was an error updating the age restriction",
     "no-user": "User does not exist",
+    "oidc-managed": "This user is managed by OIDC, cannot edit",
```
majora2007: Users managed by OIDC cannot be edited.

API/Services/AccountService.cs
```diff
@@ -29,20 +31,36 @@ public interface IAccountService
     Task<bool> HasBookmarkPermission(AppUser? user);
     Task<bool> HasDownloadPermission(AppUser? user);
     Task<bool> CanChangeAgeRestriction(AppUser? user);
+
+    /// <summary>
+    ///
```
majora2007, line 36: Fill this in

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+{
+    /// <summary>
+    /// Returns the user authenticated with OpenID Connect
+    /// </summary>
+    /// <param name="principal"></param>
+    /// <returns></returns>
+    /// <exception cref="KavitaException">if any requirements aren't met</exception>
+    Task<AppUser?> LoginOrCreate(ClaimsPrincipal principal);
+    /// <summary>
+    /// Updates roles, library access and age rating. Will not modify the default admin
```
majora2007: age rating restrictions

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+    private const string LibraryAccessPrefix = "library-";
+    private const string AgeRatingPrefix = "age-rating-";
+    private const string IncludeUnknowns = AgeRatingPrefix + "include-unknowns";
+
+    public async Task<AppUser?> LoginOrCreate(ClaimsPrincipal principal)
+    {
+        var settings = (await unitOfWork.SettingsRepository.GetSettingsDtoAsync()).OidcConfig;
+
+        var externalId = principal.FindFirstValue(ClaimTypes.NameIdentifier);
+        if (string.IsNullOrEmpty(externalId))
```
majora2007: you need {}

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+            throw new KavitaException("errors.oidc.missing-external-id");
+
+        var user = await unitOfWork.UserRepository.GetByExternalId(externalId, AppUserIncludes.UserPreferences);
+        if (user != null) return user;
+
+        var email = principal.FindFirstValue(ClaimTypes.Email);
+        if (string.IsNullOrEmpty(email))
```
majora2007, line 65: {} when you have throw statements

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+            throw new KavitaException("errors.oidc.missing-email");
+
+        if (settings.RequireVerifiedEmail && !principal.HasVerifiedEmail())
+            throw new KavitaException("errors.oidc.email-not-verified");
+
+
+        user = await unitOfWork.UserRepository.GetUserByEmailAsync(email, AppUserIncludes.UserPreferences | AppUserIncludes.SideNavStreams);
+        if (user != null)
+        {
+            logger.LogInformation("User {UserName} has matched on email to {ExternalId}", user.Id, externalId);
```
majora2007: Do you think this should be Info? I feel debug is better suited.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+            user.ExternalId = externalId;
+            await unitOfWork.CommitAsync();
+            return user;
```
majora2007, line 83: newline above return

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+    {
+        if (!settings.ProvisionAccounts) return null;
+
+        var emailClaim = claimsPrincipal.FindFirst(ClaimTypes.Email);
+        if (emailClaim == null || string.IsNullOrWhiteSpace(emailClaim.Value)) return null;
+
+        var name = await FindBestAvailableName(claimsPrincipal) ?? emailClaim.Value;
+        logger.LogInformation("Creating new user from OIDC: {Name} - {ExternalId}", name, externalId);
+
+        // TODO: Move to account service, as we're sharing code with AccountController
```
majora2007: Remove TODO

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+        var users = await unitOfWork.UserRepository.GetAllUsersAsync();
+        foreach (var user in users)
+        {
+            user.ExternalId = null;
+        }
+
+        await unitOfWork.CommitAsync();
+    }
+
+    private async Task<string?> FindBestAvailableName(ClaimsPrincipal claimsPrincipal)
```
majora2007: Please make sure you have a unit test that validates this fallback for code coverage.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+
+        return await userManager.FindByNameAsync(name) == null;
+    }
+
+    private async Task<AppUser?> NewUserFromOpenIdConnect(OidcConfigDto settings, ClaimsPrincipal claimsPrincipal, string externalId)
+    {
+        if (!settings.ProvisionAccounts) return null;
+
+        var emailClaim = claimsPrincipal.FindFirst(ClaimTypes.Email);
+        if (emailClaim == null || string.IsNullOrWhiteSpace(emailClaim.Value)) return null;
```
majora2007: Rewrite as `if (string.IsNullOrWhiteSpace(emailClaim?.Value)) return null;`

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+        user.Owner = AppUserOwner.OpenIdConnect;
+
+        accountService.AddDefaultStreamsToUser(user);
+        await accountService.AddDefaultReadingProfileToUser(user);
+
+        await SyncUserSettings(settings, claimsPrincipal, user);
+        await SetDefaults(settings, user);
+
+        await unitOfWork.CommitAsync();
+        return user;
```
majora2007, line 191: Newline above return statements always.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+        user.ExternalId = externalId;
+        user.Owner = AppUserOwner.OpenIdConnect;
+
+        accountService.AddDefaultStreamsToUser(user);
+        await accountService.AddDefaultReadingProfileToUser(user);
+
+        await SyncUserSettings(settings, claimsPrincipal, user);
+        await SetDefaults(settings, user);
+
+        await unitOfWork.CommitAsync();
```
majora2007, line 189: There is no try/catch handling in this code at all.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+        return user;
+    }
+
+    private async Task SetDefaults(OidcConfigDto settings, AppUser user)
```
majora2007, line 197: Comments to describe the method (just summary) would be helpful.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+        if (defaultAdminUser.Id == user.Id) return;
+
+        logger.LogInformation("Syncing user {UserName} from OIDC", user.UserName);
+        await SyncRoles(claimsPrincipal, user);
+        await SyncLibraries(claimsPrincipal, user);
+        await SyncAgeRestriction(claimsPrincipal, user);
+
+
+        if (unitOfWork.HasChanges())
+            await unitOfWork.CommitAsync();
```
majora2007: You need {}. You can only leave off with if statements if a jump (return, continue) and it must be on the same line.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+
+    public async Task SyncUserSettings(OidcConfigDto settings, ClaimsPrincipal claimsPrincipal, AppUser user)
+    {
+        if (!settings.SyncUserSettings) return;
+
+        // Never sync the default user
+        var defaultAdminUser = await unitOfWork.UserRepository.GetDefaultAdminUser();
+        if (defaultAdminUser.Id == user.Id) return;
+
+        logger.LogInformation("Syncing user {UserName} from OIDC", user.UserName);
```
majora2007: This will get spammy, I would move to Debug, especially since this sync is at least 4 log lines.

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+
+        var errors = await accountService.UpdateRolesForUser(user, roles);
+        if (errors.Any()) throw new KavitaException("errors.oidc.syncing-user");
+    }
+
+    private async Task SyncLibraries(ClaimsPrincipal claimsPrincipal, AppUser user)
+    {
+        var libraryAccess = claimsPrincipal
+            .FindAll(ClaimTypes.Role)
+            .Where(r => r.Value.StartsWith(LibraryAccessPrefix))
```
majora2007, line 258: How is this LibraryAccessPrefix set? If I remove a library from an account, will it be updated in my JWT instantly?

API/Services/OidcService.cs
```diff
@@ -0,0 +1,271 @@
...
+            .Select(r => r.Value.TrimPrefix(LibraryAccessPrefix))
+            .ToList();
+
+        logger.LogDebug("Syncing libraries for user {UserName}, found library roles {Roles}", user.UserName, libraryAccess);
+
+        var allLibraries = (await unitOfWork.LibraryRepository.GetLibrariesAsync()).ToList();
+        var librariesIds = allLibraries.Where(l => libraryAccess.Contains(l.Name)).Select(l => l.Id).ToList();
+
+        var hasAdminRole = await userManager.IsInRoleAsync(user, PolicyConstants.AdminRole);
+        await accountService.UpdateLibrariesForUser(user, librariesIds, hasAdminRole);
```
majora2007, line 269: Try/catch handling

API/Services/SettingsService.cs
```diff
@@ -379,6 +410,41 @@ private bool UpdateSchedulingSettings(ServerSetting setting, ServerSettingDto up
         return false;
     }
 
+    private async Task UpdateOidcSettings(ServerSetting setting, ServerSettingDto updateSettingsDto)
```
majora2007, line 414: This needs to explicitly mention it does not commit

Kavita.Common/Configuration.cs
```diff
@@ -50,6 +52,20 @@ public static long CacheSize
...
+        set => SetOidcAuthority(GetAppSettingFilename(), value);
+    }
+
+    public static string OidcClientId
+    {
+        get => GetOidcClientId(GetAppSettingFilename());
+        set => SetOidcClientId(GetAppSettingFilename(), value);
+    }
+
+    public static bool OidcEnabled => GetOidcAuthority(GetAppSettingFilename()) != "";
```
majora2007: string.Empty over ""

Kavita.Common/Configuration.cs
```diff
@@ -312,6 +328,74 @@ private static bool GetAllowIFraming(string filePath)
...
 
+    #region OIDC
+
+    private static string GetOidcAuthority(string filePath)
+    {
+        try
+        {
+            var json = File.ReadAllText(filePath);
+            var jsonObj = JsonSerializer.Deserialize<AppSettings>(json);
+            return jsonObj.OidcAuthority;
```
majora2007, line 340: newline above returns.

UI/Web/src/app/_models/user.ts
```diff
@@ -4,6 +4,9 @@ import {Preferences} from './preferences/preferences';
 // This interface is only used for login and storing/retrieving JWT from local storage
 export interface User {
   username: string;
+  // This is set by the oidc service, will always take precedence over the Kavita generated token
+  // When set, the refresh logic for the Kavita token will not run
```
majora2007: Use proper js comment (/** */)

UI/Web/src/app/_pipes/user-owner.pipe.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Pipe({
+  name: 'userOwnerPipe'
+})
+export class UserOwnerPipe implements PipeTransform {
+
+  transform(value: UserOwner, ...args: unknown[]): string {
+    switch (value) {
+      case UserOwner.Native:
+        return translate("creation-source-pipe.native");
```
majora2007: Your prefix for localization should match the component it came from. Otherwise, it is difficult to make changes quickly. This should be `user-owner-pipe.` (or rather `login-provider-pipe.` once you refactor the name like in a previous comment.

UI/Web/src/app/_pipes/user-owner.pipe.ts
```diff
@@ -0,0 +1,19 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import {UserOwner} from "../_models/user";
+import {translate} from "@jsverse/transloco";
+
+@Pipe({
+  name: 'userOwnerPipe'
+})
+export class UserOwnerPipe implements PipeTransform {
+
+  transform(value: UserOwner, ...args: unknown[]): string {
```
majora2007: Remove args if not in use

UI/Web/src/app/_services/account.service.ts
```diff
@@ -63,6 +63,8 @@ export class AccountService {
     return this.hasAdminRole(u);
   }), shareReplay({bufferSize: 1, refCount: true}));
 
+  public currentUserSignal = toSignal(this.currentUserSource);
```
majora2007: make this readonly

UI/Web/src/app/_services/nav.service.ts
```diff
@@ -173,12 +175,28 @@ export class NavService {
...
     this.hideSideNav();
     this.router.navigateByUrl('/login');
   }
 
+  handleLogin() {
+    this.showNavBar();
+    this.showSideNav();
+
+    // Check if user came here from another url, else send to library route
+    const pageResume = localStorage.getItem('kavita--auth-intersection-url');
```
majora2007: A side note that this doesn't work.

UI/Web/src/app/admin/edit-user/edit-user.component.ts
```diff
@@ -32,7 +44,16 @@ export class EditUserComponent implements OnInit {
   private readonly destroyRef = inject(DestroyRef);
   protected readonly modal = inject(NgbActiveModal);
 
-  @Input({required: true}) member!: Member;
+
+  // Needs to be models, so we can set it manually
```
majora2007: We can remove this comment now that we both understand signals better.

UI/Web/src/app/admin/library-selector/library-selector.component.ts
```diff
@@ -29,6 +29,7 @@ export class LibrarySelectorComponent implements OnInit {
   private readonly cdRef = inject(ChangeDetectorRef);
 
   @Input() member: Member | undefined;
+  preselectedLibraries = input<number[]>([]);
```
majora2007: Newline between inputs and outputs please.

majora2007: preSelectedLibraries

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+    <button type="button" class="btn btn-primary position-absolute custom-position" (click)="save()">{{t('save')}}</button>
+  </div>
+
+  <form [formGroup]="settingsForm">
+    <div class="alert alert-warning" role="alert">
+      <strong>{{t('notice')}}</strong> {{t('restart-required')}}
+    </div>
+
+    <h4>{{t('provider')}}</h4>
+    <div class="text-muted">{{t('manual-save')}}</div>
```
majora2007: Use -label for the labels. This is the standard pattern (although some pages don't have it).

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+            </ng-template>
+            <ng-template #edit>
+              <input id="oid-authority" class="form-control"
+                     formControlName="authority" type="text"
+                     [class.is-invalid]="formControl.invalid && !formControl.untouched">
+
+              @if (settingsForm.dirty || !settingsForm.untouched) {
+                <div id="oidc-authority-validations" class="invalid-feedback">
+                  @if (formControl.errors?.invalidUri) {
+                    <div>{{t('invalidUri')}}</div>
```
majora2007: This doesn't match our style. Should be invalid-uri-validation or something and this should be stored in the validation prefix with a reference here.

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
+<ng-container *transloco="let t; prefix:'oidc.settings'">
```
majora2007: prefix names should match their components.

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+                     formControlName="providerName" type="text"
+                     [class.is-invalid]="formControl.invalid && !formControl.untouched">
+            </ng-template>
+          </app-setting-item>
+        }
+      </div>
+
+      <div class="row g-0 mt-4 mb-4">
+        @if(settingsForm.get('provisionAccounts'); as formControl) {
+          <app-setting-switch [title]="t('provisionAccounts')" [subtitle]="t('provisionAccounts-tooltip')">
```
majora2007: You need to use kebab case always for locale strings.

UI/Web/src/app/admin/manage-users/manage-users.component.ts
```diff
@@ -64,6 +70,10 @@ export class ManageUsersComponent implements OnInit {
 
   ngOnInit(): void {
     this.loadMembers();
+
+    this.settingsService.getServerSettings().subscribe(settings => {
+      this.settings = settings;
```
majora2007, line 75: Does this not need a cdRef.markForCheck()?

UI/Web/src/app/admin/role-selector/role-selector.component.ts
```diff
@@ -33,6 +33,7 @@ export class RoleSelectorComponent implements OnInit {
    * This must have roles
    */
   @Input() member: Member | undefined | User;
+  preSelectedRoles = input<string[]>([]);
```
majora2007, line 36: preSelectedRoles

UI/Web/src/app/registration/user-login/user-login.component.html
```diff
@@ -1,32 +1,51 @@
...
-        </form>
-      </ng-container>
+          </form>
+        }
+
+        @if (oidcService.inUse()) {
+          <button
+            class="btn btn-outline-primary mt-2 d-flex align-items-center gap-2"
+            (click)="oidcService.login()">
+            <img
```
majora2007: use app-image whenever possible.

UI/Web/src/app/registration/user-login/user-login.component.ts
```diff
@@ -1,26 +1,46 @@
...
+  private readonly accountService = inject(AccountService);
+  private readonly router = inject(Router);
+  private readonly memberService = inject(MemberService);
+  private readonly toastr = inject(ToastrService);
+  private readonly navService = inject(NavService);
+  private readonly cdRef = inject(ChangeDetectorRef);
+  private readonly route = inject(ActivatedRoute);
+  protected readonly oidcService = inject(OidcService);
+
+  baseUrl = environment.apiUrl;
```
majora2007, line 41: readonly this

UI/Web/src/app/shared/_components/owner-icon/owner-icon.component.html
```diff
@@ -0,0 +1,9 @@
+@switch (owner()) {
+  @case (UserOwner.OpenIdConnect) {
+    <img [width]="size()" [height]="size()" ngSrc="assets/icons/open-id-connect-logo.svg" alt="open-id-connect-logo">
```
majora2007: I'd like see how this functions with app-image

UI/Web/src/app/shared/_components/owner-icon/owner-icon.component.ts
```diff
@@ -0,0 +1,19 @@
...
+import {NgOptimizedImage} from "@angular/common";
+import {UserOwner} from "../../../_models/user";
+
+@Component({
+  selector: 'app-owner-icon',
+    imports: [
+        NgOptimizedImage
+    ],
+  templateUrl: './owner-icon.component.html',
+  styleUrl: './owner-icon.component.scss'
```
majora2007: ChangeDetection is missing here, should be onPush

majora2007 (changes requested):
I just reviewed the code itself, I have not pulled this down and tested. I will do that after you make your changes (since there are quite a lot).

API/Controllers/AccountController.cs
Fesaa: I've updated the naming, the logic is correct, however. As, admins should be able to take control of a user by changing their Identity Provider to Kavita

API/Controllers/AccountController.cs
Fesaa: I'm going to hold off on this change and think a bit more on how to do this, as there's different error messages being returned

API/Extensions/EnumerableExtensions.cs
Fesaa: I've just removed this code, I ended up not using it

API/Extensions/IdentityServiceExtensions.cs
Fesaa: I did find a way this _might_ be possible, but I do not think it's worth the complexity it would bring. You're not supposed to change OIDC providers all that much, so restarting is fine here imo.

API/Extensions/IdentityServiceExtensions.cs
Fesaa: I don't think I understand this comment. It's mirrored in appsettings as I need to access it on startup (unless I can make a db query here?). 

I've added a remark in the ServerSettingKey entry mentioning it's mirrored in appsettings, if that's what you meant?

API/Extensions/IdentityServiceExtensions.cs
Fesaa: Where is this duplicated?

API/Helpers/Converters/ServerSettingConverter.cs
Fesaa: The other parses don't try catch, unless someone touches the DB this _should_ never fail. If it does fail, what should we return? 

I'm not sure if there is a valid state for this to recover from if the conversion fails

API/Services/OidcService.cs
Fesaa: You're right, if it matches it's on an email admins can already see. (And they most likely manage the OIDC as well)

API/Services/OidcService.cs
Fesaa: Have wrapped the call to the method in a try catch

API/Extensions/IdentityServiceExtensions.cs
majora2007: What I'm mentioning is in ServerSettingsKey, I don't recall seeing a remarks that it's actually managed in Configuration, like we have elsewhere:
image

API/Services/OidcService.cs
Fesaa: These are roles configured in the OIDC Provider, so will only be used if you have the SyncUserSettings enabled, and if this is, you cannot remove a library in the UI and have to do it via your OIDC provider. 

Once you've changed it there, the next token retrieved will have the changes, and these will be synced to Kavita the next time the user logs in (/ the new CurrentAccount endpoint is called)

For mine it looks like this, and I can assign these roles to users there.
image

API/Services/OidcService.cs
Fesaa: Have made it try catch in the SyncUserSettings method, so if any sync fails, none do. (To prevent half synced users in the system)

UI/Web/src/app/_pipes/user-owner.pipe.ts
Fesaa: I changed names a few times already XD

UI/Web/src/app/_services/nav.service.ts
Fesaa: I noticed, will see if I can fix it (I just moved this code). If I can't fix it in reasonable time, will add the note

UI/Web/src/app/admin/manage-users/manage-users.component.ts
Fesaa: It's only used when opening the modal so it's not needed

UI/Web/src/app/admin/role-selector/role-selector.component.ts
Fesaa: I don't see the difference 😓

UI/Web/src/app/_services/nav.service.ts
Fesaa: Fixed it (Will need some testing)

API/Services/OidcService.cs
Fesaa: I've made changes that when the OIDC token is refreshed, Kavita will re-request the account, and thus syncing. Changes will be synced at least every 3/4*token exp time (this will depend on the OIDC provider) (the 3/4th is from the library we use, we can configure this but there is no need to change it really)

API/Services/SettingsService.cs
```diff
@@ -346,6 +356,27 @@ public async Task<ServerSettingDto> UpdateSettings(ServerSettingDto updateSettin
...
+        {
+            return false;
+        }
+
+        var url = authority + "/.well-known/openid-configuration";
+        try
+        {
+            var json = await url.GetStringAsync();
+            var config = OpenIdConnectConfiguration.Create(json);
+            return config.Issuer == Configuration.OidcAuthority;
```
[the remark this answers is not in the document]
Fesaa: This is a good point, but if we allow it here; we'd need to change it in the DynamicJwt check too. And I'd rather have code running on each request be as simple as possible

As the settings won't update unless this passes, it's safe to keep it. All OIDC providers I've tried have adhered to exact equality. Unless someone brings up the issue, I think it's fine to keep

API.Tests/Extensions/EnumExtensionTests.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+
+    [Theory]
+    [InlineData("Early Childhood", AgeRating.EarlyChildhood, true)]
+    [InlineData("M", AgeRating.Mature, true)]
+    [InlineData("ThisIsNotAnAgeRating", AgeRating.NotApplicable, false)]
+    public void TryParse<TEnum>(string? value, TEnum expected, bool success) where TEnum : struct, Enum
+    {
+        Assert.Equal(EnumExtensions.TryParse(value, out TEnum got), success);
+        if (success)
```
majora2007: Can't we just leave Assert.Equal() always?

API/Controllers/AccountController.cs
```diff
@@ -72,6 +75,38 @@ public AccountController(UserManager<AppUser> userManager,
...
+        {
+            var oidcSettings = (await _unitOfWork.SettingsRepository.GetSettingsDtoAsync()).OidcConfig;
+            await _oidcService.SyncUserSettings(oidcSettings, User, user);
+        }
+
+        var roles = await _userManager.GetRolesAsync(user);
+        if (!roles.Contains(PolicyConstants.LoginRole)) return Unauthorized(await _localizationService.Translate(user.Id, "disabled-account"));
+
+        // Update LastActive on account
+        user.UpdateLastActive();
```
majora2007: Can we wrap this in a try/catch and log/swallow the exception? On some users machines, the db throws an exception on updating last login which propagates to UI.

API/Extensions/IdentityServiceExtensions.cs
```diff
@@ -47,42 +63,143 @@ public static IServiceCollection AddIdentityServices(this IServiceCollection ser
...
+
+        ctx.Success();
+    }
+
+    private static Task SetTokenFromQuery(MessageReceivedContext context)
+    {
+        var accessToken = context.Request.Query["access_token"];
+        var path = context.HttpContext.Request.Path;
+        // Only use query string based token on SignalR hubs
+        if (!string.IsNullOrEmpty(accessToken) && path.StartsWithSegments("/hubs")) context.Token = accessToken;
```
majora2007: You need {} here

API/Services/OidcService.cs
majora2007: So just to understand a bit more, if you're using OIDC and Sync Settings, you cannot remove libraries from the Kavita UI?

API/Services/OidcService.cs
```diff
@@ -0,0 +1,312 @@
...
+        var libraryAccess = claimsPrincipal
+            .FindAll(ClaimTypes.Role)
+            .Where(r => r.Value.StartsWith(LibraryAccessPrefix))
+            .Select(r => r.Value.TrimPrefix(LibraryAccessPrefix))
+            .ToList();
+
+        logger.LogDebug("Syncing libraries for user {UserName}, found library roles {Roles}", user.UserName, libraryAccess);
+
+        var allLibraries = (await unitOfWork.LibraryRepository.GetLibrariesAsync()).ToList();
+        // District as a failsafe for duplicate roles in the JWT
```
majora2007: comment doesn't make sense.

Kavita.Common/Configuration.cs
```diff
@@ -14,6 +14,8 @@ public static class Configuration
     public const int DefaultHttpPort = 5000;
     public const int DefaultTimeOutSecs = 90;
     public const long DefaultCacheMemory = 75;
+    public const string DefaultOidcAuthority = "";
```
majora2007, line 17: Use string.Empty

UI/Web/src/app/_pipes/user-owner.pipe.ts
```diff
@@ -0,0 +1,19 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import {IdentityProvider} from "../_models/user";
+import {translate} from "@jsverse/transloco";
+
+@Pipe({
+  name: 'IdentityProviderPipe'
```
majora2007: This should start with a lowercase letter.

UI/Web/src/app/_services/account.service.ts
```diff
@@ -205,6 +207,22 @@ export class AccountService {
     );
   }
 
+  loginByToken(token: string) {
+    const headers = new HttpHeaders({
+      "Authorization": `Bearer ${token}`
+    })
```
majora2007: semicolon

UI/Web/src/app/_services/nav.service.ts
majora2007: Shouldn't we use the AuthGuard property?

UI/Web/src/app/_services/oidc.service.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (this.token) {
+      this.oauth2.logOut();
+    }
+  }
+
+  getPublicOidcConfig() {
+    return this.httpClient.get<OidcPublicConfig>(this.apiBaseUrl + "oidc/config");
+  }
+
+  get token() {
```
majora2007: Let's just use a method vs a getter.

UI/Web/src/app/admin/edit-user/edit-user.component.scss
```diff
@@ -0,0 +1,3 @@
+.text-muted {
+  font-size: 12px;
```
majora2007: Use rem

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+                </div>
+              }
+            </ng-template>
+          </app-setting-item>
+        }
+      </div>
+
+      <div class="row g-0 mt-4 mb-4">
+        @if (settingsForm.get('clientId'); as formControl) {
+          <app-setting-item [title]="t('clientId')" [subtitle]="t('clientId-tooltip')">
```
majora2007: this should be `client-id`

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+            <ng-template #view>
+              {{formControl.value}}
+            </ng-template>
+            <ng-template #edit>
+              <input id="oid-clientId" aria-describedby="oidc-clientId-validations" class="form-control"
```
majora2007: Same for inputs, please make it kebab case (`client-id`)

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+
+            </ng-template>
+          </app-setting-item>
+        }
+      </div>
+
+    </ng-container>
+
+    <div class="setting-section-break"></div>
+    <h4>{{t('behaviour')}}</h4>
```
majora2007: This should be `behavior-title` and let's use English spelling (although I do normally spell it with a u)

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+    <ng-container>
+      <div class="row g-0 mt-4 mb-4">
+        @if (settingsForm.get('providerName'); as formControl) {
+          <app-setting-item [title]="t('provider-name')" [subtitle]="t('provider-name-tooltip')">
+            <ng-template #view>
+              {{formControl.value}}
+            </ng-template>
+            <ng-template #edit>
+              <input id="oid-providerName" aria-describedby="oidc-providerName-validations" class="form-control"
```
majora2007: `provider-name`

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.html
```diff
@@ -0,0 +1,195 @@
...
+                     formControlName="providerName" type="text"
+                     [class.is-invalid]="formControl.invalid && !formControl.untouched">
+            </ng-template>
+          </app-setting-item>
+        }
+      </div>
+
+      <div class="row g-0 mt-4 mb-4">
+        @if(settingsForm.get('provisionAccounts'); as formControl) {
+          <app-setting-switch [title]="t('provision-accounts')" [subtitle]="t('provision-accounts-tooltip')">
```
majora2007: You're missing the `-label` for localization keys. You only leave the label off in existing code where I didn't yet adapt this rule.

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.ts
```diff
@@ -0,0 +1,175 @@
...
+    RoleSelectorComponent
+  ],
+  templateUrl: './manage-open-idconnect.component.html',
+  styleUrl: './manage-open-idconnect.component.scss',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ManageOpenIDConnectComponent implements OnInit {
+
+  serverSettings!: ServerSettings;
+  oidcSettings = signal<OidcConfig | undefined>(undefined);
```
majora2007: Let's keep the signals together and the non-signals together.

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.ts
```diff
@@ -0,0 +1,175 @@
...
+  settingsForm: FormGroup = new FormGroup({});
+
+  ageRatings = signal<AgeRatingDto[]>([]);
+  selectedLibraries = signal<number[]>([]);
+  selectedRoles = signal<string[]>([]);
+
+  constructor(
+    private settingsService: SettingsService,
```
majora2007: This is deprecated in Angular, please move them to injects.

API/Services/OidcService.cs
Fesaa: Yes, correct. Your library access will be granted based on the roles assigned to you in OIDC. I image most users will keep this turned off, and instead use the default permissions.

But if you're managing lots of users, across several applications, it'll be nice (I hope).

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.save();
+  }
+
+  save() {
+    if (!this.settingsForm.valid || !this.serverSettings || !this.oidcSettings) return;
+
+    const data = this.settingsForm.getRawValue();
+    const newSettings = Object.assign({}, this.serverSettings);
+    newSettings.oidcConfig = data as OidcConfig;
+    newSettings.oidcConfig.defaultAgeRestriction = parseInt(newSettings.oidcConfig.defaultAgeRestriction as unknown as string, 10) as AgeRating;
```
majora2007: remove `as unknown as string` and just `defaultAgeRestriction + ''`

UI/Web/src/app/admin/manage-users/manage-users.component.html
```diff
@@ -20,6 +21,18 @@
     <tbody>
       @for(member of members; track member.username + member.lastActiveUtc + member.roles.length; let idx = $index) {
         <tr>
+          <td>
+            <div class="-flex flex-row justify-content-center align-items-center">
```
majora2007: `d-flex`

UI/Web/src/app/admin/manage-users/manage-users.component.ts
```diff
@@ -154,4 +167,8 @@ export class ManageUsersComponent implements OnInit {
   getRoles(member: Member) {
     return member.roles.filter(item => item != 'Pleb');
   }
+
+  protected readonly UserOwner = IdentityProvider;
+  protected readonly IdentityProvider = IdentityProvider;
+  protected readonly size = size;
```
majora2007: size doesn't seem like a needed import.

UI/Web/src/app/admin/settings.service.ts
```diff
@@ -78,6 +78,11 @@ export class SettingsService {
   isValidCronExpression(val: string) {
     if (val === '' || val === undefined || val === null) return of(false);
     return this.http.get<string>(this.baseUrl + 'settings/is-valid-cron?cronExpression=' + val, TextResonse).pipe(map(d => d === 'true'));
+  }
+
+  ifValidAuthority(authority: string) {
+    if (authority === '' || authority === undefined || authority === null) return of(false);
 
+    return this.http.post<boolean>(this.baseUrl + 'oidc/is-valid-authority', {authority});
```
majora2007: `    return this.http.post<boolean>(this.baseUrl + 'oidc/is-valid-authority', {authority}, TextResponse).pipe(map(r => r + '' == 'true'));`

This will ensure a boolean comes back. By default, httpClient expects a json response.

UI/Web/src/app/registration/oidc-callback/oidc-callback.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  ],
+  templateUrl: './oidc-callback.component.html',
+  styleUrl: './oidc-callback.component.scss'
+})
+export class OidcCallbackComponent implements OnInit {
+
+  showSplash = signal(false);
+
+  constructor(
+    private accountService: AccountService,
```
majora2007: You need to use inject functions and not a constructor for DI

UI/Web/src/assets/langs/en.json
```diff
@@ -5,7 +5,56 @@
...
+        "title": "OpenID Connect Callback",
+        "login": "Back to login screen",
+        "error-loading-info": "An error occurred loading OpenID Connect info, contact your administrator",
+        "timeout": "OIDC resolution has timed out or an error has occurred"
+    },
+
+    "manage-oidc-connect": {
+        "save": "{{common.save}}",
+        "notice": "Notice",
+        "restart-required": "Changing Authority, or Client ID requires a manual restart of Kavita to take effect.",
```
majora2007: Remove ,

UI/Web/src/assets/langs/en.json
```diff
@@ -5,7 +5,56 @@
...
+        "other-field-required": "{{validation.other-field-required}}",
+        "invalid-uri": "{{validation.invalid-uri}}",
+        "manual-save-label": "Changing provider settings requires a manual save",
+
+        "authority": "Authority",
+        "authority-tooltip": "The URL to your OpenID Connect provider",
+        "clientId": "Client ID",
+        "clientId-tooltip": "The ClientID set in your OIDC provider, can be anything",
+        "provision-accounts": "Provision accounts",
+        "provision-accounts-tooltip": "Create a new account when someone logs in via OIDC, without already having an account",
```
majora2007: Maybe we can word this as `Auto-create a new account when logging in via OIDC if none exists`?

UI/Web/src/assets/langs/en.json
```diff
@@ -5,7 +5,56 @@
...
+        "require-verified-email": "Require verified emails",
+        "require-verified-email-tooltip": "Requires emails to be verified when creation an account or matching with existing ones. A newly created account with a verified email, will be auto verified on Kavita's side",
```
majora2007: `Requires email verification when creating a new account or matching with an existing one. If a newly created account has a verified email, it will be automatically verified on Kavita's side.`

UI/Web/src/assets/langs/en.json
```diff
@@ -5,7 +5,56 @@
...
+        "sync-user-settings": "Sync user settings with OIDC roles",
+        "sync-user-settings-tooltip": "Users created from OIDC will be fully managed (Roles, Library Access, Age Rating) by the OIDC. If this is disabled, users will be unable to access any content after their account creation. Read the documentation for more information.",
```
majora2007: `Users created via OIDC will be fully managed by the identity provider, including roles, library access, and age rating. If this option is disabled, users will not have access to any content after account creation. Refer to the documentation for more details.`

UI/Web/src/assets/langs/en.json
```diff
@@ -5,7 +5,56 @@
...
+        "auto-login": "Auto login",
+        "auto-login-tooltip": "Auto redirect to OpenID Connect provider when opening the login screen",
+        "disable-password-authentication": "Disable password authentication",
+        "disable-password-authentication-tooltip": "Users with the admin role can bypass this restriction",
+        "provider-name": "Provider name",
+        "provider-name-tooltip": "Name show on the login screen",
```
majora2007: `Show Provider Name on login screen`

UI/Web/src/assets/langs/en.json
```diff
@@ -28,7 +77,12 @@
         "cancel": "{{common.cancel}}",
         "saving": "Saving…",
         "update": "Update",
-        "account-detail-title": "Account Details"
+        "account-detail-title": "Account Details",
+        "notice": "Warning!",
+        "out-of-sync": "This user was created via OIDC, if the SynUsers setting is turned on changes made may be lost",
```
majora2007: `This user was created via OIDC. If the SyncUsers setting is enabled, any changes made to this account may be overwritten.`

UI/Web/src/assets/langs/en.json
```diff
@@ -28,7 +77,12 @@
...
+        "oidc-managed": "This user is managed via OIDC, contact your OIDC administrator if they require changes.",
```
majora2007: `This user is managed via OIDC. Please contact your OIDC administrator to request any changes.`

UI/Web/src/assets/langs/en.json
```diff
@@ -28,7 +77,12 @@
...
+        "identity-provider": "Identity provider",
+        "identity-provider-tooltip": "Kavita users will never be synced with OIDC"
```
majora2007: `Kavita users are never synced with OIDC accounts.`

UI/Web/src/assets/langs/en.json
```diff
@@ -2421,7 +2476,16 @@
         "invalid-password-reset-url": "Invalid reset password url",
         "delete-theme-in-use": "Theme is currently in use by at least one user, cannot delete",
         "theme-manual-upload": "There was an issue creating Theme from manual upload",
-        "theme-already-in-use": "Theme already exists by that name"
+        "theme-already-in-use": "Theme already exists by that name",
+        "oidc": {
+            "missing-external-id": "OpenID Connect provider did not return a valid identifier",
```
majora2007: Don't switch between OpenID Connect and OIDC. Let's use OIDC consistently (it's more well known).

UI/Web/src/assets/langs/en.json
```diff
@@ -2421,7 +2476,16 @@
...
+            "missing-email": "OpenID Connect provider did not return a valid email",
+            "email-not-verified": "Your email must be verified to allow logging in via OpenID Connect",
```
majora2007: `logging in` -> `login`

majora2007 (changes requested):
After these changes, I'm good to merge.

Kavita.Common/Configuration.cs
Fesaa: Can't be used with const, I can make it static readonly?

majora2007: Then it's fine. I was thinking that it might not, hence why I didn't comment last time.

UI/Web/src/assets/langs/en.json
Fesaa: I changed it to `Name shown on the login screen` as it's not a toggle, but a string

majora2007 (approved):

PR #3932, opened 2025-07-18
First OIDC fixes and feedback

Fesaa:
### Added
- Added a setting to define a custom role prefix when syncing users via OIDC  
- Added a setting to specify a custom role claim when syncing users via OIDC  
- Added support for requesting custom scopes from your OIDC provider  

### Changed
- Users will now be logged out if they fail to sync via OIDC  
- Username and email are now also synced via OIDC  
- If no Age Restriction roles are configured, users will have unrestricted access to content when syncing via OIDC  

### Fixed
- Fixed an issue where users could get stuck in a login loop when failing to log in via OIDC with auto-redirect enabled (canary)  
- Fixed the authority validator incorrectly throwing errors when a trailing slash is present (canary)  
- Fixed logout not functioning correctly when auto-redirect is enabled (canary)  
- Fixed Age Restriction roles not recognizing NotApplicable, and now properly include Unknowns (canary)  
- Fixed an issue preventing settings from being saved in the General tab (canary)

### DIFF
--- a/API.Tests/Services/OidcServiceTests.cs
+++ b/API.Tests/Services/OidcServiceTests.cs
@@ -36,7 +36,7 @@ public class OidcServiceTests: AbstractDbTest
         var claims = new List<Claim>()
         {
             new (ClaimTypes.Name, "amelia"),
-            new (ClaimTypes.GivenName, "Kraft Lawrence"),
+            new (ClaimTypes.GivenName, "Lawrence"),
         };
         var identity = new ClaimsIdentity(claims);
         var principal = new ClaimsPrincipal(identity);
@@ -50,13 +50,13 @@ public class OidcServiceTests: AbstractDbTest
         await oidcService.SyncUserSettings(null!, settings, principal, user);
         var dbUser = await UnitOfWork.UserRepository.GetUserByIdAsync(user.Id);
         Assert.NotNull(dbUser);
-        Assert.Equal("Kraft Lawrence", user.UserName);
+        Assert.Equal("Lawrence", user.UserName);
 
         claims = new List<Claim>()
         {
             new (ClaimTypes.Name, "amelia"),
-            new (ClaimTypes.GivenName, "Kraft Lawrence"),
-            new (ClaimTypes.Surname, "Norah Arendt"),
+            new (ClaimTypes.GivenName, "Lawrence"),
+            new (ClaimTypes.Surname, "Norah"),
         };
         identity = new ClaimsIdentity(claims);
         principal = new ClaimsPrincipal(identity);
@@ -65,7 +65,7 @@ public class OidcServiceTests: AbstractDbTest
         await oidcService.SyncUserSettings(null!, settings, principal, user);
         dbUser = await UnitOfWork.UserRepository.GetUserByIdAsync(user.Id);
         Assert.NotNull(dbUser);
-        Assert.Equal("Kraft Lawrence", user.UserName);
+        Assert.Equal("Lawrence", user.UserName);
     }
 
     [Fact]
--- a/API.Tests/Services/OidcServiceTests.cs
+++ b/API.Tests/Services/OidcServiceTests.cs
@@ -383,17 +383,18 @@ public class OidcServiceTests: AbstractDbTest
 
         var dbUser = await UnitOfWork.UserRepository.GetUserByIdAsync(user.Id);
         Assert.NotNull(dbUser);
-        Assert.Equal(AgeRating.RatingPending,  dbUser.AgeRestriction);
-        Assert.False(dbUser.AgeRestrictionIncludeUnknowns);
+        Assert.Equal(AgeRating.NotApplicable,  dbUser.AgeRestriction);
+        Assert.True(dbUser.AgeRestrictionIncludeUnknowns);
 
-        identity = new ClaimsIdentity([new (ClaimTypes.Role, OidcService.AgeRestrictionPrefix + OidcService.IncludeUnknowns)]);
+        // Also default to no restrictions when only include unknowns is present
+        identity = new ClaimsIdentity([new Claim(ClaimTypes.Role, OidcService.AgeRestrictionPrefix + OidcService.IncludeUnknowns)]);
         principal = new ClaimsPrincipal(identity);
 
         await oidcService.SyncUserSettings(null!, settings, principal, user);
 
         dbUser = await UnitOfWork.UserRepository.GetUserByIdAsync(user.Id);
         Assert.NotNull(dbUser);
-        Assert.Equal(AgeRating.RatingPending,  dbUser.AgeRestriction);
+        Assert.Equal(AgeRating.NotApplicable,  dbUser.AgeRestriction);
         Assert.True(dbUser.AgeRestrictionIncludeUnknowns);
     }
 
--- a/API.Tests/Services/OidcServiceTests.cs
+++ b/API.Tests/Services/OidcServiceTests.cs
@@ -23,6 +23,52 @@ public class OidcServiceTests: AbstractDbTest
 {
 
     [Fact]
+    public async Task UserSync_Username()
+    {
+        await ResetDb();
+        var (oidcService, _, _, userManager) = await Setup();
+
+        var user = new AppUserBuilder("holo", "holo@localhost").Build();
+        var res = await userManager.CreateAsync(user);
+        Assert.Empty(res.Errors);
+        Assert.True(res.Succeeded);
+
+        var claims = new List<Claim>()
+        {
+            new (ClaimTypes.Name, "amelia"),
+            new (ClaimTypes.GivenName, "Kraft Lawrence"),
+        };
+        var identity = new ClaimsIdentity(claims);
+        var principal = new ClaimsPrincipal(identity);
+
+        var settings = new OidcConfigDto
+        {
+            SyncUserSettings = true,
+        };
+
+        // name is updated as the current username is not found, amelia is skipped as it is alredy in use
+        await oidcService.SyncUserSettings(null!, settings, principal, user);
+        var dbUser = await UnitOfWork.UserRepository.GetUserByIdAsync(user.Id);
+        Assert.NotNull(dbUser);
+        Assert.Equal("Kraft Lawrence", user.UserName);
+
+        claims = new List<Claim>()
+        {
+            new (ClaimTypes.Name, "amelia"),
+            new (ClaimTypes.GivenName, "Kraft Lawrence"),
+            new (ClaimTypes.Surname, "Norah Arendt"),
+        };
+        identity = new ClaimsIdentity(claims);
+        principal = new ClaimsPrincipal(identity);
+
+        // Ensure a name longer down the list isn't picked if the current username is found
+        await oidcService.SyncUserSettings(null!, settings, principal, user);
+        dbUser = await UnitOfWork.UserRepository.GetUserByIdAsync(user.Id);
+        Assert.NotNull(dbUser);
+        Assert.Equal("Kraft Lawrence", user.UserName);
+    }
+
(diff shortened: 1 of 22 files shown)


API/DTOs/Settings/OidcConfigDto.cs
```diff
@@ -22,6 +23,14 @@ public sealed record OidcConfigDto: OidcPublicConfigDto
     /// Overwrite Kavita roles, libraries and age rating with OpenIDConnect provided roles on log in.
     /// </summary>
     public bool SyncUserSettings { get; set; }
+    /// <summary>
+    /// A prefix that all roles Kavita check for during sync must have
```
Fesaa: typo checks for

API/I18N/en.json
```diff
@@ -45,6 +45,8 @@
     "email-not-enabled": "Email is not enabled on this server. You cannot perform this action.",
     "account-email-invalid": "The email on file for the admin account is not a valid email. Cannot send test email.",
     "email-settings-invalid": "Email settings missing information. Ensure all email settings are saved.",
+    "oidc-invalid-authority": "Oidc Authority is not valid",
```
majora2007: Let's keep OUDC capital across all strings

API/Services/OidcService.cs
```diff
@@ -125,31 +139,35 @@ public async Task ClearOidcIds()
...
 
         return null;
     }
 
     private async Task<bool> IsNameAvailable(string? name)
     {
-        if (string.IsNullOrEmpty(name)) return false;
-
-        return await userManager.FindByNameAsync(name) == null;
+        return !(await accountService.ValidateUsername(name)).Any();
```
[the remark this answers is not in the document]
Fesaa: The ValidateUsername method does this too

API.Tests/Services/OidcServiceTests.cs
```diff
@@ -41,7 +193,7 @@ public async Task SyncRoles()
             SyncUserSettings = true,
         };
 
-        await oidcService.SyncUserSettings(settings, principal, user);
+        await oidcService.SyncUserSettings(null!, settings, principal, user);
```
majora2007, line 196: why are we supressing the null when it's explicit?

API/Services/OidcService.cs
```diff
@@ -76,6 +81,13 @@ public class OidcService(ILogger<OidcService> logger, UserManager<AppUser> userM
         user = await unitOfWork.UserRepository.GetUserByEmailAsync(email, AppUserIncludes.UserPreferences | AppUserIncludes.SideNavStreams);
         if (user != null)
         {
+            // Don't allow taking over accounts
+            // This could happen if the user changes their email in OIDC, and then someone else uses the old one
+            if (user.OidcId is not (null or ""))
```
majora2007: Use string.Empty

API/Services/OidcService.cs
```diff
@@ -239,25 +259,111 @@ public async Task SyncUserSettings(OidcConfigDto settings, ClaimsPrincipal claim
...
-        var roles = claimsPrincipal.GetAccessRoles();
+        var email = claimsPrincipal.FindFirstValue(ClaimTypes.Email);
+        if (string.IsNullOrEmpty(email) || user.Email == email) return;
+
+        if (settings.RequireVerifiedEmail && !claimsPrincipal.HasVerifiedEmail())
+        {
+            throw new KavitaException("errors.oidc.email-not-verified");
+        }
+
+        // Ensure no other user, uses this email
```
majora2007: remove comma, doesn't make sense

API/Services/OidcService.cs
```diff
@@ -239,25 +259,111 @@ public async Task SyncUserSettings(OidcConfigDto settings, ClaimsPrincipal claim
...
+        var isValidEmailAddress = !string.IsNullOrEmpty(user.Email) && emailService.IsValidEmail(user.Email);
+        var isEmailSetup = (await unitOfWork.SettingsRepository.GetSettingsDtoAsync()).IsEmailSetup();
+        var shouldEmailUser = isEmailSetup || !isValidEmailAddress;
+
+        user.EmailConfirmed = !shouldEmailUser;
+        user.ConfirmationToken = token;
+        await userManager.UpdateAsync(user);
+
+        var emailLink = await emailService.GenerateEmailLink(request, user.ConfirmationToken, "confirm-email-update", email);
+        logger.LogCritical("[Update Email]: Email Link for {UserName}: {Link}", user.UserName, emailLink);
```
majora2007: Logger could have a more clear message for OIDC Confirmation and worded as maybe "Required"

API/Services/OidcService.cs
```diff
@@ -279,14 +385,20 @@ private async Task SyncAgeRestriction(ClaimsPrincipal claimsPrincipal, AppUser u
...
-            .Select(r => r.Value.TrimPrefix(AgeRestrictionPrefix))
-            .ToList();
+        var ageRatingPrefix = settings.RolesPrefix + AgeRestrictionPrefix;
+        var ageRatings = claimsPrincipal.GetClaimsWithPrefix(settings.RolesClaim, ageRatingPrefix);
         logger.LogDebug("Syncing age restriction for user {UserName}, found restrictions {Restrictions}", user.UserName, ageRatings);
 
-        var highestAgeRating = AgeRating.Unknown;
+        if (ageRatings.Count == 0 || (ageRatings.Count == 1 && ageRatings.Contains(IncludeUnknowns)))
+        {
+            logger.LogDebug("No age restriction found in roles, setting to RatingPending");
```
majora2007, line 394: Why are we defaulting to an age restriction here?

API/Services/SettingsService.cs
```diff
@@ -363,7 +364,8 @@ public async Task<bool> IsValidAuthority(string authority)
             return false;
         }
 
-        var url = authority + "/.well-known/openid-configuration";
+        var hasTrailingSlash = authority.EndsWith('/');
+        var url = authority + (hasTrailingSlash ? "" : "/") + ".well-known/openid-configuration";
```
majora2007: Use string.Empty whenever possible.

UI/Web/src/app/_resolvers/oidc.resolver.ts
```diff
@@ -19,7 +20,7 @@ export class OidcResolver implements Resolve<any> {
       timeout(5000),
       catchError(err => {
         console.log(err);
-        this.toastR.error("oidc.timeout");
+        this.toastR.error(translate("oidc.timeout"));
```
majora2007: refactor inject as toastr please.

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.ts
```diff
@@ -126,20 +135,35 @@ export class ManageOpenIDConnectComponent implements OnInit {
...
       error: error => {
         console.error(error);
-        this.toastr.error("errors.generic")
+        this.toastr.error(translate('errors.generic'))
       }
     })
   }
 
+  breakString(s: string) {
+    if (s) {
```
majora2007, line 160: this can easily be `(s || '').split(',')`.

UI/Web/src/app/registration/user-login/user-login.component.ts
```diff
@@ -82,7 +82,8 @@ export class UserLoginComponent implements OnInit {
     effect(() => {
       const skipAutoLogin = this.skipAutoLogin();
       const oidcConfig = this.oidcService.settings();
-      if (!oidcConfig || skipAutoLogin === undefined) return;
+      const isLoggingOut = this.oidcService.isLoggingOut();
+      if (!oidcConfig || skipAutoLogin === undefined || isLoggingOut) return;
```
majora2007, line 87: newline above if statement since multiple const assignments, will help quickly jump sections.

UI/Web/src/assets/langs/en.json
```diff
@@ -49,7 +51,16 @@
         "default-include-unknowns-tooltip": "Include unknown age ratings",
         "default-age-restriction-label": "Age rating",
         "default-age-restriction-tooltip": "Maximum age rating shown to new users",
-        "no-restriction": "{{restriction-selector.no-restriction}}"
+        "no-restriction": "{{restriction-selector.no-restriction}}",
+
+        "advanced-title": "Advanced settings",
+        "advanced-tooltip": "Only modify the following options if you know what you're doing",
```
majora2007: Only modify these options if you understand their impact.

UI/Web/src/assets/langs/en.json
```diff
@@ -49,7 +51,16 @@
...
+        "roles-prefix-label": "Roles prefix",
+        "roles-prefix-tooltip": "Kavita will filter out roles which do not have the prefix",
```
majora2007: Kavita will only consider roles that start with this prefix.

UI/Web/src/assets/langs/en.json
```diff
@@ -49,7 +51,16 @@
...
+        "roles-claim-label": "Roles claim",
+        "roles-claim-tooltip": "The claim under which Kavita must look for your roles, leave empty to reset",
```
majora2007: The claim Kavita will use to extract roles. Leave blank to use the default.

UI/Web/src/assets/langs/en.json
```diff
@@ -49,7 +51,16 @@
...
+        "custom-scopes-label": "Custom scopes",
+        "custom-scopes-tooltip": "Additional scopes Kavita will request from your OIDC when trying to login, separate by comma. If these are configured wrong, your OIDC may prevent you from login in."
```
majora2007: Additional scopes to request from your OIDC provider during login. Separate multiple scopes with commas. Incorrect values may prevent login.

majora2007 (changes requested):
Overall looks good, just a few comments.

API.Tests/Services/OidcServiceTests.cs
Fesaa: It's a HttpRequest object that's only used to construct the verification link in the email, so is safe to be null. I added the ! as otherwise, Rider would make them all yellow and add a warning

API/Services/OidcService.cs
Fesaa: This isn't possible, these pattern matches require constant values

API/Services/OidcService.cs
Fesaa: Just a random choice mostly, I'm not sure what the default behaviour is elsewhere. From messages in the discord thread; it might be nicer to have it be NotApplicable and include unknowns by default, so there is less setup needing unless you want to set one. 

Thoughts? 

```c#
user.AgeRestriction = AgeRating.NotApplicable;
user.AgeRestrictionIncludeUnknowns = ageRatings.Count == 0 || ageRatings.Contains(IncludeUnknowns);
```

API/Services/OidcService.cs
majora2007: Then replace this with `string.IsNullOrEmpty(user.OidcId)` which is preferred.

API/Services/OidcService.cs
majora2007: I'm reading this as there are no explicit roles on Age Rating, hence we shouldn't default to one. We should set to NotApplicable and include unknowns so they have an unfiltered account access, which is what I would expect.

UI/Web/src/app/admin/manage-open-idconnect/manage-open-idconnect.component.ts
Fesaa: If you split an empty string it'll return a list with the empty string as element, while we need it to be an empty list instead here. I can change it to filter empty strings out afterwards, but that's the same code in a different order. Which would be preferred?

majora2007 (approved):

PR #3933, opened 2025-07-19
Update the PDF Metadata Parser to Accept Lengths By ObjectRef

zeroSteiner:
# Changed
- Changed: Changed PDF parser exception messages to state what type was received in addition to what was expected

# Fixed
- Fixed: Fixed an issue with the PDF parser that prevented it from loading files that used a ObjectRef for the Length

Thanks!

### DIFF
--- a/API/Helpers/PdfMetadataExtractor.cs
+++ b/API/Helpers/PdfMetadataExtractor.cs
@@ -954,7 +954,7 @@ internal class PdfMetadataExtractor : IPdfMetadataExtractor
                 case "Length":
                     if (value.Type != PdfLexer.TokenType.Int)
                     {
-                        throw new PdfMetadataExtractorException("Expected integer after /Length");
+                        throw new PdfMetadataExtractorException($"Expected integer after /Length, but got {value.Type}");
                     }
 
                     length = (long)value.Value;
@@ -1361,7 +1361,7 @@ internal class PdfMetadataExtractor : IPdfMetadataExtractor
                 case "Metadata":
                     if (value.Type != PdfLexer.TokenType.ObjectRef)
                     {
-                        throw new PdfMetadataExtractorException("Expected object number after /Metadata");
+                        throw new PdfMetadataExtractorException($"Expected object number after /Metadata, but got {value.Type}");
                     }
 
                     meta = (long)value.Value;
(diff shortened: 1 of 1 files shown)


majora2007:
Can you explain a bit more about the fix and is there a way to write a unit test to trigger?

zeroSteiner:
> Can you explain a bit more about the fix 

Sure thing, I hope this helps. So the error I ran into was while importing multiple files in my Library. I tracked down the issue to the parser receiving an ObjectRef when it was expecting an int. To address it, I referred to some that was already in the file, specifically these two locations:

* https://github.com/Kareadita/Kavita/blob/152f7ad00e47c8a91beed621eee2e08040393ac5/API/Helpers/PdfMetadataExtractor.cs#L1297-L1299
* https://github.com/Kareadita/Kavita/blob/152f7ad00e47c8a91beed621eee2e08040393ac5/API/Helpers/PdfMetadataExtractor.cs#L1317-L1334

I implemented a private method that would use the existing stream to read the object reference using a temporary `Lexer` instance. Using the temporary instance ensured that the state of the primary `_lexer` is maintained since I didn't see a way to back up and restore it. Once the object is resolved, it's checked again to see if it's an integer and then things carry on as they did before.

As for the other log messages, I added their types because it was really useful while debugging to see not only what it was expecting but what it received instead. It's possible that there are other areas that could benefit from resolving ObjectRefs, but without PDF test cases on hand, I only updated the processing of the Length field.

> and is there a way to write a unit test to trigger?

That might take me a little while. I don't see any existing tests referring to the `PdfMetadataExtractor` class so I'd need to do it from scratch I think and I'm not super familiar with .NET unit tests. Is there anything you could point me to as a good example or a place for me to get started?

majora2007:
`BookServiceTests.SeriesFallBackToMetadataTitle` might be a good starting point. It uses a pdf packed with the repo to test. You can also use https://github.com/Kareadita/Kavita/pull/3912/files#diff-416981523a2401e533a3b63d10f5153322f77bee20d25af37521780073f7f778 which has unit tests around expecting the code to throw. 

It would be helpful if you could also send me (DM/email) a file that caused this issue, so I can test it out on my end as well. 

Feel free to ask for help here or discord.

majora2007:
Spoke with the dev and got the file (Murder Die Senpai 1-2.pdf). I'll be looking into reproducing and setting up unit tests later in the cycle. Thanks again @zeroSteiner for the initial contribution.

PR #3985, opened 2025-08-07
A few bugs from last nightly

Fesaa:
# Fixed
- Fixed: Fixed not being able to go past the first step in the import mappings flow (develop)
- Fixed: Fixed reading profiles, only saving the first change (develop)
- Fixed: Fixed the first page in the manga reader not updating/showing (develop)


Fesaa:
I've validated the second commit fixed the behaviour I saw and was trying to fix with it. But I can't change the commit message, as I verify my commits. (Without re-doing and that was too much effort)

majora2007 (approved):

PR #3994, opened 2025-08-14
Series/Volume/Chapter detail pages can now have title be customizable via Themes

YodaDaCoda:
# Theme
- Added CSS var for series detail title color, allowing themes to configure the color. Falls back to the default of white if color not supplied in theme.


majora2007:
Please open a PR to the wiki and add your new css property.

Issue #4010, opened 2025-08-31
Library Add and Bulk Add Buttons are overlayed


DoodleCup:
### What happened?

 I see overlayed buttons on the top right for library add. Looks like its positioned absolute. When i change the css rule "right" then it moves. I have changed ui to german. so the one box is longer cause of the longer text.

<img width="1857" height="203" alt="Image" src="https://github.com/user-attachments/assets/6e005bd4-c129-4b3c-94f3-6d20d5f1cf4a" />

<img width="297" height="60" alt="Image" src="https://github.com/user-attachments/assets/29eeb488-d16b-40ab-a9dc-b291a0fceeed" />

### What did you expect?

No overlayed Buttons

### Kavita Version Number - If you don't see your version number listed, please update Kavita and see if your issue still persists.

0.8.7.13

### What operating system is Kavita being hosted from?

Linux

### If the issue is being seen on Desktop, what OS are you running where you see the issue?

Windows 11

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### If the issue is being seen on Mobile, what OS are you running where you see the issue?

None

### Relevant log output

```shell

```

majora2007:
Thanks, this is a hard to solve problem with non-english locales. Ive seen it before, will look again for a good solution.

PR #4015, opened 2025-09-01
Fix library buttons positioning (#4010)

fpammer:
Prevent library buttons from running into each other by removing the absolute positioning for the second button and flexing them: 

image

Old: 

image

I searched for other parts in the code that had the same css classes and noticed one definition that was not used. 
Also removed the reduced padding for the `card-actionables.component` so that the two buttons are same sized. I checked the app for obvious breakage by that change and didn't find any.


jagadam97:
#4010

UI/Web/src/app/_single-module/card-actionables/card-actionables.component.scss
```diff
@@ -42,7 +42,3 @@
     float: right;
     padding: var(--bs-dropdown-item-padding-y) 0;
 }
-
-.btn {
```
majora2007, line 46: Why are we removing this?

majora2007 (commented):
Would you mind applying this fix to the other pages?

UI/Web/src/app/_single-module/card-actionables/card-actionables.component.scss
fpammer: If you check the screenshot closely you see that the `card-actionables` button is 2px smaller in height compared to the actual button. This is because the button has a default padding of 6px, this was for some reason overwritten in this component. We can pop this commit if you feel this change might be iffy but on a quick walk through the UI i didn't notice any negative consequences.

fpammer:
> Would you mind applying this fix to the other pages?

I checked for other usages of the `.custom-position-2` class and it was only in use here. It was also in the license section but only defined and unused. If I missed any other places where there are two action buttons like that I can fix them but I didn't find any other.

UI/Web/src/app/_single-module/card-actionables/card-actionables.component.scss
majora2007: I'll take a look and see if there are any issues.

majora2007 (approved):

majora2007:
Thanks, looks good. I merged into our active branch.

PR #4025, opened 2025-09-14
Added lazy-loading for PDFs in the PDF reader component

### LINKED ISSUE #3993: PDF Reader Service is Very Slow When Loading a 40MB Book
### Discussed in https://github.com/Kareadita/Kavita/discussions/3990

<div type='discussions-op-text'>

<sup>Originally posted by **pmvamsi-coder** August 13, 2025</sup>
### Idea Description

I’m using the Kavitha PDF reader service hosted on my own server, and I’ve noticed significant slowness when loading a 40 MB PDF book. The first page takes a long time to render, and during this time, the reader appears unresponsive.

Has anyone else experienced similar performance issues with large PDFs?

Is there a recommended configuration, caching mechanism, or lazy-loading approach that could improve initial load times?

Thanks!

vinilicz:
# Changed
- Changed: PDFs are now downloaded on demand in the PDF reader, helping with initial load times for large PDFs (Fixes #3993)

# Example 
On a PDF with 227 pages (~140 MB), from page 207 to 227:
Before: 2136 requests, 54 MB transferred, 141 MB resources
After: 84 requests, 10 MB transferred, 13 MB resources

https://github.com/user-attachments/assets/004950fa-47b0-4429-b812-993ea27e4b7e  

https://github.com/user-attachments/assets/0104bc7c-a96e-46c2-89b5-a6cef2558334

### DIFF
--- a/UI/Web/src/app/pdf-reader/_components/pdf-reader/pdf-reader.component.html
+++ b/UI/Web/src/app/pdf-reader/_components/pdf-reader/pdf-reader.component.html
@@ -2,23 +2,17 @@
   @if (accountService.currentUser$ | async; as user) {
     <div class="{{theme}}" #container>
 
-      @if (isLoading) {
-        <div class="loading mx-auto" style="min-width: 200px; width: 600px;">
-          <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
-          {{t('loading-message')}}
+      <div class="progress-container row g-0 align-items-center">
+        <div class="progress" style="height: 5px;">
+          <div class="progress-bar" role="progressbar" [ngStyle]="{'width': loadPercent + '%'}" [attr.aria-valuenow]="loadPercent" aria-valuemin="0" aria-valuemax="100"></div>
         </div>
-        <div class="progress-container row g-0 align-items-center">
-          <div class="progress" style="height: 5px;">
-            <div class="progress-bar" role="progressbar" [ngStyle]="{'width': loadPercent + '%'}" [attr.aria-valuenow]="loadPercent" aria-valuemin="0" aria-valuemax="100"></div>
-          </div>
-        </div>
-      }
+      </div>
 
       <ngx-extended-pdf-viewer
         #pdfViewer
         [src]="readerService.downloadPdf(this.chapterId)"
         [authorization]="'Bearer ' + user.token"
-        height="100vh"
+        height="calc(100vh - 5px)"
         [(page)]="currentPage"
         [textLayer]="true"
         [showHandToolButton]="true"
--- a/UI/Web/src/app/pdf-reader/_components/pdf-reader/pdf-reader.component.html
+++ b/UI/Web/src/app/pdf-reader/_components/pdf-reader/pdf-reader.component.html
@@ -2,17 +2,23 @@
   @if (accountService.currentUser$ | async; as user) {
     <div class="{{theme}}" #container>
 
-      <div class="progress-container row g-0 align-items-center">
-        <div class="progress" style="height: 5px;">
-          <div class="progress-bar" role="progressbar" [ngStyle]="{'width': loadPercent + '%'}" [attr.aria-valuenow]="loadPercent" aria-valuemin="0" aria-valuemax="100"></div>
+      @if (isLoading) {
+        <div class="loading-message-container">
+          <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
+          {{t('loading-message')}}
         </div>
-      </div>
+        <div class="progress-container row g-0 align-items-center">
+          <div class="progress" style="height: 5px;">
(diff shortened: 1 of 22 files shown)


UI/Web/src/app/pdf-reader/_components/pdf-reader/pdf-reader.component.html
```diff
@@ -2,23 +2,17 @@
   @if (accountService.currentUser$ | async; as user) {
     <div class="{{theme}}" #container>
 
-      @if (isLoading) {
-        <div class="loading mx-auto" style="min-width: 200px; width: 600px;">
-          <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
-          {{t('loading-message')}}
+      <div class="progress-container row g-0 align-items-center">
+        <div class="progress" style="height: 5px;">
+          <div class="progress-bar" role="progressbar" [ngStyle]="{'width': loadPercent + '%'}" [attr.aria-valuenow]="loadPercent" aria-valuemin="0" aria-valuemax="100"></div>
```
majora2007: The problem with this is it leaves the progress bar once it's at 100%. Let's remove the bar after loading is done and leave an `<span class="visually-hidden">{{t('loading-message')}}</span> here as well. 

Let's tweak the message in en.json to "loading..."

majora2007 (changes requested):

UI/Web/src/app/pdf-reader/_components/pdf-reader/pdf-reader.component.html
vinilicz: Done.

Reverted loading message and progress bar visibility changes.
Centered loading message text for better UX.
Simplified loading message translations to basic "Loading..." format.

After:  
after

majora2007:
Just for future reference, you only need to update the en.json file.

majora2007:
Just FYI, we ended up undoing this. The loading bar would get stuck indefinitely for some users and we weren't able to reproduce improved loading conditions.

vinilicz:
Wow, I'm very sad to hear…

But just out of curiosity, is it a bug or a misunderstanding? Since it’s possible for the bar to remain visible when the document isn’t fully scrolled (starting from a later page) or searched with Ctrl + F, which is the expected behavior, I believe. The hook that hides the bar when the download is complete is the same as before (from ngx-extended-pdf-viewer)…

In my first commit (before the revision), I had actually hidden the loading message, since it could be annoying, and kept only the progress (inspired by how Calibre Web handles it). But it was reverted based on feedback.

As for reproducing the issue, it’s pretty simple: just open a large PDF (e.g., 100MB+) and check the network tab. Only the scrolled sections will be downloaded, not the entire document (also inspired by Calibre Web).

I think this feature is very important for mobile or limited networks when you have a large PDF.

majora2007:
@Fesaa can you explain what you saw when validating

Fesaa:
Heya,

What are you seeing in your network tab is correct, with the option disabled less bandwidth is used after opening the PDF (as fewer parts of the PDF are loaded). 

However, PDF.js does not require the entire PDF to be loaded before displaying it. With it enabled, it will load the parts you're currently viewing, and then load the rest of the PDF in the background such that when scrolling there's no delay. 
    You can see this in your video, there is a viewable PDF behind your network tab before the network requests have finished.

If the option is disabled, it will still load the parts you're currently viewing. And instead just stop once this is done. 

While testing this, with and without bandwidth limited, I did not see a difference in load time before I was able to view and browse the PDF. I believe we'll need to work on a different solution if we're seeing longer initial load times for PDFs larger. (I haven't looked at that yet)

The documentation on the options
> Caveat: this feature seems to be broken.
Disable pre-fetching of PDF file data. When range requests are enabled PDF.js will automatically keep fetching more data even if it isn't needed to display the current page. The default value is false. NOTE: It is also necessary to disable streaming, see below, in order for disabling of pre-fetching to work correctly.

Sadly, PDF.js does not give us an event for when this pre-fetching stops. This is what we'd need to properly support disabling it together with having the load bar work correctly.

vinilicz:
Thanks for the feedback.

But the solution I propose is not to increase the initial load speed, the section mentioning it in my first message was edited and not written by me. Maybe I didn’t make it clear in my first Discord interactions... my bad.

My proposed solution is just to avoid downloading the document unnecessarily. For example, in my country, mobile networks have data quotas, and exceeding them can be very expensive. I can't fully migrate from Calibre Web to Kavita because of this feature. I know this is my use case... but I just want to give real-world example feedback too.

As for the pre-fetching stop event, it's really a problem. Perhaps a debounce time with the progress event to auto-hide the bar could help.

Anyway, thanks again for the feedback.

Fesaa:
Heya,

Yes, looks like we misunderstood why you wanted this added. I've added a Data Saver user preference that you can enable to disable the PDF pre-fetching, as well as other such behaviour in the future were it to be added.

PR #4067, opened 2025-10-01
Fix text bleeding on column layout

### LINKED ISSUE #4040: [Bug Bounty: $200] Epub Text Bleeding on Column Layout
### Overview
When a user is in 1 Column or 2 Column layout, during paging forward/backward, occasionally text can bleed between pages. 

Image

> Text bleeding on the left

Image

> Text cuttoff on right side

Image

> Vertical Layout mode showing cutoff on right

### What we know already
We have been debugging this and believe `pageWidth()` and `columnGapModifier()` are primary targets for the issue. (book-reader.component.ts). 

**Only 1 calculation method is needed, can remove the checks**
```ts
pageWidth = computed(() => {
    this.windowWidth(); // Ensure re-compute when windows size changes (element clientWidth isn't a signal)
    this.pageCalcMode();

    console.log('page width recalulated')
    const calculationMethod = this.pageCalcMode();
    const marginLeft = this.pageStyles()['margin-left'];
    const columnGapModifier = this.columnGapModifier();
    if (this.readingSectionElemRef == null) return 0;

    const margin = (this.convertVwToPx(parseInt(marginLeft, 10)) * 2);

    // console.log('page size calc, client width: ', this.readingSectionElemRef.nativeElement.clientWidth)
    // console.log('page size calc, margin: ', margin)
    // console.log('page size calc, col gap: ', ((COLUMN_GAP / 2) * columnGapModifier));
    // console.log("clientWidth", this.readingSectionElemRef.nativeElement.clientWidth, "window", window.innerWidth, "margin", margin, "left", marginLeft)
    // console.log('clientWidth: ', this.readingSectionElemRef.nativeElement.clientWidth, 'offsetWidth:', this.readingSectionElemRef.nativeElement.offsetWidth, 'bbox:', this.readingSectionElemRef.nativeElement.getBoundingClientRect().width);

    if (calculationMethod === EpubPageCalculationMethod.Default) {
      return this.readingSectionElemRef.nativeElement.clientWidth - margin + (((COLUMN_GAP) * columnGapModifier));
    } else {
      return this.readingSectionElemRef.nativeElement.clientWidth - margin + (((COLUMN_GAP) * columnGapModifier) + 10);
    }
  });

  columnGapModifier = computed(() => {
    const calculationMethod = this.pageCalcMode();
    switch(this.layoutMode()) {
      case BookPageLayoutMode.Default:
        return 0;
      case BookPageLayoutMode.Column1:
        return 1;
      case BookPageLayoutMode.Column2:
        return calculationMethod === EpubPageCalculationMethod.Default ? 1 : 1.25;
    }
  });
```

### Payout
Payout will be done via [Open Collective](https://opencollective.com/kavita) to the amount of $200 for the first person to submit a full PR fixing this issue reliably. 

### Files
https://we.tl/t-SqkgmLl12u (if expired, ping me and Ill replace)

privatestefans:
# Fixed
- Fixed: Fix column mode page calculation that cause text bleeding (Fixes #4040)

## What was changed
- Use `outline` instead of `border` for visually debug elements to prevent any layout shifts
- Obtain the `offsetWidth` from the reader container rather than the reading section element to get the actual page width

## Preview after this fix
https://github.com/user-attachments/assets/b9a7f765-e203-4224-bf84-fb9d1bba039e

/claim #4040


privatestefans:
I can't see any issue with the vertical writing style for the Japanese version.
Even after adjusting the margin, it seems already correct. 🤔

brave_jQxNyG51PF

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.scss
```diff
@@ -256,6 +256,7 @@ $action-bar-height: 38px;
     overflow: hidden;
     word-break: break-word;
     overflow-wrap: break-word;
+    padding: 0 12px; // Add padding, so we can see the page border line on the left & right
```
majora2007: The page border is only visible on debug mode (aka for local development), so this needs to be encapsulated.

majora2007:
It seems you built off my latest attempt, which has received good feedback from testing. I'll pull this down and validate and get this tested out. 

The difference seems to mainly be switching from rounding to flooring and the padding logic, which needs to be in the .debug class, as that is not visible for the non-local build (switch the flag to validate).

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.scss
privatestefans: Ah, I see. If we don't need the page border to be visible, then we don't need this extra padding to prevent the text from being too close to the edge and slightly cut off. I will update it, thanks for the review 👍

privatestefans:
I decided to remove the extra padding and calculation for now, as `column-rule: 2px dashed #666666;` is no longer add additional 2px for each column when debug flag is disabled.

### Updated preview
https://github.com/user-attachments/assets/ede1c434-4116-4625-b0fb-cb983770c6a0

majora2007:
Did some testing on this and it's preforming slightly worse than my attempt for single column layout.

image

image

This is only reproducible on this book (my attempt also had some slight page cutoff). 

page 14, single column, 15% margin, 1025x908.

https://wormhole.app/m9Oae3#hbOPG13nTuJsfGnPh0KZzA


I would say this is almost there.

privatestefans:
Thanks for the sample file and some details.

After trying to reproduce the issue by changing the screen width, it seems the page margin did not update when the window resolution changed

I think we can fix it by reapplying page style with:
```js
this.applyPageStyles(this.pageStyles())
```

I can also reproduce bug in vertical layout during screen resizing. It occurs only when the window size changes and does not persist after a browser refresh. But I will investigate it further.

privatestefans:
Ref: https://github.com/Kareadita/Kavita/pull/4067/commits/87a64e9b3b60846dfcc97ca1d91e5e580aa40e4f
I have improved the `getFirstVisibleElementXPath` function to filter out empty elements and child elements that appear in the same entry list before returning the intended first visible element. I decided to improve this function because sometimes it returns an incorrect element (as you can see from the image below), and causing the page to snap to that element instead of the actual content or paragraph.

---
eTt462FoXp_scale

majora2007:
Nice find, I do believe this will be an improvement. I've seen you're putting up a few commits. Just ping me when ready for me to retest.

privatestefans:
Sure thing!
I'm still trying to find another potential bug, but it hasn't shown up yet after a few adjustments I made.

privatestefans:
Hi @majora2007, can you review this PR?

I’ve tested by switching between mobile and desktop views using the browser’s developer tools, and by changing the page layout from scroll mode to one- and two-column layout mode. So far, I haven’t noticed any remaining bugs during my testing.

I also made an attempt to improve the **Tap Pagination** button, especially for vertical writing style page:

<details>
<summary>Before the fix</summary>

https://github.com/user-attachments/assets/09bcc7aa-3427-47d5-b68b-c7b53262b3ff

</details>

<details>
<summary>After the fix</summary>

https://github.com/user-attachments/assets/441e08ea-ca89-4c27-b650-47612cfc48f4

</details>

majora2007:
I like the new debug viewport mode, I'll pull it down today/tomorrow, if it passes my shakeout, I'll deploy this out to a dedicated environment and have a few testers validate.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -1741,28 +1771,94 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
       .filter(entry => {
         //return this.isPartiallyContainedIn(container, entry);
         return this.utilityService.isInViewport(entry, this.topOffset);
+      })
+      .filter((element, i, entries) => {
+        // Remove any children element contained in another element that exist on this entries
+        return !entries.some(item => element !== item && item.contains(element))
+
+        // Remove element that don't have any content
+        && (element.textContent.trim().length !== 0 || element.querySelectorAll('img, svg').length !== 0 || /^(img|svg)$/im.test(element.tagName));
```
Fesaa: `element.textContent.trim()` needs a nullcheck for `element.textContent`. (Throwns an error during compile)

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -1485,6 +1510,12 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   private addEmptyPageIfRequired(): void {
+    const bookContentElem = this.bookContentElemRef.nativeElement;
+
+    // Remove the empty gap first (if exist) to recalculate if we need the empty gap
+    const oldEmptyGap = bookContentElem.querySelector('.kavita-empty-gap');
```
Fesaa, line 1513: I'm seeing some weird behaviour when paging backwards where the empty page is removed, then added again. But since I'm on the last page, it "scrolls" back, and the empty page causes you to need to page over the same text twice.

https://github.com/user-attachments/assets/60af97c7-2e3b-4bf4-8ef6-0d72b9ceb333

I'm not sure how we would best solve this. Do the calcs by subtracting the width of the empty page if present, and then remove/rescale as needed?

Fesaa:
Heya,

Left some comments on issues we've found. Overall, looks really promising! If you could have a look at them (only one really, the other is a small change) and merge in develop, so your branch is up-to-date. After which we'll deploy and have other testers test as well.

During testing, we've also noticed the following issues, which we'll have a look at as well.
- Bookmark indicators not being visible (regressing somewhere)
- Invalid XPath exception when loading a page with only an image

Thanks!

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
privatestefans: Ah, I will figure out the best way to fix that. I think doing some calcs can solve it.
Thanks for the feedback, by the way 👍

privatestefans:
Sure, I will look into it and I’m still figuring out the empty page issue at the moment.

It seems like the bookmark indicator has shifted to another page.
I'm not sure what’s causing this yet, but I might be able to check it out.

Thanks @Fesaa!

privatestefans:
About the bookmark indicator.

I noticed that some image elements have their `position` overridden by a non-existent CSS variable. As a result, the element `position` default to 'static', which causes all bookmark attached the image elements with `.kavita-scale-width-container` become incorrectly positioned.

q9q8G4Z2sh

So I set `--book-reader-content-position` to 'relative' as the default for all image elements, except for the single-page element, since I think there’s no issue with that one.

Fesaa:
Heya,

To keep you in the loop, the empty page issue is resolved. So your code has been deployed to have testers test it. :tada: 
I've merged develop in before deploying, could you also do this for your branch. Thanks

Bookmarks / images are not, I'm seeing some rather odd behaviour. But I do not believe it's caused by your changes. I will be looking into those. 

Thanks again!

privatestefans:
Thanks for letting me know. I've synced my branch with develop.

Also, I added a default value for `--book-reader-content-position` in my last commit. I haven't tested it with any book other than the one I have, though.

But feel free to ask me to make changes if needed 👍

majora2007:
Doing some extended testing on this and facing a few edge cases. I tested a few books, only this ravens of the inner palace (image heavy) had issues. I can work on the bookmarking positioning, but the image cutoff and next page seem like regressions. 

Overall, from a text perspective, this seems to fix the epub bleed. If we can cleanup these small regressions, we can close this out. 

## Ravens of the Inner Palace

#### Next page broken
image
> I cannot move to the next page. `currentVirtualPage` is always 1 and the `totalVirtualPages` is 2.
> Page is also laying out incorrectly. It should be centered with 1 column, not right aligned as if in 2 column layout. (works correctly in 2 column and scroll)


#### Image cutoff
image

image

> Images are being cutoff, this seems to be due to `--book-reader-content-max-height` on the surrounding image container being too high (532px on mine, bringing it down to near `55px - 38px` fixes the layout. 


#### Bookmark icon cutoff
image

> The bookmark icon calculation needs some slight tweaking. It seems to be a few pixels off which is causing some cutoff. 

image

> 2 column layout also has some weird cutoff issue that is fixed with even more offset. (I fixed with a -10 on top/left)

privatestefans:
Whoops, looks like using `relative` as the default value for `book-reader-content-position` for fixing the bookmark is not a good idea for that book. I've reverted it and also added a small improvement for filtering element.

Can you recheck it @majora2007?

majora2007:
That fixed all my issues. It's looking great, I'll do another shakeout with some more books to validate, but we might be done with this issue.

majora2007:
Great news, everything checked out. We had found some small strangeness with japanese books (scroll position shifting between layout mode). This is something we are willing to accept as an edge case, likely just better `firstVisibleItem` code in the future.

Please submit an expense against OC for 200$ USD.

privatestefans:
I’m glad to hear the fix worked well! Thank you for reviewing it carefully. I agree `getFirstVisibleElement` could be refined. I also spent some time looking into that, but I was careful not to over-filter since it might not work for some books.

I've just submitted the expense through OpenCollective, thank you!

majora2007:
Agreed, it's a balancing act when books have the ability to put any styles and html in them. I just approved the OC, thank you again.

Issue #999, opened 2022-01-27
Implement a migration for accounts that already exist to get an email associated


majora2007:
When user logs in and we fail them due to not having confirmed email, let the login prompt them by checking if:
user.EmailConfirmed && string.IsEmptyOrNull(user.Email).

In UI, we will let them assign an email one time and go through confirm flow. 

This has security issues, but might be the easiest way to migrate the users.

PR #4070, opened 2025-10-02
Theme variable changes

gmahomarf:
In the course of trying to improve the Light Theme, I had to make some changes to the styles and CSS variables. I've tried to keep things backwards compatible

# Changed
- Changed: Refactored some styles to use CSS variables. In all cases, the previous style values were used as the default values for backwards compatibility.
- Changed: FA icons in buttons now use the same color as the containing button's text color

# Fixed
- Fixed: Fixed a bug in the font service that expected System-provided fonts to be located at the root of the web server, which would break in installations using a base URL (develop)

# Theme
- Added: new CSS variables for better theming
  + `--card-title-text-color`: Text color for card titles.
  + `--card-body-bg-color`: Background color for card bodies (`.card-body`).
  + `--card-overlay-text-color`: Text color in card overlays (e.g. the overlay when hovering over library entries).
  + `--side-nav-header-text-color`: Text color of side navigation headers.
  + `--pref-side-nav-header-text-color`: Text color of side navigation headers in the preferences page.
  + `--side-nav-item-color`:  The non-active counterpart to `--side-nav-item-active-color`, i.e. the color of the vertical bar to the left of side navigation items not currently active.

### DIFF
--- a/UI/Web/src/app/settings/_components/settings/settings.component.scss
+++ b/UI/Web/src/app/settings/_components/settings/settings.component.scss
@@ -1,7 +1,7 @@
 @use '../../../../theme/variables' as theme;
 
 h2 {
-  color: var(--side-nav-header-text-color, white);
+  color: var(--side-nav-header-text-color);
   font-weight: bold;
 }
 
--- a/UI/Web/src/app/settings/_components/settings/settings.component.scss
+++ b/UI/Web/src/app/settings/_components/settings/settings.component.scss
@@ -1,7 +1,7 @@
 @use '../../../../theme/variables' as theme;
 
 h2 {
-  color: white;
+  color: var(--side-nav-header-text-color, white);
   font-weight: bold;
 }
 
--- a/UI/Web/src/theme/components/_buttons.scss
+++ b/UI/Web/src/theme/components/_buttons.scss
@@ -267,6 +267,12 @@ button:disabled, .form-control:disabled, .form-control[readonly], .disabled, :di
     }
 }
 
+button {
+  i.fa, i.fa-regular {
+    color: var(--btn-fa-icon-color);
+  }
+}
+
 .btn-check:focus + .btn, .btn:focus {
   box-shadow: 0 0 0 0.05rem var(--btn-focus-boxshadow-color);
 }
--- a/UI/Web/src/theme/components/_buttons.scss
+++ b/UI/Web/src/theme/components/_buttons.scss
@@ -14,10 +14,18 @@
   --bs-btn-active-bg: var(--primary-color-dark-shade);
   --bs-btn-active-border-color: var(--primary-color-dark-shade);
 
+  i {
+    color: var(--btn-primary-text-color);
+  }
+
   &:hover {
     color: var(--btn-primary-hover-text-color);
     background-color: var(--btn-primary-hover-bg-color);
     border-color: var(--btn-primary-hover-border-color);
+
+    i {
+      --btncolor: var(-primary-hover-text-color);
+    }
   }
 }
 
@@ -30,10 +38,18 @@
   background-color: var(--btn-outline-primary-bg-color);
   border-color: var(--btn-outline-primary-border-color);
 
+  i {
+    color: var(--btn-outline-primary-text-color);
+  }
+
   &:hover {
     color: var(--btn-outline-primary-hover-text-color) !important;
     background-color: var(--btn-outline-primary-hover-bg-color) !important;
     border-color: var(--btn-outline-primary-hover-border-color) !important;
+
+    i {
+      color: var(--btn-outline-primary-hover-text-color) !important;
+    }
   }
 }
 
@@ -82,6 +98,10 @@
   background-color: var(--btn-secondary-outline-bg-color);
   border-color: var(--btn-secondary-outline-border-color);
 
+  i {
+    color: var(--btn-secondary-outline-text-color);
+  }
+
   &:hover {
     --bs-btn-color: var(--btn-secondary-outline-hover-text-color);
     --bs-btn-hover-bg: var(-btn-secondary-outline-hover-bg-color);
@@ -94,6 +114,10 @@
     box-shadow: inset 0px -2px 0px 0px var(--btn-secondary-outline-text-color);
     border-bottom-left-radius: 0;
     border-bottom-right-radius: 0;
+
+    i {
+      color: var(--btn-secondary-outline-hover-text-color);
+    }
   }
 }
 
@@ -103,10 +127,18 @@
   background-color: var(--btn-danger-outline-bg-color);
   border-color: var(--btn-danger-outline-border-color);
 
+  i {
+    color: var(--btn-danger-outline-text-color);
+  }
+
   &:hover {
     color: var(--btn-danger-outline-hover-text-color);
     background-color: var(--btn-danger-outline-hover-bg-color);
     border-color: var(--btn-danger-outline-hover-border-color);
+
+    i {
+      color: var(--btn-danger-outline-hover-text-color);
+    }
   }
 }
 
@@ -132,6 +164,9 @@
   background-color: var(--btn-disabled-bg-color);
   color: var(--btn-disabled-text-color);
   border-color: var(--btn-disabled-border-color);
+  i {
+    color: var(--btn-disabled-text-color);
+  }
 }
 
 button:disabled, .form-control:disabled, .form-control[readonly], .disabled, :disabled {
@@ -160,6 +195,10 @@ button:disabled, .form-control:disabled, .form-control[readonly], .disabled, :di
   color: var(--body-text-color);
   border: none;
 
+  i {
+    color: var(--body-text-color);
+  }
+
   &:disabled {
     --bs-btn-disabled-bg: transparent;
   }
@@ -178,14 +217,26 @@ button:disabled, .form-control:disabled, .form-control[readonly], .disabled, :di
 
 .btn-primary-text {
   color: var(--btn-primary-text-text-color);
+
+  i {
+    color: var(--btn-primary-text-text-color);
+  }
 }
 
 .btn-secondary-text {
   color: var(--btn-secondary-text-text-color);
+
+  i {
+    color: var(--btn-secondary-text-text-color);
+  }
 }
 
 .btn-danger-text {
   color: var(--btn-danger-text-text-color);
+
+  i {
+    color: var(--btn-danger-text-text-color);
+  }
 }
 
 
@@ -194,6 +245,10 @@ button:disabled, .form-control:disabled, .form-control[readonly], .disabled, :di
   color: var(--btn-secondary-text-color);
   background-color: var(--btn-secondary-bg-color);
   border-color: var(--btn-secondary-border-color);
+
+  i {
+    color: var(--btn-secondary-text-color);
+  }
 }
 
 .btn.btn-secondary.alt {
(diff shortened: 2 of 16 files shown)


UI/Web/src/app/settings/_components/settings/settings.component.scss
```diff
@@ -1,7 +1,7 @@
 @use '../../../../theme/variables' as theme;
 
 h2 {
-  color: white;
+  color: var(--side-nav-header-text-color, white);
```
majora2007: Anytime we introduce css variables, the values need to go in _dark.scss. Please go ahead and refactor these variables and their definition there. Ensure they are in the appropriate section (dictated by comment)

This can thus just be `color: var(--side-nav-header-text-color)`

UI/Web/src/theme/components/_buttons.scss
```diff
@@ -212,10 +212,34 @@ button:disabled, .form-control:disabled, .form-control[readonly], .disabled, :di
...
 }
 
-button i.fa {
-  color: var(--btn-fa-icon-color);
+$button_types: '', -primary, -secondary, -outline-primary, -outline-secondary, -danger;
+$icon_classes: fa, fa-regular, fa-solid;
+
+// FA icon colors for all button types
+// the .nav exception is for the Activity button in the top nav bar
+@each $button_type in $button_types {
```
majora2007: Can you explain what this change is for and why we need it?

UI/Web/src/theme/components/_input.scss
```diff
@@ -3,14 +3,16 @@ input:not([type="range"]), .form-control {
     color: var(--input-text-color);
     border-color: var(--input-border-color);
 
-    &:focus {
+    &:focus:not([type="checkbox"]) {
```
majora2007: Need help understanding what this translates to.

majora2007 (changes requested):
I validated that nothing breaks to my knowledge in the reader. Please respond to comments and take care of the requested actions. I can re-review afterwards. 

It will be helpful to raise a PR against the wiki as well with new variables.

UI/Web/src/app/settings/_components/settings/settings.component.scss
gmahomarf: Will do

UI/Web/src/theme/components/_buttons.scss
gmahomarf: I'll admit, some of this might be overkill, but I'd rather cover all the bases, which is why I chose to iterate over all classes of buttons and all classes of FA icons. For lighter themes, having all icons be the same color results in weird things. For example, in this image from the Users table with the Light theme and `--btn-fa-icon-color` set to a light green color. 
image

Here, it looks terrible, but it makes sense for outline-style buttons, like in this other screenshot 
image


If you prefer, I can write them all out individually instead of using the `@each` sytnax, or I can extract only the ones that are actually used.

UI/Web/src/theme/components/_input.scss
gmahomarf: It was a conflict when a checkbox is both focused and checked. Thinking about it, it might be better written as `:not(:checked)`, since the issue is only on checked checkboxes, not all of them.

Without this change, but with the exception to `:read-only` later in the file, this is an unfocused, checked checkbox:
image

This is a focused, checked checkbox:
image

This is checked and focused, with either `not(:checked)` or `:not([type="checkbox"])`
image

UI/Web/src/app/settings/_components/settings/settings.component.scss
gmahomarf: > This can thus just be `color: var(--side-nav-header-text-color)`

What about other themes?  I think it's a safe assumption that all current themes either keep these value as they are or override them with extra CSS. If we don't keep whatever the original value was as a default, we risk breaking existing themes.

UI/Web/src/theme/components/_sidenav.scss
```diff
@@ -227,7 +227,7 @@
           text-align: unset;
           margin-left: 0.75rem;
           font-size: 0.9rem;
-          color: #999999;
+          color: var(--side-nav-text-color, #999999);
```
gmahomarf: @majora2007 I just noticed that this is the only place in this file where you weren't using `--side-nav-text-color`. This color is different from the one in `dark.scss`. Not sure what you prefer to do here.

https://github.com/Kareadita/Kavita/blob/d4e3a2de3ea3e681052dec8d9a9fe405a61c970a/UI/Web/src/theme/themes/dark.scss#L247

UI/Web/src/app/settings/_components/settings/settings.component.scss
majora2007: All I'm asking is to move the color from here to the _dark.scss. Existing themes will then just inherit the default (white), so it will be a non-issue.

gmahomarf: Ah, I wasn't aware that all themes inherited from the dark theme. That makes perfect sense.

gmahomarf: This is done

UI/Web/src/theme/components/_buttons.scss
majora2007: Can we not just tweak the rule such that we handle the outline vs non-outline button classes differently and apply different variables? 

Something like this:
```

/* Target all outline button variants */
[class*="btn-"][class*="-outline"] icon,
[class*="btn-"][class*="-outline"] i,
[class*="btn-"][class*="-outline"] svg {
  /* your outline-specific icon styles */
}

/* Target all non-outline button variants (excludes outline) */
[class*="btn-"]:not([class*="-outline"]) icon,
[class*="btn-"]:not([class*="-outline"]) i,
[class*="btn-"]:not([class*="-outline"]) svg {
  /* your solid button icon styles */
}
```

UI/Web/src/theme/components/_sidenav.scss
majora2007: Hmm, I also see `color: #d5d5d5;` for the preference side nav (which I'm sure you'll want to scope with a css variable). 

As for the main nav, let's stick with `#999`.

UI/Web/src/theme/components/_buttons.scss
gmahomarf: I can do that, however my idea was that you need to have icon color variables to match all the different button text color variables:

```.css
--btn-danger-outline-hover-text-color
--btn-danger-outline-text-color
--btn-danger-text-text-color
--btn-disabled-text-color
--btn-outline-primary-hover-text-color
--btn-outline-primary-text-color
--btn-primary-hover-text-color
--btn-primary-text-color
--btn-primary-text-text-color
--btn-secondary-hover-text-color
--btn-secondary-outline-hover-text-color
--btn-secondary-outline-text-color
--btn-secondary-text-color
--btn-secondary-text-text-color
```

Thoughts?

UI/Web/src/theme/components/_sidenav.scss
gmahomarf: > Hmm, I also see `color: #d5d5d5;` for the preference side nav (which I'm sure you'll want to scope with a css variable).

I completely missed that one. Good catch.

> As for the main nav, let's stick with `#999`.

Will do

gmahomarf: > Hmm, I also see `color: #d5d5d5;` for the preference side nav

This one is actually being overridden by this part in `_sidenav.scss`, so the preference nav headers are white.

https://github.com/Kareadita/Kavita/blob/a3c8c9be333e947b3931813f01cac4326b7f1dcc/UI/Web/src/theme/components/_sidenav.scss#L205-L206

I can either remove the `color` property from the code referenced above entirely, or I can just set it to the same `--side-nav-header-text-color` variable

majora2007: For the preference side nav, let's just introduce a new css variable `--pref-side-nav-header-text-color`. I want to keep them as 2 different colors.

UI/Web/src/theme/components/_buttons.scss
majora2007: @Fesaa any thoughts? 

I personally don't like the extra bloat and logic we would have to maintain from the initial solution. From my understanding, you really only need 3 at the bare minimum. Outline needs a background color and non-outline needs a white (inverse of the primary), and a danger variant for outline.

gmahomarf: To clarify my previous comment, the variables I listed already existed before my changes. My goal was to provide the same level of color customization for button icons as there already was for button text. I wanted to avoid things like this screenshot: 
image

Assuming I didn't miss any other variables, there are currently 14 distinct combinations of button type (primary, outline, etc.) and status (hover, disabled, etc.) defined. While it's, admittedly, not 100% clear, my changes introduce 18 combinations of types and statuses, and they even exclude some combinations for which there are text color variables already. I'm happy to manually define only the same 14 combinations for icon color instead.

majora2007: So what I'm trying to grasp is why do we need that level of customization when in most cases, the icon will be the same color as the primary text of the button?

gmahomarf: I agree that most cases will have matching colors for icons and text. We could just make all FA icons inherit the button's text color by moving the icon styles inside each `.btn*` style currently defined. Alternatively, if we want the option to override icon colors, we can create icon color variables and default them to their respective text color variables. For the latter, you'd still need at a 1:1 ratio between text color and icon color variables (so 14 new variables). 

The fact remains, you already defined different variables for primary/secondary outline buttons and for primary/secondary regular buttons. There's no guarantee that a theme will have a single color for all outline buttons and a single different color for all non-outline buttons. 

I guess a third option would be to reduce the current customization options for button text, but that's a call I can't make.

majora2007: Theme creators can always write custom css in the theme file if they need it. The css variables are a way to customize without needing to know css. I think having the icons inherit the color of the text by default is ideal. 

In the case we did want to expand, we can go back to my original idea of:
> Outline needs an icon color (color text) and non-outline needs an icon color, and a danger variant for outline.

gmahomarf: > I think having the icons inherit the color of the text by default is ideal

Perfect! I'll refactor it later this week. I don't see a reason to expand from that since, like you said, themes can always include custom CSS

UI/Web/src/theme/components/_selects.scss
```diff
@@ -14,11 +14,6 @@
         box-shadow: 0 0 0 0.25rem var(--input-focus-boxshadow-color);
     }
 
-    &:read-only {
-        background-color: var(--input-bg-readonly-color);
-        cursor: initial;
-    }
-
```
gmahomarf, line 21: `readonly` has no effect on `<select>` elements, and like the comment about check-boxes above, when the `:read-only` selector is applied to elements that don't support it, it causes them to be selected always. This essentially caused all `<select>` elements to have `background-color: var(--input-bg-readonly-color);`

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only

UI/Web/src/theme/components/_buttons.scss
gmahomarf: Finally got some time today for this. I think this should be ready, too

majora2007:
Looks like I made a bad comment, the main side nav text nav should be hsla(0,0%,100%,.85). I can take care of this for you. Everything else checked out. 

Please make sure your PR to the wiki is aligned.

majora2007 (approved):

PR #4084, opened 2025-10-08
Polish Pass 1

majora2007:
# Added
- Added: (Kavita+) Added a Support Id to Kavita+. This is useful for me, whenever there are backend issues users are facing.

# Changed
- Changed: Optimized an API call for reading list items
- Changed: The epub reader will now remember where you came from when jumping to annotations, allowing you to go back where you left off (in the same reading session) (develop)

# Fixed
- Fixed: (OPDS) Fixed a bug where reading list continue point wasn't taking into account progress on a different page (develop)
- Fixed: Added back the chapter title in the manga reader (develop)
- Fixed: Fixed OIDC login button not being centered (develop)
- Fixed: Fixed accordion chevrons not being visible on some themes
- Fixed: Fixed confirm delete popup not being styled to the book theme (develop)
- Fixed: Fixed the page number for annotations being one off in the UI (develop)
- Fixed: Fixed width slider label having the wrong label when no override is configured (develop)
- Fixed: Fixed total size, not taking all chapter into account on the series page (develop)
- Fixed: Fixed mobile showing more series summary than on desktop on series detail page
- Fixed: Fixed a potential edge cage where genre/tags/... would not display all values on series detail page

### DIFF
--- a/UI/Web/src/app/announcements/_components/changelog/changelog.component.html
+++ b/UI/Web/src/app/announcements/_components/changelog/changelog.component.html
@@ -7,7 +7,7 @@
         <div ngbAccordion class="mb-2">
           <div ngbAccordionItem class="p-2">
             <h2 ngbAccordionHeader>
-              <button ngbAccordionButton>
+              <button ngbAccordionButton class="accordion-button">
                 <h4 class="changelog-header">{{update.updateTitle}}&nbsp;
                   @if (update.isOnNightlyInRelease) {
                     <span class="badge bg-secondary">{{t('nightly', {version: update.currentVersion})}}</span>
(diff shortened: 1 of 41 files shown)


UI/Web/src/app/announcements/_components/changelog/changelog.component.html
```diff
@@ -17,6 +17,7 @@ <h4 class="changelog-header">{{update.updateTitle}}&nbsp;
                     <span class="badge bg-secondary">{{t('available')}}</span>
                   }
                 </h4>
+                <i class="fas fa-chevron-up"></i>
```
majora2007: I can do it, but you need to aria-hidden="true" this, because the accordion already has the aria-expanded.

PR #4128, opened 2025-10-21
Improve required restart hint for OIDC

xoxys:
# Fixed
- Fixed: Improve the required restart hint for OIDC config to include all changes that require a manual restart. 

I know this might be a bit too broad now but IMO reads better than adding another one e.g. "If a, b or c was changed, then". At least changes to `Custom scopes` requires a manual restart to get properly applied.


xoxys:
Ah, lol I saw too later @Fesaa has already included this change in https://github.com/Kareadita/Kavita/pull/4126 Im going to close this PR then.

Issue #4180, opened 2025-11-05
[API] Same lastReadDate for all series


MnsieurJF:
### What happened?

Hello, 

Using the _/Series/on-deck_ and also the _/Series/all-v2_ endpoints, I noticed that the last reading date seems to be incorrect. 
In fact, every last reading date seems to be the same for all the series in the list, down to the milliseconds.

That could not be humanly possible, and maybe there's a bug somewhere.

I think the date corresponds to a real date where I read 1 chapter, but it seems to apply to every chapters.
Could be that you're using an already existing instance of an object instead of creating a new one. Therefor, it's the memory address of the value and not the value itself being transferred.

I put an example of the output I got in the log.

Let me know if you need more info.

### What did you expect?

Get the real last reading dates for all series using these endpoints

### Kavita Version Number - If you don't see your version number listed, please update Kavita and see if your issue still persists.

0.8.8 - Stable

### Are you accessing kavita through a reverse proxy? If yes, confirm that the issue persists with a direct connection

Yes

### What operating system is Kavita being hosted from?

Docker (LSIO Container)

### If the issue is being seen on Desktop, what OS are you running where you see the issue?

None

### If the issue is being seen on Mobile, what OS are you running where you see the issue?

None

### Relevant log output

```shell
[
  {
    "id": 1850,
    "name": "Conan the Barbarian (2023)",
    "originalName": "Conan the Barbarian (2023)",
    "localizedName": "Conan the Barbarian (2023)",
    "sortName": "Conan the Barbarian (2023)",
    "pages": 822,
    "coverImageLocked": true,
    "pagesRead": 657,
    "latestReadDate": "2025-11-05T11:40:19.863782",
    "lastChapterAdded": "2025-09-30T20:19:32.9720717",
    "userRating": 0,
    "hasUserRated": false,
    "format": 1,
    "created": "2024-08-23T16:49:21.3582887",
    "sortNameLocked": false,
    "localizedNameLocked": false,
    "wordCount": 0,
    "libraryId": 1,
    "libraryName": "Comics",
    "minHoursToRead": 4,
    "maxHoursToRead": 5,
    "avgHoursToRead": 4.506579,
    "folderPath": "/home/mnsieurjf/media/Comics/Titan Comics",
    "lowestFolderPath": "/home/mnsieurjf/media/Comics/Titan Comics/Conan the Barbarian (2023)",
    "lastFolderScanned": "2025-10-15T01:11:05.7264666",
[...]

PR #4190, opened 2025-11-07
Activity Graph

majora2007:
image

# Added
- Added: Added an Activity Graph for reading history per day per year. 

# Changed
- Changed: Updated the design of the tooltips across the app for better contrast

# Fixed
- Fixed: Fixed a bug where lastReadDate wasn't being calculated correctly on multiple APIs (#4180)

# Theme
- Added css variables to style tooltips: --tooltip-caret-color, --tooltip-bg-color, --tooltip-outline-color


API/Controllers/StatsController.cs
```diff
@@ -250,4 +250,20 @@ public async Task<ActionResult<StatCount<string>>> GetDeviceTypeCounts()
...
     #endregion
 
+
+    #region Reading History
+
+    [HttpGet("reading-activity")]
+    [ResponseCache(CacheProfileName = "Statistics")]
+    public async Task<ActionResult<ReadingActivityGraphDto>> GetReadingActivity(int userId, int year)
+    {
+        var user = await _unitOfWork.UserRepository.GetUserByUsernameAsync(Username!);
```
Fesaa, line 260: We can just get the user ID from the User object / claim

API/Data/ManualMigrations/v0.8.9/MigrateProgressToReadingSessions.cs
```diff
@@ -77,15 +78,19 @@ public static async Task Migrate(DataContext dataContext, ILogger<Program> logge
...
                             AppUserId = group.Key.AppUserId,
                             StartTime = sessionDate,
                             StartTimeUtc = sessionDateUtc,
-                            EndTime = sessionDate.AddHours(23).AddMinutes(59).AddSeconds(59),
-                            EndTimeUtc = sessionDateUtc.AddHours(23).AddMinutes(59).AddSeconds(59),
+                            EndTime = activityData[0].EndTime,
+                            EndTimeUtc = activityData[0].EndTimeUtc,
                             IsActive = false,
-                            ActivityData = [.. group.Select(item => CreateSessionActivityDataFromProgress(item.Progress, item.Chapter, item.Format))],
+                            ActivityData = [.. activityData],
```
Fesaa: I don't think we need the spread operator here any more? As it's already a list now, can remove the ToList above, or just pass activityData directly here

UI/Web/src/app/statistics/_components/activity-graph/activity-graph.component.html
```diff
@@ -0,0 +1,110 @@
...
+        </tr>
+        </thead>
+
+        <tbody>
+          @for (dayIndex of [0, 1, 2, 3, 4, 5, 6]; track dayIndex) {
+            <tr>
+              <td class="day-label">
+                <span class="visually-hidden">{{ dayIndex | dayLabel }}</span>
+                <span aria-hidden="true"
+                      [class.visually-hidden]="dayIndex !== 1 && dayIndex !== 3 && dayIndex !== 5">
```
Fesaa: Could be dayIndex % 2 === 0

UI/Web/src/app/statistics/_components/activity-graph/activity-graph.component.html
```diff
@@ -0,0 +1,110 @@
...
+    <ng-template #noActivityTooltip let-item="item">
+      <div class="activity-tooltip">
+        {{t('no-activity-tooltip', {date: item.date | utcToLocaleDate | ordinalDate})}}
+      </div>
+    </ng-template>
+
+    <ng-template #tooltip let-item="item">
+      <div class="activity-tooltip">
+        <div class="tooltip-header">
+          <strong>{{ item.date | utcToLocaleDate | date:'MMMM d, yyyy' }}</strong>
```
Fesaa, line 86: Do we want to hardcode the time format like this? utcToLocaleTime should be able to take a parameter so it returns without hour etc

UI/Web/src/app/statistics/_components/activity-graph/activity-graph.component.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (!entry) return 0;
+
+    if (entry.totalTimeReadingSeconds === 0 && entry.totalPages === 0) return 0;
+    if (entry.totalTimeReadingSeconds < 15 * 60) return 1; // Less than 15 minutes
+    if (entry.totalTimeReadingSeconds < 45 * 60) return 2; // Less than 45 minutes
+    if (entry.totalTimeReadingSeconds < 360 * 60) return 3; // Less than 1 hour
+    return 4; // 1 hour or more
+  }
+
+  private formatDate(date: Date): string {
```
Fesaa, line 183: Should we be hardcoding this format?

API/Controllers/StatsController.cs
majora2007: This is setup so that an admin can ideally get another user's data, not just their own.

UI/Web/src/app/statistics/_components/activity-graph/activity-graph.component.html
majora2007: The utcToLocaleDate doesn't handle formatting, it just converts from utc to a local date timezone.

UI/Web/src/app/statistics/_components/activity-graph/activity-graph.component.ts
majora2007: No, but I'm also unsure how to handle this translation to be more locale-friendly.

Issue #4211, opened 2025-11-13
[Kavita] Filter sort by your own rating


majora2007:
### Discussed in https://github.com/Kareadita/Kavita/discussions/4210

<div type='discussions-op-text'>

<sup>Originally posted by **nachfuellbar** November 13, 2025</sup>
### Idea Description

Nowadays we can only sort via Avg. Rating which can result in something you highly rated yourself being after something you disliked.
Would be great to be able to sort according to your own ratings

### Idea Category

Web UI

### Duration of Using Kavita

1,5 yeaars

### Before submitting

- [X] I've already searched for existing ideas before posting.</div>

majora2007:
This was an existing feature.

Edit: Misread the FR ask

ToniKielo:
Seconded. In plex I use ratings as "hype meters", so whenever a new episode of 5 star series is added it'll be the first one to watch.
I would like to use a similar method in kavita too.

Do we need changes anywhere else besides..
- SeriesSort.cs
- SortField.cs
- series-filter.ts
- en.json (and other languages)
- sort-field.pipe.ts
?

Fesaa:
This commit, https://github.com/Kareadita/Kavita/commit/a6c7049522b5f1408b56565dcb257a6511a6d487, adds a new filter option for annotations. It is similar (just in different methods) for the series filter, you can use it as reference. (You won't need the migration or changes in DbContext)

ToniKielo:
Added pull request. Hope it was correctly made. Let me know if you need changes.

majora2007:
Somehow this got missed, but this was delivered with v0.8.9.

PR #4234, opened 2025-11-25
sorting series by user rating

ToniKielo:
# Added
- Added: sorting series by user rating (Closes FR #4210, 1 upvote)

I made it so that series is firstly ordered by the rating, and secondly by the series sort name, so that it's less chaotic and more consistent.

Would've liked to add the rating actually visible in the UI only when this specific filter is in use, but that would be a larger change
and more of a question to the UX design. Someone who is doing UX could consider it later.

I only added english translation for the text, not sure what's the process on this.

### DIFF
--- a/API/Extensions/QueryExtensions/Filtering/SeriesSort.cs
+++ b/API/Extensions/QueryExtensions/Filtering/SeriesSort.cs
@@ -36,12 +36,8 @@ public static class SeriesSort
                 .Max(), sortOptions),
             SortField.AverageRating => query.DoOrderBy(s => s.ExternalSeriesMetadata.ExternalRatings
                 .Where(p => p.SeriesId == s.Id).Average(p => p.AverageScore), sortOptions),
-
-            SortField.UserRating => query.
-                DoOrderBy(s => s.Ratings.Where(r => r.SeriesId == s.Id && r.AppUserId == userId).Max(r => r.Rating), sortOptions)
-                .ThenBy(s => s.SortName.ToLower())
-                ,
-
+            SortField.UserRating => query.DoOrderBy(s => s.Ratings.Where(r => r.SeriesId == s.Id && r.AppUserId == userId).Max(r => r.Rating), sortOptions)
+                .ThenBy(s => s.SortName.ToLower()),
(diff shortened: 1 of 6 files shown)


ToniKielo:
references issue #4211

API/Extensions/QueryExtensions/Filtering/SeriesSort.cs
```diff
@@ -36,6 +36,12 @@ public static IQueryable<Series> Sort(this IQueryable<Series> query, int userId,
                 .Max(), sortOptions),
             SortField.AverageRating => query.DoOrderBy(s => s.ExternalSeriesMetadata.ExternalRatings
                 .Where(p => p.SeriesId == s.Id).Average(p => p.AverageScore), sortOptions),
+
+            SortField.UserRating => query.
+                DoOrderBy(s => s.Ratings.Where(r => r.SeriesId == s.Id && r.AppUserId == userId).Max(r => r.Rating), sortOptions)
+                .ThenBy(s => s.SortName.ToLower())
+                ,
```
Fesaa: Please format this the same way the other cases are. I.e comma on same line no white lines in between

UI/Web/src/assets/langs/en.json
```diff
@@ -2434,6 +2434,7 @@
         "read-progress": "Last Read",
         "average-rating": "Average Rating",
         "random": "Random",
+        "user-rating": "User rating",
```
Fesaa: "Own Rating" might be a better naming. Thoughts?

Fesaa (changes requested):
Haven't pulled down, but looks correct. Do want to check if SortName is always present

UI/Web/src/assets/langs/en.json
ToniKielo: I did think smart filters were global and not personal, so I wrote it more from admin perspective.
But since they seem to be for the creator only, "own" might be a better phrase.

majora2007:
I have some time, I'll review this. 

Edit: I need to wait till after progress-overhaul branch work since my DB wont run on this branch.

Fesaa:
Thank you! This will be released with the next nightly

PR #4254, opened 2025-12-04
Backups use SQLite VACUUM INTO

### LINKED ISSUE #3456: Automated backups not containing database files
### What happened?

Some users have reported that the `kavita.db` file is missing in backups. 

At first glance it seems like if the file is in use when the backup job fires off it will not copy the database. If nothing else, we should check the backup after it is created to make sure the DB exists in it. If not, fail the job. (Bonus, send an email to the admin if they have it setup)

Ideally, we can do some testing of the sqlite vacuum function to make sure nothing gets lost or discarded, then use the function to make a copy of the database at backup time. That would create a space optimized and compacted database users would be able restore to gain some performance back over time. 

### What did you expect?

Backups to contain the database. 

0.8.4.2 - Stable

None

None

_No response_

None

_No response_

### Relevant log output

_No response_

### Additional Notes

_No response_

DieselTech:
# Changed
- Changed: The backup service now uses the SQLite command VACUUM INTO to create a working copy of the database file at the time of backup. Resolves #3456 where the database would be missing from the backup zip file due to the file being locked when the backup would try to copy the database.


majora2007 (approved):
Tested and had no issues.

PR #4263, opened 2025-12-09
Fix container process ignoring signals, preventing clean stop

ZeroKnight:
# Fixed
- Fixed: Fixed container process ignoring signals, notably preventing cleanly stopping the container with `docker stop` or `podman stop`

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -30,7 +30,7 @@ fi
 echo "Starting Kavita"
 echo ls -l "/kavita/config/appsettings.json"
 
-./Kavita
+exec ./Kavita
 
 #if [[ "$PUID" -eq 0 ]]; then
 #    # Run as root


ZeroKnight:
This is a common issue with container entrypoints. When the main container process is run as a foreground job in a shell script, the shell does not pass signals like `SIGINT` and `SIGTERM` to the child process it is waiting on.[1]

Consequently, this inhibits container engines from cleanly stopping containers when requested via `docker stop`, `podman stop`, `systemctl stop` (Quadlets), etc. Since the stop signal (`SIGTERM` by default) is ignored by the shell running the process, after a default 10 second timeout, the container engine will result to forcefully killing (`SIGKILL`) the process; this is an immediate, **unclean** stop. This is not the expected behavior when requesting a graceful stop.

Less relevant but another consequence is that this also interferes with sending arbitrary signals via `docker kill` or `podman kill`; the shell will ignore them as well.

A very simple solution when using a shell script entrypoint is to `exec` the main process. This causes the main process to *replace* the shell process, thereby receiving all signals sent to it. Most importantly, the stop signal. The only caveat is nothing after the `exec` will run, so it should be the last action in the script.

[1] There is some nuance to this and exactly how signals are handled depends on the shell's job control settings, interactivity, and more. See the "SIGNALS" section in `bash(1)` and the docs for various flavors of `sh` for specific details.

ZeroKnight:
If running anything after `./Kavita` in the entrypoint script is necessary, then the `exec` approach won't work as anything after the `exec` builtin is essentially dead code. An alternative approach is possible with `trap` and `wait`, but is a little more involved.

majora2007:
@DieselTech or @Fesaa can one of you take a look at this?

Fesaa:
Thanks! This change will be released with the next nightly

PR #4289, opened 2025-12-18
Feature/pdf render changeable resolution

StereotypicalCat:
# Added
- Added: Added the ability to change the PDF renderer resolution (Mihon-apps) via a Server Setting. The admin can choose Default (1080x1920), High (1920x2560) or Ultra (2160x3840) (Thanks @StereotypicalCat)

# Changed
- Changed:  (Performance) Slight memory improvement in creating images from PDFs

### DIFF
--- a/API/DTOs/Settings/ServerSettingDTO.cs
+++ b/API/DTOs/Settings/ServerSettingDTO.cs
@@ -88,6 +88,10 @@ public sealed record ServerSettingDto
     /// </summary>
     public CoverImageSize CoverImageSize { get; set; }
     /// <summary>
+    /// How large rendered PDF images should be
+    /// </summary>
+    public PdfRenderResolution PdfRenderResolution { get; set; }
+    /// <summary>
     /// SMTP Configuration
     /// </summary>
     public SmtpConfigDto SmtpConfig { get; set; }
--- a/API/Entities/Enums/PdfRenderResolution.cs
+++ b/API/Entities/Enums/PdfRenderResolution.cs
@@ -15,16 +15,3 @@ public enum PdfRenderResolution
     /// </summary>
     Ultra = 3,
 }
-public static class PdfRenderResolutionExtensions
-{
-    public static (int dim1, int dim2) GetDimensions(this PdfRenderResolution size)
-    {
-        return size switch
-        {
-            PdfRenderResolution.Default => (1080, 1920),
-            PdfRenderResolution.High => (1920, 2560),
-            PdfRenderResolution.Ultra => (2160, 3840),
-            _ => (1080, 1920)
-        };
-    }
-}
--- a/API/Entities/Enums/PdfRenderResolution.cs
+++ b/API/Entities/Enums/PdfRenderResolution.cs
@@ -1,0 +1,30 @@
+﻿namespace API.Entities.Enums;
+
+public enum PdfRenderResolution
+{
+    /// <summary>
+    /// Default Size: 1080x1920 (wxh)
+    /// </summary>
+    Default = 1,
+    /// <summary>
+    /// 1920x2560
+    /// </summary>
+    High = 2,
+    /// <summary>
+    /// 2160x3840
+    /// </summary>
+    Ultra = 3,
+}
+public static class PdfRenderResolutionExtensions
+{
+    public static (int dim1, int dim2) GetDimensions(this PdfRenderResolution size)
+    {
+        return size switch
+        {
+            PdfRenderResolution.Default => (1080, 1920),
+            PdfRenderResolution.High => (1920, 2560),
+            PdfRenderResolution.Ultra => (2160, 3840),
+            _ => (1080, 1920)
+        };
+    }
+}
--- a/API/Services/BookService.cs
+++ b/API/Services/BookService.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using API.Data;
 using API.Data.Metadata;
 using API.DTOs.Reader;
 using API.Entities;
@@ -78,6 +79,7 @@ public partial class BookService : IBookService
     private const string BookApiUrl = "book-resources?file=";
     public const string BookReaderBodyScope = "//BODY/APP-ROOT[1]/DIV[1]/DIV[1]/DIV[1]/APP-BOOK-READER[1]/DIV[1]/DIV[2]/DIV[1]/DIV[1]/DIV[1]";
     private readonly PdfComicInfoExtractor _pdfComicInfoExtractor;
+    private readonly IUnitOfWork _unitOfWork;
 
     /// <summary>
     /// Setup the most lenient book parsing options possible as people have some really bad epubs
@@ -124,9 +126,10 @@ public partial class BookService : IBookService
         }
     };
 
-    public BookService(ILogger<BookService> logger, IDirectoryService directoryService, IImageService imageService, IMediaErrorService mediaErrorService)
+    public BookService(ILogger<BookService> logger, IDirectoryService directoryService, IImageService imageService, IMediaErrorService mediaErrorService, IUnitOfWork unitOfWork)
     {
         _logger = logger;
+        _unitOfWork = unitOfWork;
         _directoryService = directoryService;
         _imageService = imageService;
         _mediaErrorService = mediaErrorService;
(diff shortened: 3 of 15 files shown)


StereotypicalCat:
The option only works on new renders of course, so anything still in cache will be served at that resolution until the cache is busted and we have to render new images.

API/Entities/Enums/PdfRenderResolution.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// 1920x2560
+    /// </summary>
+    High = 2,
+    /// <summary>
+    /// 2160x3840
+    /// </summary>
+    Ultra = 3,
+}
+public static class PdfRenderResolutionExtensions
```
majora2007: Extensions need to be in a separate file

UI/Web/src/app/_pipes/pdf-render-resolution.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  transform(value: PdfRenderResolution): string {
+    switch (value) {
+      case PdfRenderResolution.Default:
+        return translate('pdf-render-resolution.default');
+      case PdfRenderResolution.High:
+        return translate('pdf-render-resolution.high');
+      case PdfRenderResolution.Ultra:
+        return translate('pdf-render-resolution.ultra');
+
```
majora2007: Remove empty line

UI/Web/src/app/admin/manage-media-settings/manage-media-settings.component.html
```diff
@@ -27,6 +27,23 @@
...
+        <div class="row g-0 mt-2">
+          @if(settingsForm.get('pdfRenderResolution'); as formControl) {
+            <app-setting-item [title]="t('pdf-render-resolution-label')" [subtitle]="t('pdf-render-resolution-tooltip')">
+              <ng-template #view>
+                {{formControl!.value | pdfRenderResolution}}
+              </ng-template>
+              <ng-template #edit>
+                <select class="form-select" formControlName="pdfRenderResolution">
+                  @for (opt of allPdfRenderResolutions; track opt) {
+                    <option [value]="opt">{{opt | pdfRenderResolution}}</option>
```
majora2007: Use [ngValue]

API/DTOs/Settings/ServerSettingDTO.cs
```diff
@@ -88,6 +88,10 @@ public sealed record ServerSettingDto
     /// </summary>
     public CoverImageSize CoverImageSize { get; set; }
     /// <summary>
+    /// How large rendered PDF images should be
+    /// </summary>
+    public PdfRenderResolution PdfRenderResolution { get; set; }
```
majora2007, line 93: Not sure if we should have the admin force it for all users or not. @Fesaa @DieselTech can you weigh in your thoughts?

StereotypicalCat:
Here is a comparison between ultra (left) and default (right)
image

majora2007 (changes requested):
I did a preliminary review on the PR. Please produce some metrics on the timings to convert a PDF on each setting for me.

API/DTOs/Settings/ServerSettingDTO.cs
StereotypicalCat: My logic here is that its the admin thats offering the compute and therefore should make the decision. Also if each user can choose their own resolution, the image cache would need to be changed, instead of having the same cache system we currently have.

StereotypicalCat:
> I did a preliminary review on the PR. Please produce some metrics on the timings to convert a PDF on each setting for me.

I got a 25 page ~140MB pdf to test this with.
To do the timings, i created a python script. First, i tried finding both min, max, average and median request times, but I realized that it seems kavita generates images for all the PDF's pages upon the first request, as all requests are pretty much instant after the first. The process I used to generate the data is as follows:
* Go into tasks and clear reading cache.
* Go into media and change to desired resolution.
* Run the script below

Following this, i have included max and total time, as all other requests are sub 0.2s
|                  | Total | Max  |
|------------------|-------|------|
| Default (~1080p) | 21.1  | 20.2 |
| High (~1440p)    | 28.8  | 27.0 |
| Ultra (~4K)      | 31.0  | 28.8 |

```import requests
import time
import statistics

BASE_URL = "http://localhost:5000"
ENDPOINT = "api/Reader/image"

API_KEY = ""  
EXTRACT_PDF = True
CHAPTER_ID = 1
PAGE_RANGE = range(1, 26)

times = []

for page in PAGE_RANGE:
    params = {
        "ExtractPdf": EXTRACT_PDF,
        "apiKey": API_KEY,
        "page": page,
        "chapterId": CHAPTER_ID,
    }

    url = BASE_URL.rstrip("/") + "/" + ENDPOINT

    start_time = time.perf_counter()
    response = requests.get(url, params=params)
    end_time = time.perf_counter()

    elapsed = end_time - start_time
    times.append(elapsed)

    print(f"Page {page}: {elapsed:.4f}s (status {response.status_code})")

# --- Statistics ---
total_time = sum(times)
min_time = min(times)
max_time = max(times)
avg_time = statistics.mean(times)
median_time = statistics.median(times)

print("\n=== Time Usage Statistics ===")
print(f"Requests:  {len(times)}")
print(f"Total:     {total_time:.4f}s")
print(f"Minimum:   {min_time:.4f}s")
print(f"Maximum:   {max_time:.4f}s")
print(f"Average:   {avg_time:.4f}s")
print(f"Median:    {median_time:.4f}s")

File: API/Entities/Enums/PdfRenderResolution.cs (hunk 1 above)
Comment (StereotypicalCat) [reply API/Entities/Enums/PdfRenderResolution.cs]: For this I was following the example from CoverImageSize.cs, where the extension is in the same file. Ive changed it to have its own file :)

File: API/DTOs/Settings/ServerSettingDTO.cs (hunk 4 above)
Comment (Fesaa) [reply API/DTOs/Settings/ServerSettingDTO.cs:93]: It should be up to the admin I think, ideally if data saver is enabled here it would serve the smaller files. But, as mentioned, unsure how worth the extra changes needed to the cache system would be

Would be interested to see how big files are, if they're 2/3x bigger. It might be worth it, people have bad internet

Comment (StereotypicalCat) [reply API/DTOs/Settings/ServerSettingDTO.cs:93]: > It should be up to the admin I think, ideally if data saver is enabled here it would serve the smaller files. But, as mentioned, unsure how worth the extra changes needed to the cache system would be
> 
> Would be interested to see how big files are, if they're 2/3x bigger. It might be worth it, people have bad internet

For the image provided in the example, the Ultra quality image is 7.852 KB, and the default is 2.393 KB, (~3,3x filesize).

Comment (StereotypicalCat) [reply API/DTOs/Settings/ServerSettingDTO.cs:93]: Having these changeable resolutions would also allow us to implement lower that 1080p resolutions in the future, for people on very low bandwidth :)
I agree that this being user-selectable would be cool, but for now just having the option for a server admin to increase resolution would be nice, and at the very least help me a lot :)

File: API/Services/BookService.cs [hunk 5]
```diff
@@ -1415,7 +1418,10 @@ public void ExtractPdfImages(string fileFilePath, string targetDirectory)
     {
         _directoryService.ExistOrCreate(targetDirectory);
 
-        using var docReader = DocLib.Instance.GetDocReader(fileFilePath, new PageDimensions(1080, 1920));
+        var settings = _unitOfWork.SettingsRepository.GetSettingsDtoAsync().Result;
```
majora2007, line 1421: Replace your implementation with this and we can merge it in:
```cs

_directoryService.ExistOrCreate(targetDirectory);

        var settings = _unitOfWork.SettingsRepository.GetSettingsDtoAsync().GetAwaiter().GetResult();
        var dims = settings.PdfRenderResolution.GetDimensions();
        var pageDimensions = new PageDimensions(dims.dim1, dims.dim2);

        int pages;
        using (var countReader = DocLib.Instance.GetDocReader(fileFilePath, pageDimensions))
        {
            pages = countReader.GetPageCount();
        }

        var parallelOptions = new ParallelOptions
        {
            MaxDegreeOfParallelism = Math.Min(Environment.ProcessorCount, 4)
        };

        Parallel.For(0, pages, parallelOptions, pageNumber =>
        {
            using var docReader = DocLib.Instance.GetDocReader(fileFilePath, pageDimensions);
            using var stream = StreamManager.GetStream("BookService.GetPdfPage");

            GetPdfPage(docReader, pageNumber, stream);

            var outputPath = Path.Combine(targetDirectory, $"Page-{pageNumber}.png");
            using var fileStream = new FileStream(
                outputPath,
                FileMode.Create,
                FileAccess.Write,
                FileShare.None,
                bufferSize: 81920);

            stream.Seek(0, SeekOrigin.Begin);
            stream.CopyTo(fileStream);
        });
```

majora2007: For reference, I just did a small optimization and replaced your .Result with .GetAwaiter().GetResult() which makes debugging slightly easier.

majora2007 (changes requested):
One change needed, then we can merge this in. Please retarget this to `fix/stats-edges`.

majora2007:
I just noticed your timezone, I'm going to make the change for you, since it's mostly my optimization I want in. Thanks for the contribution.

PR #4312, opened 2025-12-31
Device scoped reading profiles

Fesaa:
# Added
- Added: Reading profiles can now also be bound per device. Profiles with a device get precedence to profiles without. Each series can have at most one profiles for each device, and one without any. (Closes FR #2929, 27 upvotes)

# Changed
- Changed: Implicit profiles will now automatically be bound to the device you're using when they're created

# Fixed
- Fixed: Fixed NPE when opening the stats page if you had no progress yet (develop)


UI/Web/src/app/shared/_components/list-select-modal/list-select-modal.component.scss
```diff
@@ -1,4 +1,4 @@
-.clickable:hover, .clickable:focus {
```
majora2007, line 1: Do we have a distinct focus style? If not, we should add one for accessibility.

Fesaa: I'm not sure, but I removed it as it was causing some confusing visual results where it looked like you were hovering several at once. Distinct style might be helpful, but I didn't get the feeling that the focused element had much meaningful value here

majora2007 (approved):

PR #4398, opened 2026-01-25
Improve touch zones in the manga reader

kindlyfire:
This PR adds touch zone support to the longstrip reader.

# Added
- Added: Adds touch zone support for longstrip mode, which scrolls the page. Two scroll areas: bottom to scroll down, top to scroll up.
- Added: Option to control the scroll amount in longstrip mode


Fesaa:
This works great. A few comments however

- I'm not sure how I feel about having all touch zones active at once. If I'm reading a horizontally paging comic, clicking top/bottom to go left/right feels odd (Vice versa for a webtoon). Joe will need to weigh in here as well
- There's explicit key binds for paging up and down. Please hook the up & down logic here instead

majora2007:
I'm in agreement on the top/bottom mentioned, however, I have seen this in other readers and would be open to a dedicated mode for this. 

I haven't tested this out, but are you saying I can tap on the side of the webtoon reader and it will scroll a bit or will it go to the next page?

kindlyfire:
I can only speak for myself, but I really like having both zones active at once. On phones and tablets in particular, it makes moving through a comic with either hand equally simple. With just the left/right or top/bottom zones, I often have to click across the screen. It's a "must have" for me to like a reader now 🙂

I'm sure it felt odd for me at some point as well, but now I can't do without

---

> There's explicit key binds for paging up and down. Please hook the up & down logic here instead

Alright, I will look into this during the coming week

---

> I haven't tested this out, but are you saying I can tap on the side of the webtoon reader and it will scroll a bit or will it go to the next page?

Yes, it will scroll by 85% of the viewport height. I've experimentally found this to be a good number

> I'm in agreement on the top/bottom mentioned, however, I have seen this in other readers and would be open to a dedicated mode for this.

I'll turn it into a separate mode. Do you have a preferred way of showing it in the UI?

majora2007:
I think we need to control that allows to tweak how much it scrolls. 

For the icon, I imagine a box within a rectangle, I've seen that used on other readers.

Make sure when you add the mode, that you take care of disabling other options and that the mode shows the highlight, like the other ones do for pagenation zones.

kindlyfire:
I've updated the PR:

- Removed having all touch zones active at once (maybe a future PR)
- Added scrolling when using the page up/down/left/right shortcuts
- Allowed customizing the scroll amount, saved to the user profile
- Updated the PR description to match

kindlyfire:
I realized there was an issue with the touch zones; it works fine on desktop but on mobile they prevent dragging to scroll. I changed the approach to not add touch zone `div`s but listen to touch events instead.

API/Entities/User/AppUserReadingProfile.cs
```diff
@@ -91,6 +91,10 @@ public class AppUserReadingProfile
     /// Manga Reader Option: Disable the width override if the screen is past the breakpoint
     /// </summary>
     public BreakPoint DisableWidthOverride { get; set; } = BreakPoint.Never;
+    /// <summary>
+    /// Manga Reader Option: How much of the viewport to scroll per tap/key press in Webtoon mode (percentage, 50-100)
```
majora2007: I would remove the lower bound here or make it 10 rather. 50% is too high imo.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -117,7 +117,7 @@
         </div>
       } @else {
         @if (!isLoading && !inSetup) {
-          <div class="webtoon-images" appDblClick (doubleClick)="bookmarkPage($event)" (singleClick)="toggleMenu()">
+          <div class="webtoon-images" (click)="handleWebtoonClick($event)">
```
majora2007, line 120: You need to bring back the dblClick directive. You can use singleClick for your handleWebtoonClick.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -292,6 +292,15 @@
                          min="10" max="100" step="1" formControlName="darkness">
                 </div>
 
+                @if (readerMode === ReaderMode.Webtoon) {
+                  <div class="col-md-6 col-sm-12">
+                    <label for="webtoon-scroll-amount" class="form-label range-label">{{t('webtoon-scroll-amount-label')}}</label>
+                    <span class="ms-1 range-text">{{generalSettingsForm.get('webtoonScrollAmount')?.value + '%'}}</span>
+                    <input type="range" class="form-range" id="webtoon-scroll-amount"
+                           min="50" max="100" step="5" formControlName="webtoonScrollAmount">
```
majora2007: Set min to 10

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -401,6 +401,8 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
 
   // Renderer interaction
   readerSettings$!: Observable<ReaderSetting>;
+  private webtoonLastClickTime = 0;
+  private webtoonClickTimeout: any = null;
```
majora2007: You can use this type: ReturnType<typeof setTimeout> | null

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -1941,9 +1958,53 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      return;
+    }
+
+    if (y < viewportHeight * tapZoneSize) {
+      this.scrollByViewport(-1);
+    } else if (y > viewportHeight * (1 - tapZoneSize)) {
+      this.scrollByViewport(1);
+    } else {
+      // Center zone — single click = toggleMenu, double click = bookmarkPage
+      const now = Date.now();
```
majora2007, line 1996: I'm not personally a fan of duplicating the logic here from the dblClick directive. This needs a rethink.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -1941,9 +1958,53 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    const y = event.clientY;
+    const viewportHeight = window.innerHeight;
+    const tapZoneSize = 0.25;
+
+    if (this.menuOpen) {
+      this.toggleMenu();
+      return;
+    }
+
+    if (y < viewportHeight * tapZoneSize) {
```
majora2007: This logic reads a bit difficult for me, can you extract into named variables.

majora2007 (commented):
I just looked over the code briefly, I will test this out this week. I need to make sure it feels correct as well.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
kindlyfire: I had two problems that caused me to implement double click like this. I often tap very quickly in webtoon mode to move down the page, which will register as double clicks when they shouldn't, and it causes extra click delay. I implemented it such that there's no delay when clicking the next/prev zones, only in the menu zone, and bookmark double clicks work in the menu zone.

To note we can't use click/dblClick handlers on zone overlay elements, since that would stop drag-to-scroll from working (I looked, but haven't found a workaround)

Is that ok with you?

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
kindlyfire: Same reasoning here as with my comment below

kindlyfire:
I apologize for the delay on this. I should be able to update the PR tomorrow

kindlyfire:
I've made the requested changes, other than the double click stuff since there's a reason for that (given above)

majora2007:
Thanks, I'll test this out soon.

majora2007:
Sorry got busy with the reading list overhaul project. I've been working in the webtoon reader recently and still interested in testing this out.

Fesaa:
Heya, do you still have the time to finish this off? I can take over if not 🙂. Otherwise, you'll need to redo your migration, and merge develop. To redo the migration you'll need to 

Update your database to the migrate before, undo last migration, merge in develop, then create it again.

Using the Rider UI for this is will likely be easiest. 
Migrations project is Kavita.Database
Startup project is Kavita.Server

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.html
```diff
@@ -293,6 +293,15 @@
                          min="10" max="100" step="1" formControlName="darkness">
                 </div>
 
+                @if (readerMode === ReaderMode.Webtoon) {
```
Fesaa, line 296: The addition of this setting is pushing the close button of the screen on my phone. We'll want to make the setting scrollable on phone? Or look for a different solution, so I can close them

Kavita.Models/DTOs/UserReadingProfileDto.cs
```diff
@@ -78,6 +78,9 @@ public sealed record UserReadingProfileDto
     [EnumDataType(typeof(BreakPoint))]
     public BreakPoint DisableWidthOverride { get; set; } = BreakPoint.Never;
 
+    /// <inheritdoc cref="AppUserReadingProfile.WebtoonScrollAmount"/>
+    public int WebtoonScrollAmount { get; set; } = 85;
```
Fesaa, line 82: You need to add the field in ReadingProfileService#UpdateReaderProfileFields. This setting is not being saved now.

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -1989,9 +2005,58 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    const viewportHeight = window.innerHeight;
+    const tapZoneSize = 0.25;
+    const inTopZone = y < viewportHeight * tapZoneSize;
+    const inBottomZone = y > viewportHeight * (1 - tapZoneSize);
+    return inTopZone || inBottomZone;
+  }
+
+  scrollByViewport(direction: 1 | -1) {
+    const scrollPercent = (this.generalSettingsForm?.get('webtoonScrollAmount')?.value || 85) / 100;
+    document.body.scrollBy({ top: document.body.clientHeight * scrollPercent * direction, behavior: 'smooth' });
```
Fesaa, line 2058: I'm not sure how much is possible, but this doesn't feel the most amazing when starting a new scroll before the old one has ended. 

Is it possible to chain them together more smoothly?

UI/Web/src/app/manga-reader/_components/manga-reader/manga-reader.component.ts
```diff
@@ -1989,9 +2005,58 @@ export class MangaReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+      } else {
+        this.webtoonClickTimeout = setTimeout(() => {
+          this.toggleMenu();
+        }, 300);
+      }
+      this.webtoonLastClickTime = now;
+    }
+  }
+
+  private isInWebtoonScrollZone(event?: Event): boolean {
```
Fesaa, line 2045: The logic for inTopZone & inBottomZone is duplicated between this method and the above. Can we extract this to a method returning two bools?

Fesaa (changes requested):

PR #4413, opened 2026-02-07
Annotations Drawer and Card Improvements

Ark1369:
Annotations Drawer and Card Improvements to facilitate handling of longer Texts/Annotations.

# Changed
- Changed: Improved the UX of the Annotation drawer to allow large annotations to be fully viewable.


Ark1369:
### Issue
Long Annotation Texts such as 
image
Are displayed as extract in Annotations Tab for Series and All Annotations both, whether is normal view or pop up view
image
image

### Problem
It makes Annotations unusable to remember key details in long running series on the fly, since you have to constantly go into Reader to check past 1st 2 lines of content. This is extremely time-consuming in larger Web Novel style EPUBs where there are 1000s of chapters, so opening them in Reader poses a significant delay back and forth.

Also current system uses a fix Width to display Annotations causing a large waste of space on bigger screens.

### Solution
Change the Pop Up view to show the full Annotated Text regardless of Length such as below
image

- Oveflow Change adds a Scrollbar instead of hiding text beyond the clamp limit, thus allowing user to see entire text through Scroll.
- white-space and word-wrap - Preserve Linebreak and prevent horizontal word spill, thus making text look like it is in Reader instead of a clumped paragraph.
- webkit line clamp increase to 8 adds a respectable viewing window to scroll comfortably and see all contents (can be increased or maybe give user an option).

These changes would significantly improve the Annotations System QoL by not requiring jumping back into the reader to see them, just like most of dedicated EPUB Reading apps do.

Additionally, Similar improvements could be made to the Tiles Itself as showcased in image below
image

- Overflow adds scroll, Whitespace/Wordwrap adds linebreaks for neater viewing
- Height increased to 10 complements webkit line clamp of 4. Otherwise, with default 7.78 - 3 lines work the best

image

UI/Web/src/app/book-reader/_components/_drawers/view-edit-annotation-drawer/view-edit-annotation-drawer.component.scss
```diff
@@ -12,10 +12,12 @@
 .content-quote {
   color: var(--drawer-text-color);
   max-height: 320px;
```
Fesaa: Let's see if we can change this too. It can't take up all space as we do need to be able to write the text, there's some more space we could take up to display the selected text. Maybe we can make it % based? Need to think about big notes as well here

Image

Image

majora2007: Can we used fixed positioning with the input box? This would let the quote part take up it's natural space.

majora2007:
Any updates to asks?

majora2007:
I took a look at this and fixed it up with an alternative that fits better:
image

I'm now good with merging this.

PR #4565, opened 2026-03-25
A small collection of bug fixes

### LINKED ISSUE #4558: Logout not working with OIDC and custom basepath
### What happened?

I set up OIDC using Google as the OIDC provider. When I attempted to logout I get a redirect back to the basepath but I am not presented with the login screen. Upon digging in the session cookie is stored in the browser with `path=/basepath` but the set-cookie on logout redirect has the `path=/`. This is causing the actual session cookie to not be expired.

Response headers from the redirect.
Image

Cookie in browser storage
Image

### What did you expect?

I expected logout to work and to be presented with the login screen on successful logout.

0.8.9.1 - Stable

Yes

Linux

Windows

Firefox

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

### LINKED ISSUE #4561: Upload API silently fails when base64 data includes data URI prefix
## Bug Description

The `POST /api/Upload/series` endpoint silently fails to save cover images when the `url` field contains a data URI string (e.g., `data:image/jpeg;base64,/9j/4AAQ...`) instead of raw base64. The API returns HTTP 200 but no cover is written to disk.

## Root Cause

In `ImageService.CreateThumbnailFromBase64()`, the `encodedImage` parameter is passed directly to `Convert.FromBase64String()` without stripping the data URI prefix:

```csharp
using var thumbnail = Image.ThumbnailBuffer(
    Convert.FromBase64String(encodedImage), thumbnailWidth);
```

`Convert.FromBase64String()` throws a `FormatException` on the `data:image/...;base64,` prefix. The exception is caught by the generic `catch (Exception e)` block, logged as "Error creating thumbnail from url", and the method returns `string.Empty`. The controller then saves `coverImageLocked = true` and the empty cover filename, returning 200 OK.

From the caller's perspective, the upload appears successful — HTTP 200, `coverImageLocked` is set — but no cover file is created.

## Expected Behavior

The API should either:
1. **Strip the data URI prefix** before decoding (preferred — makes the API more forgiving), or
2. **Return an error response** (4xx) when the base64 data is malformed

## Suggested Fix

Add prefix stripping in `CreateThumbnailFromBase64`:

```csharp
public string CreateThumbnailFromBase64(string encodedImage, string fileName, ...)
{
    try
    {
        // Strip data URI prefix if present (e.g., "data:image/jpeg;base64,")
        var commaIndex = encodedImage.IndexOf(',');
        if (commaIndex >= 0 && encodedImage.StartsWith("data:"))
        {
            encodedImage = encodedImage[(commaIndex + 1)..];
        }

        targetDirectory ??= directoryService.CoverImageDirectory;
        using var thumbnail = Image.ThumbnailBuffer(
            Convert.FromBase64String(encodedImage), thumbnailWidth);
        // ...
```

This matches the behavior of the frontend's `_cleanBase64Url()` method in the Angular app, which strips the prefix before sending. The backend should be equally tolerant for API consumers that don't strip it.

## How to Reproduce

1. Call `POST /api/Upload/series` with:
   ```json
   {"id": 42, "url": "data:image/jpeg;base64,/9j/4AAQSkZJRg...", "lockCover": true}
   ```
2. Response: HTTP 200 (empty body)
3. Series now has `coverImageLocked: true` but `coverImage` is empty — no file on disk
4. The old cover (or default placeholder) continues to display

## Environment

- **Kavita Version:** 0.8.9.1
- **Docker:** Yes
- **Related issue:** #1779 (same `FormatException` in stack trace, closed as unreproducible)

Fesaa:
# Changed
- Changed: Bubble up format exceptions in upload controller. (Closes #4561)
- Changed: You will now be prompted to re-download when trying to download a series if part of has already been downloaded. (develop)

# Fixed
- Fixed: Fixed OIDC logout for dysfunctional idps when using a baseurl. (Fixes #4558)
- Fixed: Fix mark as read with session not creating session for chapters that have never had progress before (develop)
- Fixed: Fix bulk library delete
- Fixed: Fix external ids not being updated until the page is refreshed after changing them (develop)
- Fixed: Fix external ids not updating when changing them through the UI (develop)

### DIFF
--- a/Kavita.Database/Repositories/AppUserProgressRepository.cs
+++ b/Kavita.Database/Repositories/AppUserProgressRepository.cs
@@ -263,23 +263,41 @@ public class AppUserProgressRepository(DataContext context, IMapper mapper) : IA
 
     public Task<Dictionary<int, int>> GetUserProgressForChaptersBySeries(int userId, int seriesId, CancellationToken ct = default)
     {
-        return context.AppUserProgresses
-            .Where(p => p.SeriesId == seriesId && p.AppUserId == userId)
-            .ToDictionaryAsync(p => p.ChapterId, p => p.PagesRead, ct);
+        return context.Chapter
+            .Where(c => c.Volume.SeriesId == seriesId)
+            .GroupJoin(
+                context.AppUserProgresses.Where(p => p.AppUserId == userId && p.SeriesId == seriesId),
+                c => c.Id,
+                p => p.ChapterId,
+                (c, progresses) => new { c.Id, PagesRead = progresses.Select(p => p.PagesRead).FirstOrDefault() }
+            )
+            .ToDictionaryAsync(x => x.Id, x => x.PagesRead, ct);
     }
 
     public Task<Dictionary<int, int>> GetUserProgressForChaptersByVolumes(int userId, int seriesId, List<int> volumeIds, CancellationToken ct = default)
     {
-        return context.AppUserProgresses
-            .Where(p => p.SeriesId == seriesId && volumeIds.Contains(p.VolumeId) && p.AppUserId == userId)
-            .ToDictionaryAsync(p => p.ChapterId, p => p.PagesRead, ct);
+        return context.Chapter
+            .Where(c => c.Volume.SeriesId == seriesId && volumeIds.Contains(c.VolumeId))
+            .GroupJoin(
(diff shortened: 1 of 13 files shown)


majora2007 (approved):
LGTM

PR #4633, opened 2026-04-23
Fix mihon volume progress tracking

### LINKED ISSUE #38: Inconsistent volume syncing
### Steps to reproduce

Have a Kavita library of books with volume naming scheme.

### Expected behavior

Read status in Mihon/forks should reflect the status in my Kavita library.

### Actual behavior

Syncing is inconsistent. I have some series with volume and chapters, these seem to work fine. I have some series with volumes and either they aren't fully synced or not synced at all. 

### Mihon/Komikku version

1.13.3

### Kavita version

0.8.7.2

### Android version

Android 15

### Other details

I've been using CDisplayEx for a while and sync has always worked perfect. It doesn't use the ODPS url, instead using the IP address and login info from Kavita. 

I had originally been using Tachiyomi/Mihon to access my library but moved away due to these inconsistencies so this issue has persisted for a while, maybe since inception. I'm hoping to use Komikku to access my library because downloading of new chapters is a feature CDisplay lacks, but this feature is the only thing keeping me from fully moving over.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated Mihon/Komikku to last version.
- [x] I have updated the extension to last version.
- [x] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/mihonapp/mihon/issues/new/choose).
- [x] I will fill out all of the requested information in this form.

Nedra1998:
# Fixed
- Fixed: 500 error when calling the `/api/Tachiyomi/latest-chapter` API for Volumes (Partially Fixes: https://github.com/Kareadita/tach-extension/issues/38)

### DIFF
--- a/Kavita.Models/DTOs/ChapterDto.cs
+++ b/Kavita.Models/DTOs/ChapterDto.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices.JavaScript;
--- a/Kavita.Services/TachiyomiService.cs
+++ b/Kavita.Services/TachiyomiService.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading.Tasks;
 using System.Collections.Immutable;
 using System.Collections.Generic;
@@ -96,7 +96,8 @@ public class TachiyomiService(
         {
             // Use R to ensure that localization of underlying system doesn't affect the stringification
             // https://docs.microsoft.com/en-us/globalization/locale/number-formatting-in-dotnet-framework
-            Number = (number / 10_000f).ToString("R", EnglishCulture)
+            Number = (number / 10_000f).ToString("R", EnglishCulture),
+            Files = new List<MangaFileDto>()
         };
     }
 


Nedra1998:
When Mihon tries to pull the current reading progress it calls the `/api/Tachiyomi/latest-chapter` for the series. When the latest chapter is a Volume, there is special logic to disambiguate it from a chapter by dividing the number by 10,000 ([here](https://github.com/Kareadita/Kavita/blob/develop/Kavita.Services/TachiyomiService.cs#L95-L100)). However, the fake chapter that is created fails to serialize due to a `System.ArgumentNullException` because Files was not defined.

This then leads to a 500 error and thus Mihon cannot pull the current read progress. This does _not_ fully address the issue, but does allow Mihon to pull read progress from Kavita, although the tracking numbers in Mihon will still be wrong.

Kavita.Models/DTOs/ChapterDto.cs
```diff
@@ -176,7 +176,7 @@ public class ChapterDto : IHasReadTimeEstimate, IHasCoverImage, IHasMetadataIds
     /// <inheritdoc cref="Chapter.SecondaryColor"/>
     public string? SecondaryColor { get; set; } = string.Empty;
 
-    public MangaFormat? Format => Files.FirstOrDefault()?.Format;
+    public MangaFormat? Format => Files?.FirstOrDefault()?.Format;
```
Fesaa: Copilot is correct here. I'd rather update the CreateTachiyomiChapterDto method to include an empty list, such that missing the includes when getting from db when needed does still throw; as that would be a pain to debug.

Fesaa (changes requested):

Kavita.Models/DTOs/ChapterDto.cs
Nedra1998: Can do, I was debating between the two but I wasn't sure if anything else expected it to be null for some reason (and I wasn't familiar with C# nullability syntax).

Copilot was suggesting setting the default in the ChapterDto to be an empty list (like we do with all the other `ICollection` fields
```csharp
public ICollection<MangaFileDto> Files { get; init; } = new List<MangaFileDto>!;
```
Are you suggesting instead leave that as `default!`, and instead update the CreateTachiyomiChapterDto to be
```csharp
return new TachiyomiChapterDto()
{
    // Use R to ensure that localization of underlying system doesn't affect the stringification
    // https://docs.microsoft.com/en-us/globalization/locale/number-formatting-in-dotnet-framework
    Number = (number / 10_000f).ToString("R", EnglishCulture),
    Files = new List<MangaFileDto>()
};
```

I can do either one, but would appreciate a bit more context to why one would be preferred over the other.

Fesaa: I'd go with adding it in CreateTachiyomiChapterDto

Nedra1998: Okay, I've updated the PR accordingly :+1:

Fesaa (approved):

Fesaa:
Thank you for the fix, this will go out with the next nightly

PR #4690, opened 2026-05-08
Pass in thumbnail height to better handle thumbnail size

### LINKED ISSUE #4148: Custom media has the wrong image sizes
### What happened?

When uploading a cover image (for Issues, Categories and People), Kavita generates an image with the correct aspect ratio, but with **the width for height** from the selected preset. This happens regardless of what media type is selected (PNG, AVIF, or PNG).

Regardless of which preset I choose in **Media Settings**, the resulting image does not match the configured output size.

| Selected size (Media Settings) | Expected output | Actual output |
|--------------------------------|-----------------|----------------|
| Default | 320×455 | 226×320 |
| Medium | 640×909 | 451×640 |
| Large | 900×1277 | 634×900 |
| Extra Large | 1265×1795 | 892×1265 |

It seems the image width is being applied relative to the longer side (height) instead of the shorter side (width), resulting in a smaller-than-expected image.

This only applies to uploaded media, generated media seems to always have the correct dimensions.

### What did you expect?

The generated images should **match the preset dimensions exactly** (e.g., “Medium (640×909)” should output a 640×909 image).  
The proportions are fine, only the overall size is wrong.

0.8.8 - Stable

Yes

None

None

_No response_

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

KobraKid:
# Changed
- Changed: Uploaded thumbnails scale using configured width and height, not just width. (Closes #4148)

### DIFF
--- a/Kavita.API/Services/IImageService.cs
+++ b/Kavita.API/Services/IImageService.cs
@@ -18,9 +18,10 @@ public interface IImageService
     /// <param name="fileName"></param>
     /// <param name="encodeFormat">Convert and save as encoding format</param>
     /// <param name="thumbnailWidth">Width of thumbnail</param>
+    /// <param name="thumbnailHeight">Height of thumbnail</param>
     /// <param name="targetDirectory">If null, will write to <see cref="DirectoryService.CoverImageDirectory"/></param>
     /// <returns>File name with extension of the file. </returns>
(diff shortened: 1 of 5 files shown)


Fesaa:
Sorry for the delay on this PR, I've merged it into my bugfix branch for the next nightly. Thank you!

PR #4696, opened 2026-05-11
Exit fullscreen when closing reader

JUMBUZI:
# Fixed
- Fixed: Closing the book/manga reader now exits fullscreen properly

### DIFF
--- a/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
+++ b/UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
@@ -324,10 +324,6 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
    * Library Type used for rendering chapter or issue
    */
    libraryType: LibraryType = LibraryType.Book;
-  /**
-   * If the web browser is in fullscreen mode
-   */
-  isFullscreen: boolean = false;
 
 
   /**
@@ -699,7 +695,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
             await this.goToPage();
             break;
           case KeyBindTarget.ToggleFullScreen:
-            this.applyFullscreen();
+            this.toggleFullscreen();
             break;
           case KeyBindTarget.ToggleMenu:
             this.actionBarVisible.update(x => !x);
@@ -866,6 +862,7 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
 
     this.navService.showNavBar();
     this.navService.showSideNav();
+    this.readerService.exitFullscreen();
   }
 
   async ngOnInit() {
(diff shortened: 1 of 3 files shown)


UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
```diff
@@ -2206,26 +2203,8 @@ export class BookReaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
-      this.readerService.toggleFullscreen(this.reader().nativeElement, () => {
-        this.isFullscreen = false;
-        this.cdRef.markForCheck();
-        this.renderer.removeStyle(this.reader().nativeElement, 'background');
-      });
-    } else {
-      this.readerService.toggleFullscreen(this.reader().nativeElement, () => {
-        this.isFullscreen = true;
-        this.cdRef.markForCheck();
-        // HACK: This is a bug with how browsers change the background color for fullscreen mode
```
majora2007, line 2221: I believe we still need this. Can you validate?

JUMBUZI: I just removed this since the arguments are currently unused so this does nothing.
https://github.com/Kareadita/Kavita/blob/develop/UI/Web/src/app/_services/reader.service.ts#L344

majora2007: I meant the HACK here for setting the colors

JUMBUZI: The hack is in the code block that is the callback argument for readerService.toggleFullscreen, and readerService.toggleFullscreen never calls the callback. So the code here is currently never ran. If that hasn't caused issues so far, I assume it is not needed. Removing this does not change current behavior, but not calling the callback might have been a bug. I'm not sure since I'm not sure what this is supposed to fix. If this is still needed, I can revert this part and make readerService.toggleFullscreen actually call the callback.

majora2007: Oh interesting. I haven't gotten to test this out as I need to run it on my tablet to validate. It's on my list to get this validated in the next month or so (sorry it's been so long).

majora2007:
I still have this on my list of things to review for v0.9.1 FYI. I just need to stop being lazy and validate on my tablet.

UI/Web/src/app/book-reader/_components/book-reader/book-reader.component.ts
majora2007: Finally got around to testing and you're right, the background color issue isn't an issue anymore. If I recall, browsers used to let it leak out of the component.

majora2007 (approved):

majora2007:
FYI, I was updating code and did find a bug in this code. When you removed the isFullscreen = state and just used the toggle, it broke the icon from updating.

PR #4711, opened 2026-05-18
Kavita+ Audit Log

### LINKED ISSUE #4705: Kavita+ Audit Log
Bring visibility into what Kavita+ is doing via an Audit table and surface data via 3 new screens. Screens are non-final. 

Image

Image

Image

Functionalties:
- [x] Showcase Kavita+ Audit Events in Series-Detail page (scoped to series). Driving CTA -> My Activity or All Activity (admin-only)
- [x] My Activity screen which shows active Scrobble Providers and their state
- [ ] Ability to retry failed Scrobble events (replay the latest progress event)
- [ ] All activity tab for Admins

majora2007:
image

image

image

This is the first piece of a lot of architectural changes to make the Kavita+ integration feel understandable. Since Kavita+ launched, the systems have been evolving, but understanding what is happening under the hood has been difficult. The pressure was added with both AniList and MAL having many downtime events and users wondering when metadata is matching, needs rematching, or why a scrobble event didn't post. 

I hope this Audit system helps shed light and I am open to feedback to ensure it addresses the needs. 

# Added
- Added: Added 3 new screens that help users understand what Kavita+ is doing around their series, their scrobbling, and for admins, the whole behind the scenes logic. (Closes #4705)

# Changed
- Changed: Scrobbling screen is likely going to be replaced by the new screens

# Fixed
- Fixed: Fixed a bug where marking a chapter as read wasn't triggering scrobbling.
- Fixed: Fixed a bug where MAL was never getting a proper response for token expiration in the Account Screen
- Fixed: Fixed a bug where series detail page could refresh the cover when a chapter cover update event triggered. 

Note: Relies on #4691


Kavita.Database/Extensions/QueryableExtensions.cs
```diff
@@ -373,6 +373,28 @@ public static IQueryable<Series> FilterMatchState(this IQueryable<Series> query,
...
+
+    /// <summary>
+    /// Filters to series that are matched but whose cached metadata has expired and needs a refresh.
+    /// A stale series is still considered matched — the data is just out of date.
+    /// </summary>
+    public static IQueryable<Series> WhereStaleExternalMetadata(this IQueryable<Series> query)
+    {
+        return query
+            .WhereMatchedExternalMetadata()
+            .Where(s => s.ExternalSeriesMetadata!.ValidUntilUtc < DateTime.UtcNow);
```
Fesaa: Does this work? Where's chained are AND this is checking ValidUntilUtc > and then <. I think this returns nothing?

Kavita.Server/Controllers/CBLController.cs
```diff
@@ -123,6 +123,11 @@ public async Task<ActionResult<CblSavedFileDto>> SaveCblFromUrl(UploadUrlDto dto
 
     private async Task<(bool IsInvalid, ActionResult<CblSavedFileDto>? ActionResult)> HasInvalidExtensionAsync(string filename, string fullPath)
     {
+        if (!ValidateFilename(Path.Join(fullPath, filename)))
+        {
+            return (false, null);
```
Fesaa: Shouldn't this return true? The boolean is called IsInvalid

Kavita.Server/Controllers/KavitaPlusAuditController.cs
```diff
@@ -0,0 +1,72 @@
...
+using Kavita.Models.Constants;
+using Kavita.Models.DTOs.KavitaPlus;
+using Kavita.Server.Attributes;
+using Kavita.Server.Extensions;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Kavita.Server.Controllers;
+
+[KPlus]
```
Fesaa, line 13: Small note with the attribute, we need to make sure the endpoints are never called if a licence isn't active or there's be red toasts

Kavita.Server/Program.cs
```diff
@@ -122,7 +123,31 @@ public static async Task Main(string[] args)
...
+                    Environment.Exit(1);
+                }
+                catch (OperationCanceledException)
+                {
+                    logger.LogCritical("Database migration cancelled due to shutdown signal");
+                    throw;
+                }
+                catch (Exception ex)
+                {
+                    logger.LogCritical(ex, "Failed to run critical Migrations, restore from a backup");
```
Fesaa, line 149: Does this need an exit or throw? This will continue

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -1442,24 +1505,23 @@ private async Task<bool> UpdateChapterPublisher(Chapter chapter, MetadataSetting
...
         if (string.IsNullOrEmpty(coverUrl)) return false;
 
         if (chapter.CoverImageLocked && !HasForceOverride(settings, chapter, MetadataSettingField.ChapterCovers))
         {
-            return false;
-        }
-
-        if (string.IsNullOrEmpty(coverUrl))
-        {
+            // TODO: LogTrace/Debug here to indicate why we rejected updating image for debugging?
```
Fesaa: Trace would be helpful, debug is too much I think

Kavita.Server/Controllers/KavitaPlusAuditController.cs
majora2007: Yeah, it's handled by surrounding routes and components, shouldn't be an issue.

Kavita.Server/Controllers/CBLController.cs
majora2007: Good catch, copy/pasted wrong line

PR #4753, opened 2026-06-18
Fixed: Bound parse concurrency to prevent ThreadPool starvation during scans

junoh-moon:
tl;dr let the API server stay responsive during library scans.

# Fixed
- Fixed: Library scans of a folder containing many files (most noticeable on large EPUB libraries) could make the web UI and API unresponsive for the duration of the scan, sometimes long enough that a Docker or Kubernetes health check would restart the container in the middle of scanning. Scans now limit how many files are parsed at the same time, so a large folder no longer starves the server while it is being scanned.

### DIFF
--- a/Kavita.Services.Tests/ParseScannedFilesTests.cs
+++ b/Kavita.Services.Tests/ParseScannedFilesTests.cs
@@ -668,4 +668,103 @@ public class ParseScannedFilesTests: AbstractDbTest
         Assert.Single(scannedSeries);
         Assert.Single(scannedSeries.Values.First().DistinctBy(x => x.Series));
     }
+
+    #region ParseFiles Concurrency
+
+    /// <summary>
+    /// Wraps <see cref="MockReadingItemService"/> and records how many ParseFile calls run
+    /// concurrently, so we can assert the parallel parse path is bounded.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


majora2007:
I haven't heard of any issues with one folder of 190k files having issues, can you elaborate on the setup and machine specs?

junoh-moon:
@majora2007 My setup is smaller than 190k files, but I could still reproduce the issue on a constrained k3s deployment.

   - ~6000 books (EPUB, PDF, CBZ)
   - Ubuntu 24.04 LTS
   - Intel N100 / 16 GB RAM
   - Kavita running on k3s
   - Container limit: 2 CPU / ~1 GB RAM

During scans, the UI/API would become unresponsive, and `/api/health` could time out long enough for Kubernetes probes to fail, sometimes causing a restart mid-scan.

majora2007:
Okay I'll take a look. I got report from one user that they had something similar happen.

majora2007:
I wasn't able to reproduce an issue, but I did scan this against my prod library and found no issues, so it's likely safe to merge. I'll probably group this into one of the big feature releases we are doing. 

Thanks.

majora2007 (approved):

PR #4772, opened 2026-06-30
Improved match flow & Library Metadata Providers

Fesaa:
# Added
- Added: Added support for loading metadata from Hardcover
- Added: Added the ability to use shortcode:id (hardcover:1234, al:1234, etc) for direct lookup in the search field

# Changed
- Changed: Automatic matching now requires exactly one match above 90%
- Changed: A library now has a metadata provider (Hardcover, MangaBaka, CBR) associated with it
- Changed: Spammy audit event logs will now only be logged once every 24h (develop)
- Changed: Scrobble events will no longer be created if the required id is missing. This will be logged in the audit log for visibility

# Fixed
- Fixed: Fixed a bug where Scrobbling wouldn't properly record a failure due to a capitalization difference from K+
- Fixed: Fixed a short circuit being executed when there was no existing external metadata row, which prevented background match from properly calling the full K+ series detail API for full metadata.
- Fixed: There was an edge case where a stale file reference cover could exist and cause a throw for finding the better image from an online source. Kavita will now properly catch and log a warning (not error). Chapter will now behave like other entities and accept a cover when there is no cover (data is better than nothing).
- Fixed: Fixed Writer/Artist people metadata not working for MangaBaka
- Fixed: Fixed an issue where a manual match wasn't writing the external metadata Ids, when it should (develop)

image


Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -139,26 +139,131 @@ public async Task FetchExternalDataTask(CancellationToken ct = default)
...
+        if (!await _licenseService.HasActiveLicense(ct: ct)) return null;
+
+        var series = await _unitOfWork.SeriesRepository.GetSeriesByIdAsync(seriesId, SeriesIncludes.Library | SeriesIncludes.Chapters, ct: ct);
+        if (series == null) return null;
+
+        if (!series.WillScrobble() || !series.Library.AllowMetadataMatching) return null;
 
-        // Generate key based on seriesId and libraryType or any unique identifier for the request
-        // Check if the request is allowed based on the rate limit
         if (!RateLimiter.TryAcquire(string.Empty))
```
Fesaa: HasRequiredId checks needs to be before this I think?

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -139,26 +139,131 @@ public async Task FetchExternalDataTask(CancellationToken ct = default)
...
+            MangaBakaId = series.MangaBakaId,
+            CbrId = series.CbrId,
+            HardcoverId = series.HardcoverId
+        };
+
+        series.MangaBakaId = match.Series.MangabakaId ?? 0;
+        series.AniListId = match.Series.AniListId ?? 0;
+        series.MalId = match.Series.MALId ?? 0;
+        series.HardcoverId = match.Series.HardcoverId ?? 0;
+        series.CbrId = match.Series.CbrId ?? 0;
```
Fesaa, line 246: Needs standalone bool

UI/Web/src/app/_models/series-detail/external-series-detail.ts
```diff
@@ -31,6 +31,7 @@ export interface ExternalSeriesDetail {
   malId?: number | null;
   mangabakaId?: number | null;
   hardcoverId?: number | null;
+  isStandAlone: boolean;
```
Fesaa, line 34: Do we want to add a manual switch for this in series edit?

UI/Web/src/app/sidenav/_modals/library-settings-modal/library-settings-modal.component.ts
```diff
@@ -131,9 +134,34 @@ export class LibrarySettingsModalComponent implements OnInit {
...
+        return [MetadataProvider.Hardcover, MetadataProvider.Mangabaka];
+      case LibraryType.Images:
+        return [MetadataProvider.Mangabaka, MetadataProvider.ComicBookRoundup];
+      case LibraryType.LightNovel:
+        return [MetadataProvider.Mangabaka, MetadataProvider.Hardcover];
+      case LibraryType.ComicVine:
+        return [MetadataProvider.ComicBookRoundup, MetadataProvider.Hardcover];
+    }
+
+    console.warn("Did not match the selected library type: ", libraryType, libraryType === LibraryType.ComicVine);
```
Fesaa: Remove ComicVine equal log

Kavita.API/Repositories/ISeriesRepository.cs
```diff
@@ -122,7 +123,7 @@ public Task<IList<Series>> GetAllSeriesByAnyNameAsync(string seriesName, string
     Task RemoveFromOnDeckAsync(int seriesId, int userId, CancellationToken ct = default);
     Task ClearOnDeckRemovalAsync(int seriesId, int userId, CancellationToken ct = default);
     Task<PagedList<SeriesDto>> GetSeriesDtoForLibraryIdAsync(int userId, UserParams userParams, SeriesFilterV2Dto seriesFilterDto, QueryContext queryContext = QueryContext.None, CancellationToken ct = default);
-    Task<PlusSeriesRequestDto?> GetPlusSeriesDtoAsync(int seriesId, CancellationToken ct = default);
+    Task<SeriesDetailRequestV3Dto?> GetSeriesDetailRequestV3Dto(int seriesId, CancellationToken ct = default);
```
majora2007: We should still scope the name with KavitaPlus

Kavita.Database/Repositories/SeriesRepository.cs
```diff
@@ -600,19 +600,20 @@ public async Task<PagedList<SeriesDto>> GetSeriesDtoForLibraryIdAsync(int userId
...
+            .Select(series => new SeriesDetailRequestV3Dto
             {
-                MediaFormat = series.Library.Type.ConvertToPlusMediaFormat(series.Format),
+                Provider = series.Library.MetadataProvider,
+                Format = series.Library.Type.ConvertToPlusMediaFormat(series.Format),
                 SeriesName = series.Name,
-                AltSeriesName = series.LocalizedName,
+                AlternativeNames = new List<string> { series.LocalizedName },
                 // TODO: Refactor this to check from ExternalMetadataIds first then ExternalSeriesMetadata. Weblink parsing is pointless as it updates in externalMetadata
                 AniListId = series.ExternalSeriesMetadata.AniListId != 0
```
majora2007: We really got to figure out how to handle this comment. I think we should send it from series.AniListId directly

Kavita.Models/DTOs/KavitaPlus/ExternalMetadata/ExternalMetadataIdsDto.cs
```diff
@@ -13,6 +13,7 @@ public sealed record ExternalMetadataIdsDto
     public int? AniListId { get; set; }
     public int? MangabakaId { get; set; }
     public int? HardcoverId { get; set; }
+    public bool IsStandAlone { get; set; }
```
majora2007, line 20: We should give context like IsHardcoverStandAlone

Kavita.Models/DTOs/KavitaPlus/ExternalMetadata/MetadataRequest.cs
```diff
@@ -5,6 +5,7 @@ public record MetadataRequest
     public int? AniListId { get; set; }
     public long? MalId { get; set; }
     public int? HardcoverId { get; set; }
+    public string? HardcoverSlug { get; set; }
     public int? CbrId { get; set; }
     public long? MangabakaId { get; set; }
```
majora2007, line 10: Another thing we should do (not in this PR), but change MangabakaId long -> int and capitalize the B.

Kavita.Models/Entities/Library.cs
```diff
@@ -66,6 +66,8 @@ public class Library : IEntityDate, IHasCoverImage
     /// </summary>
     public string DefaultLanguage { get; set; } = "";
```
majora2007: Please use string.Empty instead of ""

Kavita.Server/Controllers/LibraryController.cs
```diff
@@ -726,6 +738,25 @@ private void UpdateLibrarySettings(UpdateLibraryDto dto, Library library, bool u
         unitOfWork.LibraryRepository.Update(library);
     }
 
+    private static void ValidateMetadataProvider(LibraryType type, MetadataProvider provider)
+    {
+        List<MetadataProvider> validProviders = type switch
```
majora2007: Let's move this to a service/extension as a static readonly instance, so it's one instance across the app.

Kavita.Services/Plus/ScrobbleService/ChapterScrobbleService.cs
```diff
@@ -234,6 +265,15 @@ public async Task ScrobbleReadingUpdate(ScrobbleUpdateContext ctx, CancellationT
     {
         if (!SupportedEvents.Contains(ScrobbleEventType.ChapterRead) || ctx.Chapter == null) return;
 
+        if (!HasRequiredIds(ctx.Chapter))
+        {
+            await auditService.LogTemperedAsync(al => al.SubjectId == ctx.Chapter.Id && al.UserId == ctx.User.Id, KavitaPlusAuditCategory.Scrobble,
```
majora2007: This code is repeated a lot. Let's make a private method so we aren't repeating ourselves so much.

Kavita.Services/Plus/ScrobbleService/SeriesScrobbleService.cs
```diff
@@ -38,11 +38,22 @@ public abstract class SeriesScrobbleService<T>(ILogger<T> logger, IUnitOfWork un
...
+    protected abstract bool HasRequiredIds(Series series);
+
     public async Task ScrobbleReadStatusUpdates(ScrobbleUpdateContext ctx, ScrobbleReadStatus status,
         TransitionRuleKind? ruleKind = null, string? ruleHash = null, CancellationToken ct = default)
     {
         if (!SupportedEvents.Contains(ScrobbleEventType.ReadStatusUpdate) || ctx.Chapter != null) return;
 
+        if (!HasRequiredIds(ctx.Series))
+        {
+            await auditService.LogTemperedAsync(al => al.SeriesId == ctx.Series.Id && al.UserId == ctx.User.Id, KavitaPlusAuditCategory.Scrobble,
```
majora2007: DRY this out

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -234,39 +339,41 @@ public async Task<IList<ExternalSeriesMatchDto>> MatchSeries(MatchSeriesDto dto,
...
 
-        // TODO: Match needs to be overhauled
-        var matchRequest = new MatchSeriesRequestDto()
+        var matchV3Request = new MatchRequestV3Dto
         {
-            Format = format,
-            Query = query,
+            AniListId = potentialAnilistId ?? ExternalIdParser.GetAniListId(series.Metadata.WebLinks),
+            MalId = potentialMalId ?? ExternalIdParser.GetMalId(series.Metadata.WebLinks),
+            HardcoverId = null, // This is complex (There's two types of Ids; Series & Books)
```
majora2007: Does this comment still apply? Aka do we need to do more work around it?

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -454,212 +547,162 @@ private async Task<SeriesDetailPlusDto> FetchExternalMetadataForSeries(int serie
...
-                await _unitOfWork.CommitAsync(ct);
-            }
+        externalSeriesMetadata.ValidUntilUtc = DateTime.UtcNow.Add(_externalSeriesMetadataCache);
+        externalSeriesMetadata.AverageExternalRating = extRatings.Count != 0 ? (int) extRatings
+            .Average(r => r.AverageScore) : 0;
+
+        // prefer what was passed in (manual match), fall back to what K+ returned
+        var beforeIds = new AuditLogMatchExternalIdsParamsDto { AniListId = series.AniListId, MalId = series.MalId, MangaBakaId = series.MangaBakaId, CbrId = series.CbrId, HardcoverId = series.HardcoverId };
+
+        // TODO: Need to rethink how all these Ids work and only write on MetadataFetchTrigger.OnDemand
```
majora2007: We can remove this right? With the guard in place, we shouldn't have bad matches?

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -454,212 +547,162 @@ private async Task<SeriesDetailPlusDto> FetchExternalMetadataForSeries(int serie
...
+        await _auditService.LogMatchAsync(KavitaPlusEventType.SeriesMatched, seriesId,
+            new AuditLogMatchedParamsDto {
+                SeriesName = series.Name,
+                Before = beforeIds, After = afterIds,
+                MatchedName = result.Series?.Name
+            }, ct: ct);
+
+        // If there is metadata and the user has metadata download turned on
+        var madeMetadataModification = false;
+        if (result.Series != null && (series.Library!.AllowMetadataMatching || fromMatchFlow))
```
majora2007, line 703: Shouldn't we exit earlier if metadata is off?

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -1398,15 +1415,18 @@ private async Task<bool> UpdateChapters(Series series, MetadataSettingsDto setti
         var matchedChapters = allChapters
             .Join(
                 externalMetadata.ChapterDtos,
-                chapter => chapter.Range,
-                dto => dto.IssueNumber,
+                chapter => Parser.IsLooseLeafVolume(chapter.Range) ? chapter.Volume.Name : chapter.Range,
```
majora2007, line 1479: Why not use Min/Max numbers?

UI/Web/src/app/_models/series-detail/external-series-detail.ts
majora2007: Yes, we'd need something like that (unless we can get the id and validate the value)

UI/Web/src/app/_single-module/match-series-modal/match-series-modal.component.html
```diff
@@ -74,29 +72,26 @@ <h4 class="modal-title" id="modal-basic-title">
           </div>
 
           <div class="small text-muted">
-            <app-transloco-inject key="match-series-modal.selected-desc">
```
majora2007, line 77: I still want to render out who we are searching against to the user.

UI/Web/src/app/sidenav/_modals/library-settings-modal/library-settings-modal.component.ts
```diff
@@ -131,9 +134,34 @@ export class LibrarySettingsModalComponent implements OnInit {
...
+    metadataProvider: new FormControl<MetadataProvider>(MetadataProvider.Mangabaka, {nonNullable: true, validators: []}),
     // TODO: Missing excludePatterns
   });
 
+  selectedLibraryType = toSignal(this.libraryForm.get('type')!.valueChanges.pipe(
+    map(() => this.libraryForm.getRawValue().type as LibraryType),
+  ), { initialValue: LibraryType.Manga });
+  validMetadataProviders = computed(() => {
+    const libraryType = parseInt(this.selectedLibraryType() + '') as LibraryType;
+    switch (libraryType) {
```
majora2007: Can we do this in the backend so we aren't duplicating logic?

Kavita.Database/Migrations/20260623102348_MatchRework.cs
```diff
@@ -0,0 +1,40 @@
...
+                type: "INTEGER",
+                nullable: false,
+                defaultValue: false);
+
+            migrationBuilder.AddColumn<int>(
+                name: "MetadataProvider",
+                table: "Library",
+                type: "INTEGER",
+                nullable: false,
+                defaultValue: 0);
```
majora2007: I don't see logic that sets this on fresh library creation.

majora2007 (changes requested):

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -139,26 +139,131 @@ public async Task FetchExternalDataTask(CancellationToken ct = default)
...
+        {
+            AniListId = series.AniListId,
+            MalId = series.MalId,
+            HardcoverId = series.HardcoverId,
+            CbrId = series.CbrId,
+            MangabakaId = series.MangaBakaId,
+            MetronId = series.MetronId,
+            ComicVineId = series.ComicVineId,
+            IsStandAlone = series.Volumes.Sum(v => v.Chapters.Count) == 1,
+            Provider = series.Library.MetadataProvider,
```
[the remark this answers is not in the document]
Fesaa: We cannot use this value for automated matches, as they run before this value is ever set (If it's set and correct, HasRequiredId returns true).

The match flow also doesn't use it for true distinction but rather for loading extra options.

Kavita.Models/DTOs/KavitaPlus/ExternalMetadata/ExternalMetadataIdsDto.cs
Fesaa: I've added a comment, as it's possible future metadata provider will have this distinction too. And seeing as we do not support more than one metadata provider it's not needed to separate them

Kavita.Server/Controllers/LibraryController.cs
Fesaa: I've moved all (What I found / remembered) configuration like this for K+ to a static class called KavitaPlusConfiguration. So in the future when we add new things it should be easy to do so. Will need some testing

I noticed image library type didn't support metadata in the past, I had set it to CBR & MangaBaka (I.e. cbz types). Not sure if we want to support this? Or if the UI will need updates for this. We can just remove the lines in that class to remove

Kavita.Services/Plus/ExternalMetadataService.cs
Fesaa: Yeah, the manual match now also doesn't use those ids to create the request so it won't cause loops

Kavita.Services/Plus/ExternalMetadataService.cs
Fesaa: Depends on if we want to load recommendations & review if metadata is on/off I suppose.

Kavita.Services/Plus/ExternalMetadataService.cs
Fesaa: Don't think we can do that in a join, and it does a equals compare. We can change to logic to look for a matching chapter in the loop, rather than a join. Needs some thinking

UI/Web/src/app/_single-module/match-series-modal/match-series-modal.component.html
Fesaa: This happens
<div class="d-flex align-items-center gap-2 flex-wrap">
            <span class="fw-semibold small text-truncate">
              {{t('searching-against', {provider: match.primaryProvider! | metadataProviderTitle })}}
            </span>
          </div>

Kavita.Services/Plus/ExternalMetadataService.cs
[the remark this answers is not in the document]
majora2007: Couldn't we is standalone || volume count? 

Let's leave a comment for us about it running before a manual input

Kavita.Server/Controllers/LibraryController.cs
majora2007: I'm not sure. I'm torn as people can have them nice enough but it's more likely to be poor. 

We can keep as-is for now.

PR #4783, opened 2026-07-10
Fixed: Inverted greater/less than comparisons on the Read Last filter

### LINKED ISSUE #4716: "Less Than" and "Greater Than" is backwards on the "Read Last" filter
### What happened?

Just like the title states. "Less Than" and "Greater Than" is backwards on the "Read Last" filter. So currently if I put Read Last greater than 10000 days ago it shows volumes read within the last 10000 days and vice versa.

Image

### What did you expect?

I expect it to show the volumes I have not read in 10000 days. So none.

0.9.0 - Stable

Yes

Linux

None

_No response_

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

TowyTowy:
# Fixed
- Fixed: The 'Read Last' smart filter had its 'greater than' and 'less than' comparisons reversed. Because this filter matches on how many days it has been since a series was last read, 'read last greater than N days' now correctly returns the series you have not read in over N days, instead of the ones you just read (and vice versa for 'less than'). (Fixes #4716)

### DIFF
--- a/Kavita.Database/Extensions/Filters/SeriesFilter.cs
+++ b/Kavita.Database/Extensions/Filters/SeriesFilter.cs
@@ -319,24 +319,33 @@ public static class SeriesFilter
 
             var date = DateTime.Now.AddDays(-timeDeltaDays);
 
+            // ReadLast filters on the number of days since the series was last read, not on a raw date.
+            // A larger day-delta means the series was last read further in the past, i.e. an *earlier*
+            // MaxDate, so the numeric comparisons are inverted relative to the date they map to. For
+            // example 'read last greater than 30 days' means 'haven't read in over a month', which is
+            // MaxDate < date. (See #4716)
             switch (comparison)
             {
                 case FilterComparison.Equal:
                     subQuery = subQuery.Where(s => s.MaxDate != null && s.MaxDate.Equals(date));
                     break;
-                case FilterComparison.IsAfter:
                 case FilterComparison.GreaterThan:
-                    subQuery = subQuery.Where(s => s.MaxDate != null && s.MaxDate > date);
-                    break;
-                case FilterComparison.GreaterThanEqual:
-                    subQuery = subQuery.Where(s => s.MaxDate != null && s.MaxDate >= date);
-                    break;
-                case FilterComparison.IsBefore:
-                case FilterComparison.LessThan:
                     subQuery = subQuery.Where(s => s.MaxDate != null && s.MaxDate < date);
                     break;
-                case FilterComparison.LessThanEqual:
+                case FilterComparison.GreaterThanEqual:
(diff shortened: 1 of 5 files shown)


TowyTowy:
Root cause: in `SeriesFilter.HasReadLast` the day-delta value is turned into a cutoff date (`Now - N days`), but the numeric comparisons were applied as if the value were a raw date. A larger day count maps to an earlier `MaxDate`, so the numeric operators had to be inverted; `IsAfter`/`IsBefore` keep their existing date semantics. The method's own doc comment ("Haven't read in a month") describes the intended behaviour the code contradicted. Added regression tests in `SeriesFilterTests` covering greater/less than and the boundary case (they fail before the fix, pass after; full class 77 green).

Disclosure: this change was prepared with the assistance of an AI agent (Claude) and reviewed/verified before submission.

Fesaa:
This needs a manual migration to fix all existing smart filters

It's probably also a good idea to have an info icon in the UI to explain how the comparison works as they're both correct interpretations, I think there's a way to do this already.

TowyTowy:
Added the migration in a follow-up commit: `ManualMigrateReadLastSmartFilterComparison` (a v0.9.1 ManualMigration registered in `Startup.cs`, following the existing smart-filter migration precedents rather than a schema EF migration). It decodes each saved `AppUserSmartFilter` via `SmartFilterHelper` and, on Read Last statements only, flips the numeric operators (GreaterThan↔LessThan, GreaterThanEqual↔LessThanEqual). Equal/NotEqual are symmetric, and IsAfter/IsBefore keep raw-date semantics in both the old and new `SeriesFilter` code, so those - and every non-ReadLast statement - are left untouched. Net effect: existing filters return the exact same series after the fix as before it (proved by predicate-equivalence and covered by unit tests: each flip case, the untouched cases, a mixed filter, and value preservation). Idempotent via ManualMigrationHistory.

One thing to confirm: I preserved each filter's *observed* behavior (least surprise) rather than reinterpreting intent - which is what a migration is for here, since the code fix alone would otherwise silently invert existing filters. Shout if you'd rather not carry the old behavior forward. I left the optional UI info icon out to keep this focused on the migration; happy to add it separately.

majora2007:
Hey, I'm not comfortable merging this PR in. We can take the code and thank you in the notes, but your original statement: "Disclosure: this change was prepared with the assistance of an AI agent (Claude) and reviewed/verified before submission." doesn't seem correct as the code was committed by Claude and the comment you left is completely AI generated. 

You can either a) Close this and redo it with the learning or b) Close this and we will fix it ourselves.

TowyTowy:
Understood, and no problem - closing it out. Appreciate the review and the candor. Please feel free to take the fix and the migration and use them however works best for the project. Thanks!

PR #4807, opened 2026-07-29
Retroactively apply metadata settings

### LINKED ISSUE #4804: OIDC: Kavita skipping custom scopes
### What happened?

Thanks for the great tool and the continued support!

I am trying to integrate Kavita with authelia, and sync user settings with OIDC roles, and would like to use a custom claim for this, rather than authelia's default groups claim. For this, I have created a custom scope.

When starting kavita, I get the following warning message:
`[Kavita] [2026-07-28 10:35:52.800 +02:00  46] [Warning]  Scope foobar is configured, but not supported by your OIDC provider. Skipping`

After this, my custom scope is not requested by Kavita when logging in.

I suspect Kavita is checking `.well-known/openid-configuration` to determine the list of available scopes, and dropping those not shown in there. In authelia's case, `.well-known/openid-configuration` only seems to be showing the default scopes. As per the [OIDC specification](https://openid.net/specs/openid-connect-discovery-1_0.html) I don't think they are required to show custom scopes in there.

I see how this startup check makes sense for the `roles` scope, which Kavita will try to request by default. But for custom scopes I don't think it is really needed: I am manually configuring Kavita to request those so I know they will be available regardless of what `.well-known/openid-configuration` says.

Is there a way to make Kavita ignore these checks?

PD: Link to a discussion on this on authelia's side: https://github.com/authelia/authelia/discussions/12616

### What did you expect?

I expect Kavita to always asks for custom scopes on OIDC login regardless of what `.well-known/openid-configuration` says.

0.9.0 - Stable

Yes

Linux

None

_No response_

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

Fesaa:
To make our Metadata mappings more powerful we're adding a button to quickly run libraries through all your settings. Removing the need to force scan a library to retroactively apply changes. Making metadata management easier than ever!

Only one rerun can run at a time, and will block library scans while it's running.

image

image

# Added
- Added: Added a button to rerun all metadata mappings without needing to force scan your library

# Changed
- Changed: Lower log level for flavicon retry failures to reduce log spam.
- Changed: Black- and Whitelists will now be sorted alphabetically to help manage long lists. 
- Changed: User defined OIDC scopes will no longer be filtered by the supported scope list from the discovery document. Please ensure your configured scopes are all supported. (Closes #4804) 
- Changed: Kavita is now better at finding ISBNs in epubs

# Fixed
- Fixed: Fixed an incorrect message when deleting bookmarks in bulk
- Fixed: Fixed K+ metadata from Mangabaka sometimes not being able to be processed (develop)
- Fixed: Fixed a heap of missing access checks (low impact)

# API
- Changed: all enum values in request bodies will now be validated. Invalid values will return 400 Bad Request


Kavita.Models/DTOs/Metadata/ReRunMappingsRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Libraries to skip, can be used to request all - 1 libraries
+    /// </summary>
+    public List<int> ExcludedLibraries {  get; init; }
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        if (!AllLibraries && IncludedLibraries?.Count == 0)
+        {
+            yield return new ValidationResult(
+                "You must either select 'All Libraries' or specify at least one included library.",
```
majora2007: This needs to be localized

Kavita.Models/DTOs/Metadata/ReRunMappingsRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+                [nameof(AllLibraries), nameof(IncludedLibraries)]
+            );
+        }
+
+        if (ExcludedLibraries?.Intersect(IncludedLibraries ?? []).Count() > 0)
+        {
+            yield return new ValidationResult(
+                $"{nameof(ExcludedLibraries)} and {nameof(IncludedLibraries)} cannot intersect",
```
majora2007: Same here, localization needed.

Kavita.Models/DTOs/SignalR/MessageFactory.cs
```diff
@@ -882,4 +886,19 @@ public static SignalRMessage ExternalMetadataUpdateEvent(int seriesId)
             }
         };
     }
+
+    public static SignalRMessage ReRunMappingsProgressEvent(string progressEventType, float progress)
+    {
+        return new SignalRMessage()
+        {
+            Name = ReRunMappingsProgress,
+            Title = "Re-run mappings",
```
majora2007: We should say metadata mappings here.

Kavita.Models/DTOs/SignalR/MessageFactory.cs
```diff
@@ -204,6 +204,10 @@ public static class MessageFactory
     /// The K+ Metadata for a series has been updated
     /// </summary>
     public const string ExternalMetadataUpdate = nameof(ExternalMetadataUpdate);
+    /// <summary>
+    /// Progress event send after a batch completes
+    /// </summary>
+    public const string ReRunMappingsProgress = nameof(ReRunMappingsProgress);
```
majora2007: Let's say RerunMetadataMappingsProgress

Kavita.Server/Controllers/SettingsController.cs
```diff
@@ -241,6 +245,27 @@ public async Task<ActionResult<MetadataSettingsDto>> UpdateMetadataSettings(Meta
...
+            return BadRequest(await localizationService.TranslateAsync("mappings-re-run-in-progress"));
+        }
+
+        BackgroundJob.Enqueue<IMetadataService>(s => s.ReRunMappings(request, CancellationToken.None));
+
+        return Ok();
+    }
+
+    [Authorize(PolicyGroups.AdminPolicy)]
+    [HttpGet("re-run-mappings-running-or-queued")]
```
majora2007: No...I don't like this name. I would prefer to move this to ServerController and allow for it to give you if a task is running. We can control which tasks can be checked against. 

This would allow plugins to check if scans are going on, etc.

Kavita.Server/I18N/en.json
```diff
@@ -232,6 +232,7 @@
     "scrobble-provider-not-setup": "Scrobble Provider {0} is not setup",
     "cbl-import-download-from-url": "Unable to download file from URL",
     "cbl-import-validation-types": "Only .cbl and .json files are allowed",
+    "mappings-re-run-in-progress": "A re run mappings request is currently being processed. Wait until it is complete to requeue",
```
majora2007: Metadata needed and re-run is one word.

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -308,11 +308,11 @@ private static bool HasRequiredId(Series series, MetadataProvider metadataProvid
             .Where(e => e.Format.Equals("Digital", StringComparison.OrdinalIgnoreCase))
             .ToList();
 
-        return parsedSeriesEditionMatches.SingleOrDefault()
-            ?? parsedSeriesEditionMatches.SingleOrDefault(MatchEditionToCount)
-            ?? digitalEditions.SingleOrDefault()
-            ?? digitalEditions.SingleOrDefault(MatchEditionToCount)
-            ?? editions.SingleOrDefault(MatchEditionToCount);
+        return parsedSeriesEditionMatches.OneOrDefault()
```
majora2007, line 312: Let's just add a comment that I want exactly one match otherwise null.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+            logger.LogWarning("Ignoring re-run mappings request as extended metadata processing is not enabled");
+            return;
+        }
+
+        var seriesIds = unitOfWork.DataContext.Series
+            .WhereIf(!request.AllLibraries, s => request.IncludedLibraries.Contains(s.LibraryId))
+            .WhereIf(request.ExcludedLibraries.Count > 0, s => !request.ExcludedLibraries.Contains(s.LibraryId))
+            .AsQueryable()
+            .AsSplitQuery()
+            .AsNoTracking()
```
majora2007: We can remove this when we are selecting a single property and not the underlying entity, as it's implied.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+        if (!settings.EnableExtendedMetadataProcessing)
+        {
+            logger.LogWarning("Ignoring re-run mappings request as extended metadata processing is not enabled");
+            return;
+        }
+
+        var seriesIds = unitOfWork.DataContext.Series
+            .WhereIf(!request.AllLibraries, s => request.IncludedLibraries.Contains(s.LibraryId))
+            .WhereIf(request.ExcludedLibraries.Count > 0, s => !request.ExcludedLibraries.Contains(s.LibraryId))
+            .AsQueryable()
```
majora2007: This can run all in the DB

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+        var settings = await unitOfWork.SettingsRepository.GetMetadataSettingDto(cancellationToken);
+
+        if (!settings.EnableExtendedMetadataProcessing)
+        {
+            logger.LogWarning("Ignoring re-run mappings request as extended metadata processing is not enabled");
+            return;
+        }
+
+        var seriesIds = unitOfWork.DataContext.Series
+            .WhereIf(!request.AllLibraries, s => request.IncludedLibraries.Contains(s.LibraryId))
```
majora2007: We need protections against large selections of libraries. Anything over 50 when translating to an `IN()` statement has massive performance implications on SQLite. We have Batch extensions to help with this.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+            ProgressEventType.Started, 0), ct: cancellationToken);
+
+        await foreach (var idBatch in seriesIds.BatchToAsyncEnumerable(batchSize, cancellationToken))
+        {
+            currentBatch++;
+
+            using var scope = scopeFactory.CreateScope();
+            var scopedUnitOfWork = scope.ServiceProvider.GetRequiredService<IUnitOfWork>();
+
+            var seriesBatch = await scopedUnitOfWork.DataContext.Series
```
majora2007: Since this is quite heavy, we should likely inform the user the performance is similar to a scan.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+                {
+                    await ReRunMappingsForSeries(scopedUnitOfWork, series, settings);
+                }
+                catch (Exception ex)
+                {
+                    logger.LogError(ex, "Failed to re-run mappings for series {SeriesName} - {SeriesId}", series.Name, series.Id);
+                }
+            }
+
+            await scopedUnitOfWork.CommitAsync(cancellationToken);
```
majora2007: You need better try/catch/finally (finally to send the complete event). 

You aren't catching exceptions on the db work at all, which is most likely to throw.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+
+            foreach (var series in seriesBatch)
+            {
+                try
+                {
+                    await ReRunMappingsForSeries(scopedUnitOfWork, series, settings);
+                }
+                catch (Exception ex)
+                {
+                    logger.LogError(ex, "Failed to re-run mappings for series {SeriesName} - {SeriesId}", series.Name, series.Id);
```
majora2007: I like to use `Series Name (Series Id)` and have been using in other places. Let's align.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+                continue;
+            }
+
+            var allTagsAndGenres = chapter.Genres
+                .Select(g => g.Title)
+                .Concat(chapter.Tags.Select(t => t.Title));
+
+            chapter.AgeRating = ExternalMetadataService.DetermineAgeRating(allTagsAndGenres, settings.AgeRatingMappings);
+
+            chapter.TagsLocked = true;
```
majora2007: Don't we also need to update to have a Kavita+ override or do we assume? 

This code seems to override any sort of locked fields. We probably want to only do it when we have a K+ override or a force boolean.

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+            .Select(g => g.Title)
+            .Concat(series.Metadata.Tags.Select(t => t.Title));
+
+        series.Metadata.AgeRating = ExternalMetadataService.DetermineAgeRating(allSeriesTagsAndGenres, settings.AgeRatingMappings);
+
+        series.Metadata.TagsLocked = true;
+        series.Metadata.GenresLocked = true;
+        series.Metadata.AgeRatingLocked = true;
+
+        logger.LogTrace("Completed re-running mappings for series {SeriesName} - {SeriesId}", series.Name, series.Id);
```
majora2007: Formatting

Kavita.Services/MetadataService.cs
```diff
@@ -326,6 +337,128 @@ public async Task RemoveAbandonedMetadataKeys(CancellationToken ct = default)
...
+                .Concat(chapter.Tags.Select(t => t.Title));
+
+            chapter.AgeRating = ExternalMetadataService.DetermineAgeRating(allTagsAndGenres, settings.AgeRatingMappings);
+
+            chapter.TagsLocked = true;
+            chapter.GenresLocked = true;
+            chapter.AgeRatingLocked = true;
+        }
+
+        var genres = series.Volumes.SelectMany(v => v.Chapters).SelectMany(c => c.Genres).ToList();
```
majora2007, line 490: Need some comments to at least break up the chap tags then series tags processing

Kavita.Services/SettingsService.cs
```diff
@@ -72,8 +72,16 @@ public async Task<MetadataSettingsDto> UpdateMetadataSettings(MetadataSettingsDt
 
         existingMetadataSetting.AgeRatingMappings = dto.AgeRatingMappings ?? [];
 
-        existingMetadataSetting.Blacklist = (dto.Blacklist ?? []).Where(s => !string.IsNullOrWhiteSpace(s)).DistinctBy(d => d.ToNormalized()).ToList() ?? [];
-        existingMetadataSetting.Whitelist = (dto.Whitelist ?? []).Where(s => !string.IsNullOrWhiteSpace(s)).DistinctBy(d => d.ToNormalized()).ToList() ?? [];
+        existingMetadataSetting.Blacklist = (dto.Blacklist ?? [])
```
majora2007: Let's refactor this to a method `NormalizeTags` so we aren't repeating ourselves here

Kavita.Services/TaskScheduler.cs
```diff
@@ -791,6 +791,23 @@ public static bool RunningAnyTasksByMethod(IEnumerable<string> classNames, strin
...
+
+        var monitoring = JobStorage.Current.GetMonitoringApi();
+
+        var processing = monitoring.ProcessingJobs(0, 100);
+        if (processing.Any(j => j.Value.Job?.Method.Name == methodName))
+            return true;
+
+        var enqueued = monitoring.EnqueuedJobs(queue, 0, 100);
+        if (enqueued.Any(j => j.Value.Job?.Method.Name == methodName))
+            return true;
```
majora2007, line 809: brackets

Kavita.Services/TaskScheduler.cs
```diff
@@ -791,6 +791,23 @@ public static bool RunningAnyTasksByMethod(IEnumerable<string> classNames, strin
...
 
+    public static bool IsMethodRunningOrEnqueued(string methodName, string queue = DefaultQueue)
+    {
+        using var connection = JobStorage.Current.GetConnection();
+
+        var monitoring = JobStorage.Current.GetMonitoringApi();
+
+        var processing = monitoring.ProcessingJobs(0, 100);
+        if (processing.Any(j => j.Value.Job?.Method.Name == methodName))
+            return true;
```
majora2007, line 803: brackets

UI/Web/src/assets/langs/en.json
```diff
@@ -1299,6 +1299,8 @@
         "export-file-name": "Kavita Metadata Settings Export.json",
         "import-settings": "Import",
         "import-tooltip": "Import your or someone else's settings",
+        "re-run-mappings": "Re Run Mappings",
```
majora2007: Inconsistent naming

majora2007 (changes requested):

Kavita.Services/MetadataService.cs
Fesaa: Actually, I forgot to remove the locks. It makes no sense to lock because we're not losing data. A scan will give the same result, as the mappings work on exisitng tags. 

Okey, I guess you could actually lose data if run it twice and change settings in between. But I don't think we should care about that. 

Also seems alright to ignore locks because the settings were decided by the user. Seems weird to me to have a blacklist then run this on a library but have nothing work because you only use the UI for metadata and that locks everything. 

Thoughts? I've removed the locking for now, I forgot to do so

Kavita.Services/MetadataService.cs
Fesaa: There's no I/O work at all, so it's actually very fast. Sqlite is pretty fast, and people are the slowest part of the scan loop (After covers and other I/O). 

The current description of the toast is "Processing will kick of shortly, please be patient.". Can add some extra warning there?

Kavita.Services/MetadataService.cs
Fesaa: Do people have that many libraries? I'm not sure how to batch this as there's two things to batch at once. At least once needs to be fully then right? Or we need load them separately and then A\B them?

Kavita.Services/MetadataService.cs
majora2007: We could remove the locks and just leave it as-is. Essentially re-run and leave whatever was locked as locked.

Kavita.Services/MetadataService.cs
majora2007: I meant heavy on RAM usage

Fesaa: We can update the tooltip under the button to warn for it

Kavita.Models/Attributes/EnumCollectionAttribute.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.Collections;
+using System.ComponentModel.DataAnnotations;
+
+namespace Kavita.Models.Attributes;
+
+public class EnumCollectionAttribute(Type enumType, bool allowNull = false): ValidationAttribute
```
majora2007, line 12: This needs some light documentation to explain what it does and that it should be on all enums.

Kavita.Models/DTOs/Account/AgeRestrictionDto.cs
```diff
@@ -9,6 +10,7 @@ public sealed record AgeRestrictionDto
     /// <summary>
     /// The maximum age rating a user has access to. -1 if not applicable
     /// </summary>
+    [EnumDataType(typeof(AgeRating))]
```
majora2007, line 13: Out of curiosity, did you try to annotate the Enum itself and see if it carried forward to the DTO validation?

Kavita.Services/MetadataService.cs
majora2007: Yes, 200 libraries is not uncommon. I would just do a quick count check and only batch one list.

Kavita.Services/SettingsService.cs
```diff
@@ -113,6 +105,15 @@ public async Task<MetadataSettingsDto> UpdateMetadataSettings(MetadataSettingsDt
 
         // Return updated settings
         return await unitOfWork.SettingsRepository.GetMetadataSettingDto(ct);
+
+        List<string> NormalizeTags(IEnumerable<string>? tags)
```
majora2007: Probably wouldn't be bad to refactor this to a static method in TagHelper and have a unit test for it.

UI/Web/src/app/_services/server.service.ts
```diff
@@ -8,6 +8,15 @@ import {KavitaMediaError} from '../admin/_models/media-error';
 import {TextResonse} from "../_types/text-response";
 import {map} from "rxjs/operators";
 
+export enum TaskMethodNames {
+  ReRunMappings = 'ReRunMappings'
```
majora2007: Looks like this missed the Metadata naming update

UI/Web/src/assets/langs/en.json
```diff
@@ -1299,8 +1299,8 @@
         "export-file-name": "Kavita Metadata Settings Export.json",
         "import-settings": "Import",
         "import-tooltip": "Import your or someone else's settings",
-        "re-run-mappings": "Re Run Mappings",
-        "re-run-mappings-tooltip": "Re run your configured metadata mappings",
+        "re-run-mappings": "Run Metadata Mappings",
+        "re-run-mappings-tooltip": "Process series through your metadata mappings. This may take a while and consome more resources.",
```
majora2007: Typos:

This may take a while and consume more resources.

majora2007 (changes requested):

Kavita.Models/DTOs/Account/AgeRestrictionDto.cs
Fesaa: The Attribute is to be applied on Fields, Properties, etc not classes/enum so this wouldn't be possible

PR #4808, opened 2026-07-29
Kavita+ Series Name/Localized Name Enhancements

### LINKED ISSUE #4380: Natural Sort Order (or your equivalent) not working on my issues
### What happened?

I use a number field formatting that often can denote an additional informational suffix onto the issue number. Most often this is " (X Story)" to denote a specific story within the issue. Otherwise I use comicvine's numbering formats.  I was told by Amelia this should be sorting these things accurately via a natural sort order (or your customized equivalent) but it looks like this is not the case. In this screenshot, you can see all the alphanumeric numbers (those with "<whitespace>(X Story)" after their numbers are all clumping in a random-seeming order at the beginning.

Image

They may look like multiples of the same number but you can see the "(" before they get cut off. The comicinfo xml is correct, the numbers are like:
< Number >13< /Number> for the base issue
< Number >13 (A Story)< /Number>
< Number >13 (B Story)< /Number> etc for the files with just individual stories.

The sort order for the alphanumeric numbered issues are all 0, indicating to me that it's a "we don't know how to parse this number so just default to 0", but even at that they appear not to be sorted by strings within those 0s, it's probably just index order or random order.

### What did you expect?

What I would like to see is a natural sort order... effectively, taking the Number comicinfo tag and chopping off the  leading digits to determine an "effective number" (which I think is basically what you're already putting in the sort order field, unless that is meant to be absolutely sequential within the volume), but then taking the rest of the string and storing it separately, say as "remaining_string". then, in SQL terms, "Order by effective_number, remaining_string, number" so that when the effective number is the same, they then get string-sorted by whatever's left.

So for instance, assuming these are all the literal Number values from the xml, you'd get sorting like this:

2
02
02 (A Story)
02 (B Story)
002 (C Story) <- an example of why the remaining_string would need to be separate... if just sorting by "effective_number then number", the next line would sort before this one
02 (D Story)

I don't expect this to be perfect nor do I expect it to handle absolutely every variation perfectly. But just getting the "parent issues" in the right order would be a big help

0.8.9.1 - Stable

Yes

Mac

Mac

Chrome

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

### LINKED ISSUE #4716: "Less Than" and "Greater Than" is backwards on the "Read Last" filter

### LINKED ISSUE #4820: TASM (2018) showing issues out of order
### What happened?

In Kavita, The Amazing Spider-Man (2018) has multiple issues out of order and showing up at the beginning of the series instead of being in order. After reaching out in the Discord channel, I was told to make a bug-request. Issues that are showing before 1 are:
19.HU
20.HU
54.LR
92.BEY
16.HU
51.LR
78.BEY
50.LR
52.LR
80.BEY
18.HU
88.BEY

This is the exact order they show up in before issue #1. They are supposed to be in order based on LOCG. This issue is showing up on Kavita across multiple machines, multiple browsers, multiple OS systems.

Personal machine: Aurora Linux + Firefox
Work computer: Windows 11 + Chrome
iPad + Panels + Kavita Reader (Beta app)

Image

### What did you expect?

The expectation is that they would show up in release order/numerical order. For example, *).BEY is supposed to come after 80.

https://leagueofcomicgeeks.com/comics/series/137131/the-amazing-spider-man

0.9.0 - Stable

Yes

Docker (Dockerhub Container)

Linux

Firefox

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

_No response_

majora2007:
This should be the last big feature around Series Name updates. You can now tell Kavita a list of language priorities on how to write Series Name/Localized Names with library overrides. Feedback is critical before this goes live to Stable. 

image

# Added
- Added: Added the ability to configure Metadata Settings for Kavita+ around what languages to prioritize for setting series name and localized name. You can configure Global settings and per-library and use multiple language codes, which act as a fallback. (ie en;ja-Latn. If english is not available, will try japanese (Romanji), else don't update. The selected language code will not be used for localized name)
- Added: Added support for {Native} and {Romaji} reserved tokens for the language to fill in the Native and Romaji titles respectively.
- Added: Kavita will now not allow any localized name changes that would orphan files. This means that if a localized name is used to merge a set of files into one series, you can't change it. (ie - dont toy with me v01.cbz and Ijiranaide, Nagatoro-san v2.cbz in one series, changing localized name from Ijiranaide, Nagatoro-san will be rejected as next scan would split these into 2 series)
- Added: Added collapse series relationships as an option for series filter. Overwriting your global preferences for that filter. (Closes #4664)

# Changed
- Changed: Match modal now uses a rich url (pre-filtered) for Hardcover
- Changed: Manage Metadata is now in accordions to make it easier to navigate
- Changed: Inverted the behaviour of the last read filter field to be more intuitive. (Closes #4716)

# Fixed
- Fixed: Fixed a bug where SortName wouldn't update when Series Name was (develop)
- Fixed: Fixed series with non numerical chapter positions always being sorted at the start. (Fixes #4820, Fixes #4380)


Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -2113,27 +2130,44 @@ private static (bool, MetadataFieldChangeDto?) UpdateLocalizedName(Series series
 
         var from = series.LocalizedName;
 
+        // A localized name that normalizes to another series' Name/LocalizedName/OriginalName in the same library+format
+        // breaks the scanner
+        var takenNames = await _unitOfWork.SeriesRepository.GetTakenNormalizedNamesInLibraryAsync(
```
Fesaa: Might need checking how many alternative names series have, but it is likely better to do the reverse. 

Pass all options, normalize them (A) and then do an query that checks if existing are contained in the noramlized list. From the matching ones (B) you return the names. And then you can take A\B

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -2113,27 +2130,44 @@ private static (bool, MetadataFieldChangeDto?) UpdateLocalizedName(Series series
...
                 .ToList();
 
-            if (validSynonyms.Count == 0) return (false, null);
-
-            series.LocalizedName = validSynonyms[^1];
-            series.LocalizedNameLocked = true;
+            chosen = validSynonyms.FirstOrDefault(synonym => !takenNames.Contains(synonym.ToNormalized()));
         }
-        else if (IsRomanCharacters(externalMetadata.Name))
+        else if (IsRomanCharacters(externalMetadata.Name) && !takenNames.Contains(externalMetadata.Name.ToNormalized()))
```
Fesaa: Why does it need to be Roman? I know I put the Japanese letters as Localized title. But this is likely something for metadata provider settings (I.e. which title/language to prefer)

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -2113,27 +2130,44 @@ private static (bool, MetadataFieldChangeDto?) UpdateLocalizedName(Series series
...
         {
-            series.LocalizedName = externalMetadata.Name;
-            series.LocalizedNameLocked = true;
+            chosen = externalMetadata.Name;
         }
 
+        if (chosen == null)
+        {
+            _logger.LogInformation("[K+] Skipping localized name write for Series {SeriesId}: no unique candidate was found", series.Id);
```
Fesaa: These might want to be Debug? I don't know if we have any real idea's on what is what level. I tend to go by how often it gets called and how needed it is to know

Kavita.Services/Plus/ExternalMetadataService.cs
majora2007: Yeah good idea, would be much faster.

Kavita.Services/Plus/ExternalMetadataService.cs
majora2007: This is the existing code. I haven't yet hooked in the new K+ structure which will do away with this.

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -843,60 +849,114 @@ public async Task<bool> WriteExternalMetadataToSeries(ExternalSeriesDetailDto ex
         var settings = await _unitOfWork.SettingsRepository.GetMetadataSettingDto(ct);
         if (!settings.Enabled) return false;
 
-        var series = await _unitOfWork.SeriesRepository.GetSeriesByIdAsync(seriesId, SeriesIncludes.Metadata | SeriesIncludes.Related, ct);
-        if (series == null) return false;
+        var sw = Stopwatch.StartNew();
```
Fesaa: Should we remove debug in the release build?

UI/Web/src/app/admin/manage-metadata-settings/manage-metadata-settings.component.ts
```diff
@@ -144,9 +252,15 @@ export class ManageMetadataSettingsComponent implements OnInit {
...
 
+      this.settingsForm.valueChanges.pipe(
+        takeUntilDestroyed(this.destroyRef),
+      ).subscribe(() => this.cdRef.markForCheck());
+
       this.settingsForm.valueChanges.pipe(
         debounceTime(300),
         takeUntilDestroyed(this.destroyRef),
+        // Never persist a malformed language code
+        filter(() => !this.hasMalformedLanguageCodes()),
```
Fesaa: This should probably just be filter(() => this.settingsForm.valid),

PR #4826, opened 2026-08-07
Feature/series metadata provider override

StereotypicalCat:
# Added
  - Added: Series can now override their Library's default Metadata Provider. The override is set per Series from the Edit Series modal (External Metadata Ids tab) or directly from the Match dialog, and applies to matching, metadata fetching, reviews, ratings, recommendations and external covers
  - Added: Pasting a provider URL or header into the Match dialog's search box (hardcover.app, mangabaka.org, anilist.co, myanimelist.net, comicbookroundup.com, or a 'hardcover:'/'mb:'/'al:' style header) now switches the Series to that provider automatically before searching, instead of searching the wrong
  provider and returning nothing
  - Added: A Hardcover book or series id can now be pulled out of a pasted hardcover.app link, so you no longer have to hunt down the id by hand to match a Series
  - Added: Changing a Series' Metadata Provider is now written to the Kavita+ audit log, so you can see when a Series moved from one provider to another

  # Changed
  - Changed: Changing a Series' Metadata Provider now clears the reviews, ratings, recommendations and cached covers that were fetched from the previous provider, and lifts the blacklist, so the Series is matched and refetched against the provider you just picked instead of continuing to show the old provider's data
  - Changed: Match results are cached per provider, so switching provider and searching again returns results from the new provider instead of the previous provider's cached results for up to a minute
  - Changed: A Series can be pinned to any Metadata Provider, including ones that are not the usual choice for its Library type. Library defaults themselves are still restricted to the providers valid for the Library type

  # Fixed
  - Fixed: Fixed a bug where pasting a Hardcover link copied from Kavita's own Series page (https://hardcover.app/series/id/12345) into the Match dialog was read as a series slug named 'id', which threw away the search query and returned nonsense results (develop)
  - Fixed: Fixed a bug where a failed Metadata Provider change in the Match dialog (expired Kavita+ license, network error) left the dialog spinning forever with the search button disabled, with no way to recover short of reopening it. The dropdown now reverts to the provider the server actually holds (develop)


Kavita.Database/Repositories/SeriesRepository.cs
```diff
@@ -656,7 +656,9 @@ public async Task<PagedList<SeriesDto>> GetSeriesDtoForLibraryIdAsync(int userId
             .Include(s => s.ExternalSeriesMetadata)
             .Select(series => new SeriesDetailRequestV3Dto
             {
-                Provider = series.Library.MetadataProvider,
+                // Inlined rather than calling Series.GetEffectiveMetadataProvider(): this projection is translated to SQL by EF Core,
```
Fesaa: This comment isn't needed

Kavita.Database/Repositories/SeriesRepository.cs
```diff
@@ -690,6 +692,15 @@ public async Task<PagedList<SeriesDto>> GetSeriesDtoForLibraryIdAsync(int userId
         return result;
     }
 
+    public async Task<MetadataProvider?> GetEffectiveMetadataProviderAsync(int seriesId, CancellationToken ct = default)
+    {
+        // Inlined rather than calling Series.GetEffectiveMetadataProvider(), see GetKavitaPlusSeriesDetailRequestV3Dto
```
Fesaa: This comment is also not needed

Kavita.Services.Tests/KavitaPlusConfigurationTests.cs
Fesaa, line 1: Please remove this test, the logic for this is trivial. And this is just going to cause failed tests in GA because we forget to update this

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -389,7 +389,7 @@ public async Task<IList<ExternalSeriesMatchDto>> MatchSeries(MatchSeriesDto dto,
             ? mangabakaId : ExternalIdParser.GetMangaBakaId(query);
 
         var potentialHardcoverSlug = ExternalIdParser.TryParseHardcoverHeader(query, out var hardcoverId)
-            ? hardcoverId : null;
+            ? hardcoverId : ExternalIdParser.GetHardcoverSlugFromUrl(query);
```
Fesaa: The GetHardcoverSlugFromUrl call knows if the url is a book or series. It should set IsStandAlone accordingly. (In case the user forgot / is wrong).

This entire method is getting rather messy with the fallbacks and ternary operators. We might want to change it into

var request = new ();
SetMangabakaId(request, query); // Or similar
// etc.

Fesaa: We can do this afterwards

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -633,6 +633,51 @@ await _auditService.LogMatchAsync(KavitaPlusEventType.SeriesDontMatchSet, series
...
 
+    public async Task UpdateSeriesMetadataProviderOverride(int seriesId, MetadataProvider? metadataProviderOverride, CancellationToken ct = default)
+    {
+        var series = await _unitOfWork.SeriesRepository.GetSeriesByIdAsync(seriesId,
+            SeriesIncludes.Library | SeriesIncludes.ExternalMetadata, ct);
+        if (series == null) return;
+        if (series.MetadataProviderOverride == metadataProviderOverride) return;
+
+        var previousProvider = series.GetEffectiveMetadataProvider();
+        series.MetadataProviderOverride = metadataProviderOverride;
```
Fesaa, line 735: Lets set the override to null if it's equal to the library's provider.

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -906,7 +951,7 @@ public async Task<bool> WriteExternalMetadataToSeries(ExternalSeriesDetailDto ex
     private async Task ApplyExternalCovers(Series series, MetadataSettingsDto settings, CancellationToken ct = default)
     {
         if (!settings.EnableVolumeCoverImage && !settings.EnableChapterCoverImage) return;
-        if (series.Library?.MetadataProvider == MetadataProvider.ComicBookRoundup) return;
+        if ((series.MetadataProviderOverride ?? series.Library?.MetadataProvider) == MetadataProvider.ComicBookRoundup) return;
```
Fesaa: Use your GetEffectiveMetadataProvider method

UI/Web/src/app/_pipes/kavita-plus-event-description.pipe.ts
```diff
@@ -137,4 +143,14 @@ export class KavitaPlusEventDescriptionPipe implements PipeTransform {
 
     return '';
   }
+
+  /** MetadataProviderTitlePipe can't be instantiated here as it relies on DI, so resolve the same keys directly */
+  private metadataProviderName(provider: MetadataProvider | null): string {
```
Fesaa: Update MetadataProviderTitlePipe to not inject the service, and use translate instead. So we can create the pipe above, and not have duplicate code to add new translations too (See ScrobbleReadStatusPipe)

UI/Web/src/app/_single-module/match-series-modal/match-series-modal.component.ts
```diff
@@ -137,6 +141,32 @@ export class MatchSeriesModalComponent implements OnInit {
...
       takeUntilDestroyed()
     ).subscribe(() => this.search());
+
+    this.formGroup.controls.metadataProviderOverride.valueChanges.pipe(
+      takeUntilDestroyed()
+    ).subscribe(provider => this.changeProvider(provider));
+  }
+
+  private changeProvider(provider: MetadataProvider | null) {
+    this.seriesService.updateMetadataProviderOverride(this.series().id, provider).subscribe({
```
Fesaa: This is hard to follow, because of the nested Obserables:

```ts
updateMetadataProviderOverride(...).pipe(
  switchMap(() => {
     // ...
     return this.seriesService.getMatchInfo(...),
    },
   tap(res => {
     this.matchInfo.set(res);
     this.search();
   }),
   catchError(...),
).subscribe();
```

UI/Web/src/app/_single-module/match-series-modal/match-series-modal.component.ts
```diff
@@ -156,8 +186,20 @@ export class MatchSeriesModalComponent implements OnInit {
...
 
+    const query = this.formGroup.value.query ?? '';
+    const detectedProvider = this.detectProviderFromQuery(query);
+    const currentProvider = this.matchInfo()?.primaryProvider ?? null;
+
+    if (detectedProvider !== null && detectedProvider !== currentProvider && this.autoSwitchedQuery !== query) {
+      // Switches the override to the detected provider; changeProvider() re-runs search() once that lands
+      this.autoSwitchedQuery = query;
+      this.isLoading.set(true);
+      this.formGroup.controls.metadataProviderOverride.setValue(detectedProvider);
```
Fesaa: This flow won't work. It changes the provider without picking a match. So it would mess up if I'm just checking and then not going with a match.

I would instead

- Switch provider for calls to K+ if the url is found (Not on series yet)
- Have it return the matches & which provider used (If step above changed it)
- When making a match. Send the ids AND provider.
- If the provider is different here; then write the override

The UI can then show the different provider when results are shown

UI/Web/src/app/_single-module/match-series-modal/match-series-modal.component.ts
```diff
@@ -176,6 +218,31 @@ export class MatchSeriesModalComponent implements OnInit {
     ).subscribe();
   }
 
+  /**
+   * Mirrors the header/URL detection ExternalIdParser does server-side, so the dialog can switch the
+   * provider override to match a pasted URL/header before actually issuing the search
+   */
+  private detectProviderFromQuery(query: string): MetadataProvider | null {
```
Fesaa: You can ignore the comments below as this would all be moved to the backend

Kavita.Services/Plus/ExternalMetadataService.cs
Fesaa: Okey, circling back to this after the comments in the UI. You'll likely want some helper methods so this method doesn't get to big and complex

Fesaa (changes requested):
Some feedback to work through. You'll also need to pull in develop, I can't remember from the top of my head if we have had any DB migration. But if we had you'll need to

- Update Database
- Remove last migration
- Merge develop
- Create migration again

Using the Rider UI is the easiest for this

majora2007:
I'm not sure we are going to have time to merge this in for the v0.9.1 release, we may need to retarget v0.9.2. If you could, please try and fix this up in the next few days so it can be included with the Kavita+ overhaul.

Kavita.Services/Plus/ExternalMetadataService.cs
StereotypicalCat: of course!

UI/Web/src/app/_single-module/match-series-modal/match-series-modal.component.ts
```diff
@@ -176,6 +218,31 @@
...
+    const text = (query || '').trim().toLowerCase();
+    if (!text) return null;
+
+    if (text.startsWith('hardcover:') || text.includes('hardcover.app/')) {
+      return MetadataProvider.Hardcover;
+    }
+
+    if (text.startsWith('mangabaka:') || text.startsWith('mb:') || text.includes('mangabaka.org/')
+      || text.startsWith('anilist:') || text.startsWith('al:') || text.includes('anilist.co/')
+      || text.startsWith('mal:') || text.includes('myanimelist.net/')) {
```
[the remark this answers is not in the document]
StereotypicalCat: For these, should I just replace the includes with the common startsWith's instead? as in
* myanimelist.net/
* www.myanimelist.net/
* https://myanimelist.net/
* http://myanimelist.net/
* https://www.myanimelist.net/
* ... etc
It will probably bloat everything up a bit 😅

Kavita.Services/Plus/ExternalMetadataService.cs
StereotypicalCat: Yeah I agree, its kinda ugly. Ill try to make Uncle Bob proud 🫡

StereotypicalCat:
Apologies for the delay - I wanted to look at this earlier, but I think github shit itself yesterday, so I decided to wait for github to stop being weird.

Fesaa (approved):
Thanks for the update, this looks great. We'll try and fit this in with the current release still.

Kavita.Server/Controllers/SeriesController.cs
```diff
@@ -272,6 +271,10 @@ public async Task<ActionResult<SeriesDto>> UpdateSeries(UpdateSeriesDto updateSe
             return BadRequest(await localizationService.TranslateAsync(UserId, "generic-series-update"));
         }
 
+        // Routed through the service (rather than set inline) so that changing the provider also invalidates the
```
majora2007, line 274: You don't need to do anything, but this type of comment isn't needed at all.

Kavita.Models/DTOs/Metadata/Matching/MatchSeriesInfoDto.cs
```diff
@@ -22,10 +22,16 @@ public sealed record MatchSeriesInfoDto
     [EnumDataType(typeof(MetadataProvider))]
     public MetadataProvider? MatchedProvider { get; set; }
     /// <summary>
-    /// Who we would match against
+    /// Who we would match against. This is the effective provider: the Series' <see cref="MetadataProviderOverride"/>
+    /// when set, otherwise the Library's default.
```
majora2007, line 26: You don't need to do anything, but this is overkill and confusing. 

We realistically only need the Provider which is derived from the override or the library setting. I don't see any value add in having a PrimaryProvider which may or may not be in use, the override, and MatchedProvider (as matches will always align with PrimaryProvider).

Kavita.Models/Entities/Series.cs
```diff
@@ -209,4 +214,13 @@ public bool WillScrobble()
     {
         return !IsBlacklisted && !DontMatch;
     }
+
+    /// <summary>
+    /// The <see cref="MetadataProvider"/> that should be used for this Series: <see cref="MetadataProviderOverride"/>
+    /// when set, otherwise the parent <see cref="Library"/>'s default.
+    /// </summary>
+    public MetadataProvider GetEffectiveMetadataProvider()
```
majora2007, line 222: Again, these are comments to you, we will take care of them. 

You can use a property here like:
public MetadataProvider EffectiveMetadataProvider => MetadataProviderOverride ?? Library.MetadataProvider;

which will serialize to a field on the DTO.

majora2007: Also we would want to throw an exception is Library is null (aka not loaded from the source query)

majora2007 (approved):
As mentioned, we will take over this PR and clean it up for inclusion into the stable. Thanks for raising it, hope to see more PRs from you in the future.

UI/Web/src/app/_pipes/metadata-provider-title.pipe.ts
```diff
@@ -1,22 +1,22 @@
...
 
 @Pipe({
   name: 'metadataProviderTitle',
 })
 export class MetadataProviderTitlePipe implements PipeTransform {
 
-  private readonly transloco = inject(TranslocoService);
-
-  transform(value: MetadataProvider): string {
+  transform(value: MetadataProvider | null): string {
```
majora2007, line 10: You can't use translate here, need to inject the pipe for safety.

Kavita.Models/DTOs/Metadata/Matching/MatchSeriesInfoDto.cs
StereotypicalCat: Apologies! I was considering a future case where we might have a primary provider for metadata, and then pull missing metadata from a match.

StereotypicalCat:
Thank you for the comments and feedback 🫡 I hope to make future PRs even better, as I get more knowledge about the codebase and C# 💖

PR #4839, opened 2026-08-11
Workflow security fixes for the release pipelines

Totara-thib:
# Changed

- Changed: All workflow actions are now referenced by an exact commit instead of a floating tag, so the release and publish pipelines only ever run action code that existed at review time. Versions stay exactly where they were.
- Changed: Workflow jobs now declare the minimal token permissions they need. Only the issue commenter keeps a write scope, the docker publish jobs already had theirs scoped.

# Fixed

- Fixed: Fixed a shell injection risk in the debug steps of the nightly and stable workflows. Values now pass through environment variables, the printed output is unchanged.
- Fixed: Fixed the canary version bump referencing a GitHub repository that no longer exists, which would fail the next canary build. It now uses the same maintained fork the nightly workflow already uses, pinned like the rest.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: [ '**' ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Build and Test PR
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -10,12 +10,12 @@ jobs:
     runs-on: windows-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v5
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
             fetch-depth: 1
 
       - name: Setup .NET Core
-        uses: actions/setup-dotnet@v5
+        uses: actions/setup-dotnet@26b0ec14cb23fa6904739307f278c14f94c95bf1 # v5.4.0
         with:
             dotnet-version: 10.0.x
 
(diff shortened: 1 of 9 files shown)


Totara-thib:
Extra context since I have not raised this through an issue or Discord first, per the template note.

Why these changes, in security terms:

The stable workflow triggers on closed pull requests against develop, and its debug step expanded an expression built on github.head_ref inside a shell script. A branch name is attacker-controlled on pull request events, so a crafted name could inject shell commands into the runner that holds the publish credentials. Environment variable indirection removes that path with identical output.

SiqiLu/dotnet-bump-version is gone from GitHub. Beyond breaking the canary build at action download, a deleted repository name can be re-registered by anyone, and whoever claims it would run their code in that workflow. Pointing at the majora2007 fork restores what the workflow did before the upstream repo vanished.

On the pins: a version tag is a movable pointer, whoever controls an action repository can point it at different code after the fact. The tj-actions/changed-files incident (CVE-2025-30066) leaked CI secrets from thousands of repositories exactly this way. An exact commit cannot be retargeted. Each pin keeps the version as a trailing comment so it can be checked against the releases page, and if you ever enable Dependabot for github-actions it understands the format and keeps pins updated.

One heads-up: if the org ever restricts allowed actions in the repo settings with tag patterns such as owner/action@v3, those patterns stop matching SHA refs and workflows fail at startup. The fix is owner/action@* in that setting. Nothing to do if no such restriction is configured.

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.

majora2007:
Hi, thanks for the CI PR. It's too late in the release window for us to merge this, we will look at it for v0.9.2 release. FYI, canary is rarely used, but we appreciate fixing it up.

PR #4840, opened 2026-08-11
A few extra Kavita+ Metadata Mapping options

### LINKED ISSUE #4833: Current library metadata provider isn't reflected after update.
### What happened?

If you go change the library metadata provider under library -> advanced -> metadata provider -> save , the change functionally works, as can be seen by then going to match a series to metadata. However, if you go back into library -> advanced and look at metadata provider, it will go back to the original option set, not reflecting the new one.

### What did you expect?

Metadata Provider setting to accurately reflect the current setting.

Nightly Testing Branch

No

Docker (Other)

Windows

Firefox

None

_No response_

### Relevant log output

```shell

```

### Additional Notes

Tested on library type Comic (Flexible) after switching from CBR to Hardcover.

majora2007:
# Added
- Added: (Kavita+) Added the ability to take a raw string content rating from Mangabaka and map it to your own Age Rating. This will override the default mapping Kavita+ provides. Kavita will always take the highest age rating produced from the whole mapping process (so a tag mapping that produces a higher rating will be selected)
- Added: (Kavita+) Added the ability to filter out tag weightings (Mangabaka only) from above a weight. So if you select recurrent, incidental and unweighted will be removed. Does not affect remapped tags (ie tag generated from a genre mapping) or white listed tags. 
- Added: (Kavita+) Added the ability to turn off Kavita writing age ratings

# Changed
- Changed: (Kavita+) Tag weight filtering is only guarding the writing of tags now. Age Rating mapping will now run against all tags/genres (even before black/whitelisting) to ensure a string doesn't escape deriving an age rating.
- Changed: Tweaked some of the Kavita+ setting item language

# Fixed
- Fixed: Fixed MetadataProvider not showing correctly in the UI (Closes #4833)
- Fixed: Fixed chapter summary not always being written by K+
- Fixed: Fixed artists not always being written for Hardcover metadata
- Fixed: Fixed fields not being locked after K+ writes to them
- Fixed: Fixed Hardcover weblink id parsing (develop)


Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -971,6 +964,22 @@ public async Task<bool> WriteExternalMetadataToSeries(ExternalSeriesDetailDto ex
...
     }
 
+    private static HashSet<string> GetTagsToRemove(ExternalSeriesDetailDto externalMetadata, MetadataSettingsDto settings)
+    {
+        var whitelist = settings.Whitelist is { Count: > 0 }
+            ? settings.Whitelist.Select(s => s.ToNormalized()).ToHashSet()
+            : null;
+
+
+        var tagsToRemove = settings.FilterAboveWeight == null
```
Fesaa: I'd use some if statements here instead. This is a bit hard to follow

UI/Web/src/app/shared/_components/age-rating-mapper/age-rating-mapper.component.html
```diff
@@ -0,0 +1,44 @@
+<div [formGroup]="form()">
+  <div [formArrayName]="controlName()">
+    @for (mapping of mappingRows.controls; track mapping; let i = $index) {
+      <div [formGroupName]="i" class="row mb-2">
+        <div class="col-md-4 d-flex align-items-center justify-content-center">
+          @if (sourceLabel()) {
+            <label class="visually-hidden" [for]="idPrefix() + '-' + i">{{sourceLabel()}}</label>
+          }
+          <input [id]="idPrefix() + '-' + i" type="text" class="form-control" formControlName="str"
```
Fesaa, line 9: I think having this be free strings, isn't the best expierence for external age rating mappings. As user will need to look up specific values for Mangabaka. 
I do understand why, this way when Hardcover adds any it'll just work. Not sure how to work around this, but something to think about?
Could have a button that adds defaults for Mangabaka?

Kavita.Services/Plus/ExternalMetadataService.cs
```diff
@@ -1709,14 +1719,22 @@ private static List<string> ApplyBlackWhiteList(MetadataSettingsDto settings, Me
...
-            var ageRating = DetermineAgeRating(totalTags, settings.AgeRatingMappings);
-            if (series.Metadata.AgeRating <= ageRating)
-            {
-                series.Metadata.AgeRating = ageRating;
-                series.Metadata.AddKPlusOverride(MetadataSettingField.AgeRating);
 
-                return (true, new MetadataFieldChangeDto(MetadataFieldChangeKind.AgeRating, from.ToString(), ageRating.ToString()));
-            }
+            // Find the highest age rating from the different mapping mechanisms + the age rating from Kavita+
+            var externalAgeRating = externalMetadata.AgeRating;
```
Fesaa, line 1727: Us mapping this by default forces a minimum age rating on all series. I think we might want to only use this if there's no external ratings configured? As now users won't be able to lower age ratings, if they don't agree with us.

UI/Web/src/app/shared/_components/age-rating-mapper/age-rating-mapper.component.html
majora2007: My thinking was this is raw, so it could work for anything, because we already map for them. This is if they want to override.

PR #4877, opened 2026-08-25
Add RepoCloud to hosting options

cosark:
This PR adds [RepoCloud](https://repocloud.io/details/Kavita/) as a hosting option in the "I don't have a Server" section of the README.

RepoCloud provides one-click cloud deployment for open-source applications, making it easy for users to self-host Kavita without managing their own server infrastructure.


majora2007:
Unless there is an Open Source kickback, like the other providers provide, I'm not interested in free advertising for your site.

PR #4920, opened 2026-09-07
Log a warning when the Kavita+ license check returns no data for a saved license

### LINKED ISSUE #4914: A lapsed Kavita+ licence fails silently: license/info returns 204 with nothing logged
### What happened?

My Kavita+ licence stopped being accepted after a routine restart, and **nothing anywhere on the
server said so**. `GET /api/license/info` returned a bare **204 No Content**, no response body, and
not a single line at any log level. Kavita+ features simply stopped appearing in the UI.

The complete server log record for those three minutes is five 204s and nothing else:

```
08:29:42 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 733.6953 ms
08:29:48 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 223.1012 ms
08:29:57 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 213.4391 ms
08:30:29 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 218.8342 ms
08:30:29 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 207.4383 ms
```

No `LicenseService` entry at all. The server had already tried to re-register and been refused — but
that information is only surfaced if a human happens to open the settings page and re-submit the
key, at which point the real cause finally appears:

```
08:30:43 [Error] Kavita.Services.Plus.LicenseService Kavita+ registration failed.
         Code: "AlreadyRegistered", Message: This license is already registered to a
         different server - Support Token: <redacted>
```

So the diagnostic exists and is well-written. It just never reaches the log on the automatic path.

**Two related behaviours made this considerably harder to diagnose:**

- **`has-license` does not mean "licensed".** It returns `true` whenever a key is stored locally,
  regardless of whether that key is currently accepted. It read `true` throughout the outage. The
  only endpoint that answers the real question is `valid-license?forceCheck=true`.
- **`forceCheck=false` serves a cache, and the response is indistinguishable from a live one.** The
  only way I could tell them apart was response time — sub-5 ms is cached, 200–1100 ms went to the
  network. A run of apparently-healthy 200s after a failed registration can be entirely cache.

**Steps to reproduce**

1. Have a working Kavita+ subscription.
2. Put the licence into a state the server will refuse (in my case a stale server-side registration
   binding, surfaced as `AlreadyRegistered`).
3. Restart Kavita, which triggers re-registration.
4. Watch the log at any level — nothing is written. `GET /api/license/info` returns 204,
   `has-license` still returns `true`, and Kavita+ features are silently gone.

### What did you expect?

I expected a failed licence check or re-registration to write at least a warning to the log,
including the error code and support token that Kavita already has in hand.

As it stands there is no signal an administrator or an external monitor can act on: the endpoint
returns a success-shaped status, the log is empty, and `has-license` still reads `true`. I only found
it by noticing missing Kavita+ data by eye. On a self-hosted server that is the difference between
noticing in minutes and noticing in weeks.

Suggestions, roughly in order of value:

1. **Log a warning when a licence check or re-registration fails**, with the error code and support
   token. This is the important one — right now the failure path is entirely silent.
2. Consider returning a body on `license/info` describing the failed state instead of a bare 204, so
   an admin dashboard or external monitor can distinguish "not licensed" from "endpoint healthy,
   nothing to report".
3. Optionally surface a persistent admin notice in the UI when Kavita+ is configured but not
   currently valid.

### Kavita Version Number

0.9.1 - Stable (specifically 0.9.1.4)

### Are you accessing kavita through a reverse proxy?

Yes — but this is not a proxy issue. All the API calls above were made directly against the server on
localhost, bypassing the proxy entirely.

Linux (native install, not Docker — Debian 13, x64)

### Relevant log output

```shell
# The entire log record while the licence was lapsed - five 204s, nothing else at any level:
08:29:42 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 733.6953 ms
08:29:48 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 223.1012 ms
08:29:57 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 213.4391 ms
08:30:29 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 218.8342 ms
08:30:29 [Information] HTTP GET /api/license/info?forceCheck=false responded 204 in 207.4383 ms

# Only after manually re-submitting the key in the UI does the cause appear:
08:30:43 [Error] Kavita.Services.Plus.LicenseService Kavita+ registration failed.
         Code: "AlreadyRegistered", Message: This license is already registered to a
         different server - Support Token: <redacted>
08:30:43 [Information] HTTP POST /api/license responded 200 in 430.9097 ms
08:30:47 [Information] Kavita.Server.Controllers.LicenseController Resetting license on file for Server
08:30:48 [Information] HTTP POST /api/license/reset responded 200 in 1154.7046 ms
08:30:49 [Information] HTTP POST /api/license responded 200 in 1016.9508 ms
08:30:50 [Information] HTTP GET /api/license/info?forceCheck=true responded 204 in 796.2313 ms

# It then recovered on its own ~14 seconds later, with no further action - also unlogged:
08:31:04 [Information] Kavita.Services.Plus.ScrobblingService Processing Scrobbling want to read event for ...
```

### Additional Notes

This issue is only about the silence, not about my particular registration problem — I've raised the
underlying `AlreadyRegistered` binding separately with support, quoting the support token, since that
part is server-side. I've redacted the token here as it's tied to my subscription; happy to share it
privately if it helps correlate.

I searched for an existing report and didn't find one. The nearest are **#4257** ("Kavita+ licence
not receiving") and **#4705** ("Kavita+ Audit Log"), both closed, neither covering the silent-failure
behaviour itself.

atirna:
When a saved Kavita+ license is refused upstream (installId mismatch or no active subscription), `LicenseService.GetLicenseInfo` hit the `response == null` branch and returned without any log entry, so the automatic check failed completely silently: `GET /api/license/info` answered a bare 204 while `has-license` still read true and there was nothing at any log level for an admin or external monitor to act on (#4914).

The network-error path already logs (`FlurlHttpException` in both `KavitaPlusApiService` and `LicenseService`), so this change only covers the null-response branch: a `LogWarning` carrying the support token, same shape as the manual registration path (`RegisterLicense`) and `LinkDiscord` already use. The endpoint's response contract (`Ok(null)` / 204) is untouched since the UI and external monitors depend on it.

## Fixed
- A failed automatic Kavita+ license check (null upstream response for a saved license) now writes a Warning with the support token instead of passing silently

## Verification
- `dotnet test Kavita.Services.Tests --filter "Category!=Integration"` - all green, including 4 new `LicenseServiceTests` cases (null upstream response logs the warning; no saved license and cache-hit paths stay silent; valid response path stays silent)
- The warning case fails on the pre-fix code for the reported reason (no warning emitted) and passes with the change

Fixes #4914

### DIFF
--- a/Kavita.Services.Tests/Plus/LicenseServiceTests.cs
+++ b/Kavita.Services.Tests/Plus/LicenseServiceTests.cs
@@ -1,0 +1,152 @@
+using EasyCaching.Core;
+using Kavita.API.Database;
+using Kavita.API.Services;
+using Kavita.API.Services.Plus;
+using Kavita.API.Services.SignalR;
+using Kavita.Database;
+using Kavita.Database.Tests;
+using Kavita.Models.DTOs.KavitaPlus.License;
+using Kavita.Models.Entities.Enums;
+using Kavita.Services.Plus;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using NSubstitute;
+using Xunit.Abstractions;
+
+namespace Kavita.Services.Tests.Plus;
+
+#nullable enable
+
+/// <summary>
+/// Exercises the failure branches of <see cref="LicenseService.GetLicenseInfo"/> against mocked
+/// collaborators, without touching the Kavita+ HTTP boundary.
+/// </summary>
+public class LicenseServiceTests(ITestOutputHelper outputHelper) : AbstractDbTest(outputHelper)
+{
+    private const string LicenseKey = "encrypted-license-key";
+
+    private ILogger<LicenseService> _logger = null!;
+    private IEasyCachingProviderFactory _cacheFactory = null!;
+    private IKavitaPlusApiService _kavitaPlusApiService = null!;
+
+    private LicenseService CreateService(IUnitOfWork unitOfWork)
+    {
+        _logger = Substitute.For<ILogger<LicenseService>>();
+
+        var cacheProvider = Substitute.For<IEasyCachingProvider>();
+        // No cache hits: every check in these tests is an upstream call
+        cacheProvider.GetAsync<LicenseInfoDto>(Arg.Any<string>(), Arg.Any<CancellationToken>())
+            .Returns(new CacheValue<LicenseInfoDto>(null, false));
+        cacheProvider.GetAsync<bool>(Arg.Any<string>(), Arg.Any<CancellationToken>())
+            .Returns(new CacheValue<bool>(false, false));
+
+        _cacheFactory = Substitute.For<IEasyCachingProviderFactory>();
+        _cacheFactory.GetCachingProvider(Arg.Any<string>())
+            .Returns(cacheProvider);
+
+        _kavitaPlusApiService = Substitute.For<IKavitaPlusApiService>();
+
+        return new LicenseService(
+            _cacheFactory,
+            unitOfWork,
+            _logger,
+            Substitute.For<IVersionUpdaterService>(),
+            _kavitaPlusApiService,
+            Substitute.For<IFileCacheService>(),
+            Substitute.For<IEventHub>());
+    }
+
+    private static async Task SaveLicenseKey(DataContext context, string value)
+    {
+        var setting = await context.ServerSetting
+            .SingleAsync(s => s.Key == ServerSettingKey.LicenseKey);
+        setting.Value = value;
+        await context.SaveChangesAsync();
+    }
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


majora2007:
Thanks for the identification, I'm going to manually patch it in since I don't need all the text and it's a one liner.

PR #4923, opened 2026-09-08
Fix bad request if EPUB already encodes image path

LuigiARC:
# Fixed
- Fixed: Fixes an issue that results in images not being able to be displayed on the web reader (also tested on Kover on Android) due to the EPUB file already having an encoded path resulting in double encoding. In this case, an error 400 would be thrown by the api.

## Example
In an EPUB, from my understanding both of these are acceptable in terms of the format itself:
- images/V02 TBC.jpg
- images/V02%20TBC.jpg
Current behaviour results in https://kavita.REDACTED/api/book/4955/book-resources?apiKey=REDACTED&file=images%2FV02%2520TBC.jpg for the second option which does not fully decode and thus does not correctly point to the image inside the EPUB.

### DIFF
--- a/Kavita.Services.Tests/BookServiceTests.cs
+++ b/Kavita.Services.Tests/BookServiceTests.cs
@@ -204,4 +204,13 @@ public class BookServiceTests
         Assert.True(result.IsSuccess);
         Assert.Equal("image/png", result.ContentType);
     }
+
+    [Theory]
+    [InlineData("images/V02%20TBC.jpg", "images%2FV02%20TBC.jpg")]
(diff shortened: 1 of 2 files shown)


LuigiARC:
Fixed it in the simplest way I could think of: 
Trying to decode the path before then calling EscapeDataString on the already decoded path.

Not sure if this type of image pathing is uncommon but it was annoying me quite a lot because the main book I am reading at the moment has loads of images that are already encoded and the only way I had to see them was to use another reader on my desktop.

LuigiARC:
Is a literal % in an image name in a book an actual concern? I can't think of a reasonable example that would throw...

majora2007:
Please email or dm (discord) me a book where you were having this issue on originally.

Issue #4918, opened 2026-09-06
A library scan wipes the whole page cache while the reader is serving from it, deleting pages mid-chapter


Sawtaytoes:
### What happened?

A library scan enqueues a recursive delete of the **entire** page cache directory. That job runs while the reader is still serving pages out of it, so pages of the chapter you are reading are deleted underneath you. The delete then races with the live reads and throws, which can leave a cache directory partially deleted.

`ScannerService.cs` enqueues the wipe in two places after a scan finishes:

https://github.com/Kareadita/Kavita/blob/develop/Kavita.Services/Scanner/ScannerService.cs#L332-L333

```csharp
BackgroundJob.Enqueue(() => cacheService.CleanupChapters(existingChapterIdsToClean));
BackgroundJob.Enqueue(() => directoryService.ClearDirectory(directoryService.CacheDirectory));
```

and again at line 577:

```csharp
BackgroundJob.Enqueue(() => directoryService.ClearDirectory(directoryService.CacheDirectory));
```

Neither one checks whether a chapter under `CacheDirectory` is in use.

#### Steps to reproduce

1. Open a long chapter in the reader. A webtoon image-folder chapter of 100+ pages is the easiest case, because the cache directory is large enough that the delete and the read overlap.
2. Start reading, so the reader keeps requesting pages from `config/cache/<chapterId>/`.
3. Trigger a library scan on **any** library while you are still reading. Touching files in a watched folder is enough — `LibraryWatcher` fires the scan by itself.
4. Pages stop loading part-way through the chapter.

#### Evidence

Reading chapters 57 and 58 of one series while a watched folder in a *different* library was being rescanned. Chapter id `87289` is chapter 57 and `87573` is chapter 58.

Six cache-delete jobs failed inside one 11 minute reading session:

```
23:07:26  job 20688  IOException: Directory not empty : '/kavita/config/cache/8679'
23:11:58  job 24131  IOException: Directory not empty : '/kavita/config/cache/87289'
23:11:59  job 24138  IOException: Directory not empty : '/kavita/config/cache/87289'
23:12:39  job 24227  DirectoryNotFoundException: '/kavita/config/cache/87289/079.jpg'
23:14:59  job 24528  IOException: Directory not empty : '/kavita/config/cache/87573'
23:18:03  job 24582  DirectoryNotFoundException: '/kavita/config/cache/87289/'
```

Two of the traces:

```
[...]

PR #4924, opened 2026-09-08
Fixed: Library scan no longer wipes cached pages out from under a live reader

bunnysayzz:
# Fixed
- Fixed: Fixed a library scan wiping the whole page cache while the reader is serving from it, which deleted pages mid-chapter and failed the cleanup jobs with IOExceptions (Fixes #4918)

After a scan, ScannerService wiped the entire cache directory without checking whether any cached chapter was being read. The wipe now goes through CacheService.CleanupCacheExceptActiveChaptersAsync, which excludes chapters tied to currently active reading sessions (chapter ids from AppUserReadingSession activity data, which already close out on expiry, so the exclusion set stays fresh). Skipped chapters are picked up by a later scan once their sessions close. In addition, DirectoryService.ClearDirectory now skips (with a warning, instead of failing the job) a chapter directory that changes under it mid-wipe, which was the source of the reported Directory not empty / DirectoryNotFound failures.

### DIFF
--- a/Kavita.Services/CacheService.cs
+++ b/Kavita.Services/CacheService.cs
@@ -290,7 +290,7 @@ public class CacheService(
         {
             if (keepPaths.Contains(NormalizeDirPath(dir)))
             {
-                logger.LogDebug("Skipping cache cleanup for {Directory} with an active reading session", dir);
+                logger.LogTrace("Skipping cache cleanup for {Directory} with an active reading session", dir);
                 continue;
             }
             directoryService.ClearAndDeleteDirectory(dir);
@@ -298,7 +298,8 @@ public class CacheService(
 
         static string NormalizeDirPath(string path)
         {
-            return Path.GetFullPath(path).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            return Parser.NormalizePath(Path.GetFullPath(path)
+                .TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar));
         }
     }
 
--- a/Kavita.Services/CacheService.cs
+++ b/Kavita.Services/CacheService.cs
@@ -275,6 +275,34 @@ public class CacheService(
     }
 
     /// <summary>
+    /// Wipes the whole page cache directory, except chapters tied to currently active
+    /// reading sessions, so a scan never deletes pages out from under a live reader.
+    /// Skipped chapters are cleaned by a later scan once their sessions close.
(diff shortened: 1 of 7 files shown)


Kavita.Services/CacheService.cs
```diff
@@ -274,6 +274,34 @@ public void CleanupChapters(IEnumerable<int> chapterIds)
...
+        {
+            if (keepPaths.Contains(NormalizeDirPath(dir)))
+            {
+                logger.LogDebug("Skipping cache cleanup for {Directory} with an active reading session", dir);
+                continue;
+            }
+            directoryService.ClearAndDeleteDirectory(dir);
+        }
+
+        static string NormalizeDirPath(string path)
```
majora2007, line 299: We already have a NormalizePath in Parser or you can call the extension from a string.

Kavita.Services/CacheService.cs
```diff
@@ -274,6 +274,34 @@ public void CleanupChapters(IEnumerable<int> chapterIds)
...
+        var keepPaths = (await unitOfWork.ChapterRepository.GetActiveReadingChapterIdsAsync())
+            .Select(GetCachePath)
+            .Select(NormalizeDirPath)
+            .ToHashSet(StringComparer.OrdinalIgnoreCase);
+
+        foreach (var dir in directoryService.GetDirectories(directoryService.CacheDirectory))
+        {
+            if (keepPaths.Contains(NormalizeDirPath(dir)))
+            {
+                logger.LogDebug("Skipping cache cleanup for {Directory} with an active reading session", dir);
```
majora2007: This should be on Trace.

majora2007 (changes requested):

bunnysayzz:
Thanks for the quick review, both done. The skip log is on Trace now, and the path comparison goes through Parser.NormalizePath (kept the GetFullPath plus trailing-separator trim underneath so the keep-set and the directory listing still compare apples to apples). CacheService suite still 23/23 green.

bunnysayzz:
Saw this was closed. The round-1 feedback is applied and green on my side, so if something else was off, happy to fix it, just say what. Otherwise no worries, thanks for the review.

Fesaa:
Ah, this got auto closed because we merged (deleted) the branch it was targeting. If you could open a new PR targeting develop instead

bunnysayzz:
Thanks, that explains it. Re-opened as #4943 against develop, same fix plus your reviewer's round-1 feedback carried over.

